PR #52, opened 2022-03-25
Add dependent types

vsbogd:
Rewrote type checking procedure to support dependent types. As a side effect all tuples are always validated because they always have the product type. There is a question should we explicitly forbid this and require types like `(Pair ...)` for representing tuples or allow validating any tuple. I also have small questions on some `asserts` in tests on dependent types, but it is simpler to check one by one.


vsbogd:
I have fixed logic to pass unit tests. I didn't touch `Undefined` logic yet as we didn't finalized if it should be equal to `Any` type behavior.

vsbogd:
I have changed expression validation logic:
- Validate expression as a function call if first atom has arrow type;
- Validate expression as a product if first atom has non-expression type.
If atom has both then it can be used in both contexts.

Necr0x0Der (approved):

PR #71, opened 2022-04-21
return in call + call &self test

Necr0x0Der:


vsbogd:
The original `call` implementation suppose the called method gets atoms and returns atoms, so no additional wrapping/unwrapping is required. If it is not convenient for a specific scenario then we need specific call version which unwraps arguments and wraps the result.

Necr0x0Der:
The original call implementation didn't return any result at all. So, this is an extension of its functionality - not replacement. And while we don't have other use cases, I prefer the extension, which is needed right now.

vsbogd (approved):

PR #95, opened 2022-06-02
Interpreter version based on type checks

vsbogd:
Get type from grounded atom for type checks. Add interpreter operation to interpret the atom using required type. Rewrite interpreter using types:
- for single atoms the interpreter casts atom to the required type
- for expression it gets the first atom which designates operation and gets its type:
  - if type is functional then other atoms in the expression are interpreted using argument types
    - after all arguments are interpreted the results form the list of possible calls
    - each item in the list is called results are returned
  - if type is not functional then all atoms in the expressions are interpreted as Undefined type
    - after this the whole expression is called and results are returned.

`if` type definition is added to each test which uses `(if $cond $then $else)` expressions.


python/tests/test_grounded_type.py
```diff
@@ -24,9 +24,6 @@ def test_apply_type(self):
         self.assertEqual(
             metta.interpret("(untop)")[0].get_grounded_type(),
             metta.parse_single("untyped").get_grounded_type())
-        self.assertNotEqual(
```
Necr0x0Der, line 27: Maybe, I would not remove this, but just comment out. While it is clear that having an untyped operation of the type `(-> Undefined)` is incorrect, because it is arity will be wrong, but we still may want to distinguish types of operations and non-operational atoms. We discussed that it would be nice to have something like `*args` in the future. The type of untyped operations is yet another example showing this can be useful. So, the type of untyped operations might become something like `(-> *Undefined Undefined)` in the future. Of course, if this happens, we can add this unit test later, so it is not actually necessary to return it in the commented form.

vsbogd: Yes, I also hesitated to remove it. And main reason is to not forget add check for this. I can add separate unit test for it and ignore it with TODO comment. Then it will be more visible as ignored unit test rather then commented out part of the test.

Necr0x0Der: Sounds good

python/tests/test_minelogy.py
```diff
@@ -111,11 +111,11 @@ def test_minelogy(self):
             )
             ''')
         output = utils.interpret('(how-get cobblestone)')
-        self.assertEqual(repr(output[0]),
-            '(do-mine ((: stone type) (: stone variant)))')
+        self.assertEqual(repr(output),
+            '[(let $t (get-ingredients cobblestone) (do-craft $t)), (do-mine ((: stone type) (: stone variant)))]')
```
Necr0x0Der, line 115: I wonder, why `(let $t (get-ingredients cobblestone) (do-craft $t))` remains here. Was it there before, but as `output[1]`, which was ignored? `(get-ingredients cobblestone)` should return no result as it calls `match`. There is a general problem, as we discussed, to understand if "no result" should be propagated to the parent expression, so it also returns empty set of results, and when it means that the expression just should not be reduced. It might be possible to control this behaviour by the return type (or type annotation?) of the functional atom (in this case, `let`). Maybe, it worths somehow marking this example for further consideration within this general question.

python/tests/test_grounded_type.py
vsbogd: Fixed in #99

python/tests/test_minelogy.py
vsbogd: I investigated it further and found that I changed behavior while unified alternative planning for the argument reduction and matching/execution. The `(get-ingredients cobblestone)` is reduced to empty result because it is transformed to the grounded `match` operation and it in turn returns empty result. There is no `Err` which is returned by embedded matching. I reverted behavior to the previous state in #99.

Necr0x0Der: Nice. But in any case, it seems that we still need to sort out, when we do expect an empty result instead of a not reduced expression besides direct calls to `match`. For example, for
```
(= (if! True $x) $x)
```
we may want to return no result if the first argument is `False`. Now, `println! (if! False (+ 1 1)))` will print `(if! False (+ 1 1))`. There can be different approaches to it:
- to have real `void` or `abort` ("element" of `Void`). This "contradicts" to type and category theory, which say that no function can return "no result" (constructing a function of type `A -> Void` proves that `A` is false)... but who cares... so, `(= (if! False $_) void)` will explicitly abort this branch of evaluation of the parent expression... oh, well, we already have `nop` for this, which differs from `void` only in that it is a function..
- wrap the expression into a specific function forcing the interpreter to find equality for the wrapped expression or abort, e.g. `(force-reduce (if! False (+ 1 1)))`. The advantage is that we can select ad hoc behavior.
- explicitly define behavior via types (this might also make implementing `force-reduce` naturally).

PR #98, opened 2022-06-03
Fix variable name conflicts while querying atomspace

### LINKED ISSUE #84: Atomspace query doesn't return result because of variable names conflict
```
(= (append Nil $x1 $x1) True) 
 
(= (append (Cons $X1 $Y) $Z (Cons $X1 $W)) 
  (append $Y $Z $W) 
)
```
Working prints True
```
!(println! (append (Cons a2 Nil) (Cons b1 Nil) $W2222)) 
```

not working 
```
!(println! (append (Cons a2 Nil) (Cons b1 Nil) $W))
```

vsbogd:
This PR It fixes #84 but introduces equality issues after variable renaming.

What was done:
- unit tests are added to reproduce the issue, and to demonstrate renaming equality issue
- `id` field is added into `VariableAtom` to make unique variable
- `make_unique` method is added and used in `replace_variables` method to replace all variables inside atom by unique ones
- in `GroundingSpace::single_query` function `replace_variables` method is used to instantiate new unique instance of the atom before matching
- function `atoms_are_equivalent` is added to check that first atom can be converted to other by renaming the variables
- `atoms_are_equivalent` is used to check tests asserts
- fixed issue with running complex queries where variable is used in different subqueries
- removed old atom matching code, fixed corresponding unit tests

### DIFF
--- a/lib/src/atom/matcher.rs
+++ b/lib/src/atom/matcher.rs
@@ -302,11 +302,11 @@ pub fn apply_bindings_to_bindings(from: &Bindings, to: &Bindings) -> Result<Bind
         // Check that variable is not expressed via itself, if so it is
         // a task for unification not for matching
         if !matches!(applied, Atom::Variable(_)) && atom_contains_variable(&applied, key) {
-            log::trace!("apply_bindings_to_bindings: self expression: key: {}, applied: {}", key, applied);
+            log::trace!("apply_bindings_to_bindings: rejecting binding, variable is expressed via itself: key: {}, applied: {}", key, applied);
             return Err(())
         }
         if !res.check_and_insert_binding(key, &applied) {
-            log::trace!("apply_bindings_to_bindings: cannot insert: ({}, {}) into {}", res, key, value);
+            log::trace!("apply_bindings_to_bindings: rejecting binding, new value is not equal to previous one: ({}, {}) into {}", res, key, value);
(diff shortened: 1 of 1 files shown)


lib/src/atom/matcher.rs
```diff
@@ -336,15 +336,18 @@ fn atom_contains_variable(atom: &Atom, var: &VariableAtom) -> bool {
...
 
 pub fn apply_bindings_to_bindings(from: &Bindings, to: &Bindings) -> Result<Bindings, ()> {
+    log::trace!("apply_bindings_to_bindings: from: {}, to: {}", from, to);
     let mut res = Bindings::new();
     for (key, value) in to {
         let applied = apply_bindings_to_atom(value, from);
         // Check that variable is not expressed via itself, if so it is
         // a task for unification not for matching
         if !matches!(applied, Atom::Variable(_)) && atom_contains_variable(&applied, key) {
+            log::trace!("apply_bindings_to_bindings: self expression: key: {}, applied: {}", key, applied);
```
noskill, line 346: better log message would be "rejecting binding, variable is expressed via itself"

vsbogd: Fixed

Necr0x0Der (approved):

noskill (approved):

PR #105, opened 2022-06-09
support ; and ;; style comments

noskill:
; skips one line
;; skips s-expression


vsbogd:
It doesn't look like this code works. Two items here:
1. Let's agree if we need this for now, I don't have objections, but I would also listen to @Necr0x0Der;
2. If we agree it should be implemented, please add unit tests on this functionality.

Necr0x0Der:
I'd find functionality of commenting both single lines and expressions convenient. But indeed, it should be covered by unit tests. I also wonder what will happen when `;` or `;;` will appear inside expressions. This would also be nice to cover in unit tests. Also, I'd remove processing of `;` in Python, but this can be done separately from this PR.

python/tests/test_run_metta.py
```diff
@@ -48,3 +48,33 @@ def test_list_concatenation(self):
         result = MeTTa().run(program)
         self.assertEqual('[[(Cons a1 (Cons a2 (Cons b1 (Cons b2 Nil))))]]', repr(result))
 
+    def test_comments(self):
```
vsbogd, line 51: As changes in functionality are made in Rust code, we need adding Rust unit tests to check them. Otherwise other platforms are not covered. Could you migrate tests to Rust please.

python/tests/metta.py
```diff
@@ -1,10 +1,13 @@
 import os
 import sys
+from hyperonpy import init_logger
+
 
 from common import MeTTa
 
 if __name__ == "__main__":
     os.system('clear')
+    init_logger()
```
vsbogd, line 10: This is not needed anymore after #80 is merged.

lib/src/metta/text.rs
```diff
@@ -50,6 +50,26 @@ impl<'a> SExprParser<'a> {
...
+                    self.it.next();
+                    let next = self.it.peek();
+                    match next {
+                        Some(n) => {
+                            if *n == c1 {
+                                // found ;; parse sexpr and skip it
+                                self.skip_sexpr(tokenizer);
+                            } else {
+                                // it's just ; iterate until \n
+                                self.skip_line()
```
vsbogd, line 65: Here looks like token after `;` if it is not `;` again is skipped inside `self.skip_line()`. On `;\n` sequence it will skip the next line. Taking it into account I would propose removing `self.it.next();` from code of `skip_sexpr` and `skip_line`. And add it separately into the first branch of the `if`.

Making code more compact the final version:
```rust
';' => {
  self.it.next();
  match self.it.peek() {
    Some(c) => {
      if c == ';' {
        self.it.next();
        self.skip_sexpr();
      } else {
        self.skip_line();
      }
    None => return None;
  }
},

**Review by vsbogd — CHANGES_REQUESTED** [review]

File: python/tests/metta.py (hunk 2 above)
Comment (noskill) [reply python/tests/metta.py:10]: It's needed on my machine. I don't have logs without this line.

Comment (vsbogd) [reply python/tests/metta.py:10]: It is strange, could you please check if https://github.com/trueagi-io/hyperon-experimental/blob/067d627e708e9841f1cda9dcc2df479d809642eb/lib/src/lib.rs#L14 is called when `hyperon` library is loaded from Python in your environment?
For example you can run Python in GDB like this:
```
$ gdb --args python -c 'import hyperon'
(gdb) break on_load
(gdb) r
```

**Comment by noskill** [comment]
It's not getting called:

(gdb) break on_load
Function "on_load" not defined.
Make breakpoint pending on future shared library load? (y or [n]) y
Breakpoint 1 (on_load) pending.
(gdb) r
Starting program: /usr/bin/python -c import\ hyperon
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
[Inferior 1 (process 32147) exited normally]

пт, 10 июн. 2022 г. в 14:09, Vitaly Bogdanov ***@***.***>:

File: python/tests/metta.py (hunk 2 above)
Comment (vsbogd) [reply python/tests/metta.py:10]: It stopped working starting from `rustc` version `1.60`.

Comment (vsbogd) [reply python/tests/metta.py:10]: But it will be fixed in 1.63. You can roll back rust toolchain to `1.59` to make it work

Comment (vsbogd) [reply python/tests/metta.py:10]: Fixed it in #108

File: lib/src/metta/text.rs [hunk 4]
```diff
@@ -58,23 +78,62 @@ impl<'a> SExprParser<'a> {
...
+                }
+            }
+        }
+    }
+
+    fn skip_sexpr(&mut self, tokenizer: &Tokenizer) {
+        self.it.next();
+        // parse s-expression and skip it
+        let res: Option<Atom> = self.parse(tokenizer);
+        match res {
```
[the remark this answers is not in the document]
noskill: it looks more clean that way - when code has a name like skip_sexpr vs unnamed piece of code

lib/src/metta/text.rs
noskill: yes, i added test to check if it's handled correctly

python/tests/metta.py
```diff
@@ -1,5 +1,7 @@
 import os
 import sys
+from hyperonpy import init_logger
+
```
vsbogd, line 4:
```suggestion
```

lib/src/metta/text.rs
```diff
@@ -242,4 +286,58 @@ mod tests {
...
+    #[test]
+    fn test_parser_comments(){
+        println!("test_parser_comments");
+        let mut i = 0;
+        let program = " (a ; 4)
+                  5)
+                !(match &self (a $W)   $W)";
+        let expected = vec!("(a 5)", "!", "(match &self (a $W) $W)");
+        println!("running test {}", i); i+=1;
+        parse_compare(program, &expected);
```
vsbogd, line 299: Few comments here:
1. Each test should be as simple as it is possible while demonstrating the feature works. Thus `!(match &self (a $W) $W)` is not needed at first case for instance
2. Each test case ideally should be in separate test
3. Using custom results comparison procedure like `parse_compare` makes tests results less clear, in cases below using `assert_eq!` is enough 
4. Standard syntax for `vec!` macros is `vec![]` not `vec!()`

```rust
    fn parse(text: &str) -> Vec<Atom> {
        let tokenizer = Tokenizer::new();
        let mut parser = SExprParser::new(program);
        let mut result = Vec::new();
        while let Some(atom) = parser.parse(&tokenizer) {
            result.push(atom);
        }
        result
    }

    #[test]
    fn test_comment_out_tail_of_expression(){
        let atoms = parse("(a ; 4)
                  5)");
        assert_eq!(atoms, vec![expr!("a", "5")]);
    }

    #[test]
    fn test_comment_out_next_expression(){
        let atoms = parse("a ;;(a  1) b");
        parse_compare(atoms, vec![sym!("a"), sym!("b")]);
    }

    #[test]
    fn test_comment_out_multiline_expression(){
        let atoms = parse(";; (a
                        1)");
        assert_eq!(atoms, vec![]);
    }

    #[test]
    fn test_comment_out_empty_line(){
        let atoms = parse("(a 1);
            b
         ");
        assert_eq!(atoms, vec![expr!("a", "1"), sym!("b")]);
    }
```

noskill:
How to place a comment inside of s-expression then?

вт, 14 июн. 2022 г., 14:21 Vitaly Bogdanov ***@***.***>:

vsbogd:
> How to place a comment inside of s-expression then?

I don't think we need them.

noskill:
Who are 'we'? In every non-toy programming language you can place comments
inside functions, including scheme family
https://www.gnu.org/software/guile/manual/html_node/Comments.html . If you
need a semicolon for example for "or" operator we can choose different
symbol for comments like # and use it with '#' quotation if we need this
symbol inside of s-expression.

вт, 14 июн. 2022 г. в 16:12, Vitaly Bogdanov ***@***.***>:

noskill:
I agree about this issue. Perhaps a better alternative is to introduce
block comments, like in c++ /*   */ or indeed to process ;; on python level.

вт, 14 июн. 2022 г. в 14:28, Vitaly Bogdanov ***@***.***>:

vsbogd:
I don't think that comments inside expressions are essential for any programming language.
I think that ability to use `;` as a symbol inside expression is more valuable than using it for commenting out half of `s-expr`. At least I would ask @Necr0x0Der opinion as well as he raised this question.

vsbogd (approved):

PR #108, opened 2022-06-10
Fix toolchain version to 1.59

vsbogd:
Starting from 1.60 `ctor` library doesn't work properly. This is expected to be fixed in 1.62 or later.
See https://github.com/rust-lang/rust/issues/47384 and https://github.com/mmastrac/rust-ctor/issues/27.


noskill:
the issue https://github.com/rust-lang/rust/issues/47384 is closed.. so does the code work with rustc from master branch?

vsbogd:
Yes, but there is no stable version of toolchain which includes this change.

noskill (approved):

Issue #107, opened 2022-06-10
Replace nose testing framework in Python hyperon library


vsbogd:
PyPi version of the Nosetests test runner is not compatible with Python 3. Moreover Nosetests is not maintained anymore (see https://github.com/nose-devs/nose/issues/1099 and #106). This issues is to migrae Python part of the Hyperon project to the new test runner. The possible options are:
- [`unittest`](https://docs.python.org/3/library/unittest.html#command-line-interface);
- [`nose2`](https://docs.nose2.io);
- [`pytest`](https://docs.pytest.org).

PR #109, opened 2022-06-10
remove nosetest

noskill:
fixes https://github.com/trueagi-io/hyperon-experimental/issues/107


python/setup.py
```diff
@@ -11,9 +12,11 @@ def finalize_options(self):
 
     def run(self):
         import sys, subprocess
-        raise SystemExit(subprocess.run([ sys.executable, "-m", "nose",
-            "--with-coverage", "--cover-erase", "--cover-package=hyperon",
-            "--cover-html", "./tests" ]))
+        result = subprocess.run(shlex.split("coverage run --source hyperon -m unittest discover"), check=False,
```
vsbogd, line 15:
```suggestion
        result = subprocess.run(shlex.split("python -m coverage run --source hyperon -m unittest discover"), check=False,
```

vsbogd:
I tried it looks like it gets coverage of the part of the tests only. To me it shows coverage 31% when previous version shows 98%.

PR #119, opened 2022-06-24
get rid of quote hack in match_op

Necr0x0Der:


Necr0x0Der:
The hack with `q` in `match_op` is not needed anymore. The only test, which required it, is working now without `q`. If new issues are encountered, they should be dealt with using meta-types.

python/tests/scripts/d4_selfmod.metta
```diff
@@ -43,7 +46,7 @@
    (nop (change-state greeted True)))
 
 ;!(answer (listen (Hello)))
-!(answer (listen (My name is "Alexey")))
+!(answer (listen (My name is "Alexey"))) ;doesn't work correctly now due to caching?
```
vsbogd, line 49: Yes, probably, I made some simple check for grounded sub-expression, but it is not enough. Anything grounded in calculation trace should prevent caching.

vsbogd (approved):

PR #126, opened 2022-07-02
more tests on matching symbolic expressions

Necr0x0Der:


Necr0x0Der:
`b2_backchain.metta` script has the second part not covered as unit tests, which might be connected to #127 - check both `!(if! (deduce (Evaluation (mortal $x))) $x)` and `!(explain (Evaluation (mortal $x)))`

vsbogd (approved):

PR #139, opened 2022-07-14
Fix expressions matching

### LINKED ISSUE #127: Incomplete substitutions in (composite) queries
For the program
```
(implies (Frog $x) (Green $x))
(implies (Frog $x) (Eats-flies $x))
(Frog Sam)
(Robot Sophia)
```
the result of the query
```
!(match &self
  (, (Frog $x)
     (implies (Frog $x) $y))
  $y)
```
is correct (`[(Green Sam), (Eats-flies Sam)]`).
However, if the query expressions are swapped
```
!(match &self
  (, (implies (Frog $x) $y)
     (Frog $x))
  $y)
```
then the output is `[(Green $x-23), (Eats-flies $x-24)]`.

Also, the query
```
!(match &self
    (, (implies ($P $x) (Green $x))
        (implies ($P $x) (Eats-flies $x)))
    ($x is definitely $P))
```
returns `[($x-73 is definitely Frog)]`, which is not incorrect, because `$x` remains unbound, but could the final result of match restore original variables? There are tricky situations, e.g.
```
(equals $x $x)
!(match &self
    (equals $y $z)
    ($y $z))
```
should not return `($y $z)`.  Ideally, it should return `($y $y)`. Now, it returns `($x-96 $x-96)`, which is ok, but not always convenient. Possibly, this is connected with the main issue, but if not, we can leave it as is for now.

vsbogd:
Fixes #127
This PR reworks pattern matcher to handle different tricky matches which were not processed before. I didn't implemented logic to substitute variables by their previous names because of reasons explained here: https://github.com/trueagi-io/hyperon-experimental/issues/127#issuecomment-1181645146
Particular cases which are handled now can be seen at the tests section of the `matcher.rs`.

### DIFF
--- a/lib/src/atom/matcher.rs
+++ b/lib/src/atom/matcher.rs
@@ -201,6 +201,20 @@ impl VariableMatch {
         }
     }
 
+    fn match_values(&self, current: &Atom, value: &Atom) -> Option<VariableMatch> {
+        let sub_match: Vec<VariableMatch> =
+            match_atoms_recursively(current, value).collect();
+        assert!(sub_match.len() <= 1, concat!(
+                "Case when sub_match returns more than ",
+                "one matcher because match_() is overloaded ",
+                "inside grounded atom is not implemented yet"));
+        if sub_match.len() == 1 {
+            VariableMatch::merge(self, &sub_match[0])
+        } else {
+            None
+        }
+    }
+
     fn merge_var_sets(&mut self, a_var_set: u32, b_var_set: u32) -> bool {
         fn move_set(vars: &mut HashMap<VariableAtom, u32>, from: u32, to: u32) {
             vars.iter_mut().for_each(|(_var, set)| {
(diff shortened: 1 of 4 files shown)


vsbogd:
Cases which are implemented by this PR:

```rust
    #[test]
    fn match_spread_value_via_pattern_variable() {
        assert_match(
            expr!("v" a a),
            expr!( x  x y),
            vec![bind!{x: sym!("v"), y: sym!("v")}]);
    }

    #[test]
    fn match_spread_value_via_pattern_variable_reverse_order() {
        assert_match(
            expr!(a "v" a),
            expr!(x  x  y),
            vec![bind!{x: sym!("v"), y: sym!("v")}]);
    }

    #[test]
    fn match_variable_via_itself() {
        assert_match(
            expr!(a  a ),
            expr!(x (x)),
            vec![]);
    }

    #[test]
    fn match_stub_variable_is_unique() {
        let data =    expr!(a a);
        let pattern = expr!(x y);
        let x = VariableAtom::new("x");

        let bindings_a = data.match_(&pattern).next().unwrap();
        let bindings_b = data.match_(&pattern).next().unwrap();

        assert_ne!(bindings_a.get(&x), bindings_b.get(&x));
    }

    #[test]
    fn match_equality_of_pattern_variables_inside_expression() {
        assert_match(
            expr!( a    a   a),
            expr!((x) ("v") y),
            vec![bind!{x: expr!("v"), y: expr!(("v"))}]);
    }

    #[test]
    fn match_equality_of_data_variables_inside_expression() {
        assert_match(
            expr!((a) ("v") a),
            expr!( x    x   y),
            vec![bind!{x: expr!(("v")), y: expr!("v")}]);
    }

    #[test]
    fn match_match_values_when_merging_two_variable_sets() {
        assert_match(
            expr!((a)  b   b a),
            expr!( x ("v") x y),
            vec![bind!{x: expr!(("v")), y: expr!("v")}]);
    }
```

Necr0x0Der (approved):

PR #142, opened 2022-07-15
Atom argument for match + let* to fix test

### LINKED ISSUE #96: Pattern argument of match is reduced before calling while it can return result without being reduced
```
(:= (Green Sam) True)
(= (Green Tom) False)
!(println! (match &self (:= (Green $who) True) ($who is really green)))
```
The second line *(= (Green Tom) False)* changes output  
Without it program prints "(Sam is really green)"
with it matching fails

Necr0x0Der:
Should fix #96

### DIFF
--- a/python/tests/common.py
+++ b/python/tests/common.py
@@ -10,6 +10,16 @@ def let_op(pattern, atom, templ):
     space.add_atom(atom)
     return space.subst(pattern, templ)
 
+def letrec_op(subs, body):
+    # just unsugaring `let*`` into `let` substitution by substitution
+    subs = subs.get_children()
+    if len(subs) == 0:
+        return [body]
+    next_sub = subs[0].get_children()
+    if len(subs) == 1:
+        return [E(letAtom, next_sub[0], next_sub[1], body)]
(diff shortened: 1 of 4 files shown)


python/tests/common.py
```diff
@@ -10,6 +10,16 @@ def let_op(pattern, atom, templ):
...
 
+def letrec_op(subs, body):
+    # just unsugaring `let*`` into `let` substitution by substitution
+    subs = subs.get_children()
+    if len(subs) == 0:
+        return [body]
+    next_sub = subs[0].get_children()
+    if len(subs) == 1:
+        return [E(letAtom, next_sub[0], next_sub[1], body)]
+    return [E(letAtom, next_sub[0], next_sub[1], E(letrecAtom, E(*subs[1:]), body))]
```
noskill, line 21: will it work with 3 or more subexpressions? There is no test for it

Necr0x0Der: Yes, it will. This PR is more about `match` than `let*`. It would be better to introduce `let*` in one PR with unit tests for it, and then to change `match` type in another PR, but `match` is more prioritized.

vsbogd (approved):

noskill (approved):

vsbogd (approved):

PR #149, opened 2022-08-01
ubuntu 22 docker image

noskill:

PR #148, opened 2022-08-01
Move docker instructions to the top of readme

noskill:

PR #150, opened 2022-08-02
Fix README.md and docker to provide ready to use environment

vsbogd:
PR to rework README.md and Dockerfile according to suggestions in #148 and #149. I removed `make install` completely and prepared `Dockerfile` with ready-to-use environment. README.md is changed accordingly.


README.md
```diff
@@ -58,26 +79,19 @@ make
...
-make install
-```
-Note: after installation you will need re-install Python library to apply
-changes in Python files.
-
 # Running Python and MeTTa examples from command line
 
-In order to run examples you need either install Python library or
-add Python libraries into the `PYTHONPATH` after compilation:
+In order to run examples you need adding Python libraries into the `PYTHONPATH`
```
noskill, line 84: `need to add`

noskill (changes requested):

noskill (approved):

PR #153, opened 2022-08-04
flatten the list of results in import_op

### LINKED ISSUE #152: Nested imports raises exception
# Problem
Importing a metta file that itself imports another metta file raises the exception
```
[(Error (import! &self B.metta) Exception caught:
AttributeError: 'list' object has no attribute 'catom')]
```

# How to reproduce
1. Create empty `A.metta` file
2. Create `B.metta` file with content `!(import! &self "A.metta")`
3. Create `C.metta` file with content `!(import! &self "B.metta")`
4. Running `metta.py C.metta` raises the exception, while neither `A.metta` nor `B.metta` does.

Necr0x0Der:
Should fix #152


ngeiswei (approved):
It works for me.

PR #155, opened 2022-08-04
Fix type check of parameterized types

### LINKED ISSUE #130: MeTTa type checker in Rust cannot match two parameterized types
The following MeTTa code

```scheme
;; Declaration of List data structure
(: List (-> $a Type))
(: Nil (List $a))
(: Cons (-> $a (List $a) (List $a)))

!(println! ("========= Test equality ========="))
!(println! (== True True))
!(println! (== True False))
!(println! (== Nil Nil))
!(println! (== 1 1))
!(println! (== (Cons 1 Nil) (Cons 1 Nil)))
```

prints

```
("========= Test equality =========")
True
False
True
[]
[]
[]
[]
[]
[]
```

If the type declarations of `List`, `Nil` and `Cons` are removed, then it prints

```
("========= Test equality =========")
True
False
(== Nil Nil)
True
(== (Cons 1 Nil) (Cons 1 Nil))
[]
[]
[]
[]
[]
[]
```

What does that mean?

Generally speaking should I overload `==` for non primary data types?

Last and least, what do the `[]`s mean?

vsbogd:
Fixes #130. Add unit test on type check of parameterized types. Fix matching variables.


Necr0x0Der (approved):

ngeiswei (approved):
Works for me, awesome!

Issue #147, opened 2022-07-22
Symbol (operations) overloading


Necr0x0Der:
There are different cases, when we want some symbols (especially, functional) to be overloaded, that is, to have different behavior for different types. Two prominent examples are comparisons and `fmap`. Let us consider
```
(: Maybe (-> $a Type))
(: Nothing (Maybe $a))
(: Just (-> $a (Maybe $a)))
```
We may want to have `fmap : ($a -> $b) -> $F $a -> $F $b` for any functor `$F`.
```
(= (fmap $f (Just $x)) (Just ($f $x)))
(= (fmap $f Nothing) Nothing)
```
We can have another definition for `List` or another container. Thus, more equalities to the same `fmap` will be added. MeTTa supports this, and one may hope that wrong equalities will not be applied due to pattern-matching against particular constructors (e.g. `(Cons 1 Nil)` will not be confused with `(Just 1)` or `Nothing`), but one may want to define `(= (fmap $f $any) Nothing)`, but only for `(: $any (Maybe ()))` or anything else weird (e.g. to define a particular `fmap` for particular `$a`). Thus, we cannot be sure that a desirable definition of `fmap` will be selected due to constructors. In any case, there is a natural desire to group equalities for `fmap` by implementations per type. This can be done explicitly, e.g.
```
(: fmap-maybe (-> (-> $a $b) (Maybe $a) (Maybe $b))
(= (fmap-maybe $f (Just $x)) (Just ($f $x)))
(= (fmap-maybe $f Nothing) Nothing)
(= (fmap $f $xs) (fmap-maybe $f $xs))
```
Thus, if we pass `(fmap f (Cons 1 Nil))` it will not go to `fmap-maybe`, because types will not match. Another option would be not to introduce `fmap-maybe`, but to have means to indicate that `fmap` in certain equalities is restricted to certain types. One may propose different syntactic representation for this, but the fundamental question here is whether we want symbol to have one global type or each instance of a symbol may have its own (specialized) type, which is a different issue to discuss, but which may influence our approach to symbol overloading (well, types of symbol instances is also about symbol overloading, but cases can be different).
If we consider comparisons, they will have similar issues. For example,
```
(= (< (Just $x) (Just $y)) (< $x $y))
(= (< (Just $x) Nothing) False)
(= (< Nothing (Just $x)) True)
```

[...]

PR #156, opened 2022-08-05
modify == to handle symbolic expressions

Necr0x0Der:


Necr0x0Der:
I've changed `==` implementation to handle symbolic expressions. The reason is that it is difficult to implement `eq` in pure MeTTa to return `False`, while it is needed in practice. The downside is that `==` behaves differently from, say, `<` or `>` in that it will not remain unreduced if its arguments are not reduced. It will always return `True` or `False`. This behavior makes sense, but it may cause some confusion.

vsbogd (approved):

ngeiswei:
Thanks @Necr0x0Der, that's quite useful.

I think supporting `<` and `>` for any expression could also be useful, for instance to define a sorted list of expressions, but that can probably be done in pure MeTTa.  Or, to be maximally useful, it would probably need to be overwritable for handling cases where expressions require their own order, such as `(: Z Nat), (: S (-> Nat Nat))`, etc.

Necr0x0Der:
@ngeiswei , well, I had a thought on this. The problem is that `==` returning always `True` or `False` makes some sense (e.g., even if its arguments are of different type, then the result is just `False`). At the same time, `<` or `>` should not always return `True` or `False`. Well, with `(-> $t $t Bool)` type signature, the case of different argument types can be handled (as a typing error), but still even for the same `$t`, binary comparison operators will not always make sense. (well, I agree that equality relations are not necessarily binary as well, but order-related comparisons are way much choosy). Thus, "supporting `<` and `>` for any expression` seems questionable to me, they "needed to be overridable" sounds right. The latter is exactly a usecase for #147

ngeiswei:
I wonder if we could have a "syntactic less than" as we do in Atomese.  It wouldn't need to use the `<` character though.

Necr0x0Der:
@ngeiswei , could you please elaborate on this? I'm not sure what "syntactic less than" means.

ngeiswei:
@Necr0x0Der, specifically I mean

https://github.com/opencog/atomspace/blob/1fd6118ecfd0a4781cd3bb765ca521b03aaf719e/opencog/atoms/base/Atom.h#L534

implemented here

https://github.com/opencog/atomspace/blob/1fd6118ecfd0a4781cd3bb765ca521b03aaf719e/opencog/atoms/base/Node.cc#L116-L131
https://github.com/opencog/atomspace/blob/1fd6118ecfd0a4781cd3bb765ca521b03aaf719e/opencog/atoms/base/Link.cc#L130-L165

and else where.

Well, this one is more than just "syntactic order".  First it compares the hash keys, and only resort to "syntactic order" if the keys are the same.  I'm sure there's a colloquial term for it but I don't know it, it might be something like "prefix tree lexicographic order", or such.

Necr0x0Der:
> I wonder if we could have a "syntactic less than" as we do in Atomese. It wouldn't need to use the `<` character though.

> First it compares the hash keys, and only resorts to "syntactic order" if the keys are the same. 

After Vitaly introduced indexing, something similar might be possible to implement. However, I wonder if it would be correct to have this comparison dependent on the way of how indexing is done. Or should it be invariant? Maybe, knowing the purpose of "syntactic less than" would be useful.

ngeiswei:
A potential use of a "syntactic less than" operator would be if we want to write an algorithm in metta (or python or even rust) code that needs an ordered set or an order map of atoms, for instance.

PR #172, opened 2022-08-16
Fix matching meta-types in arguments of functions

vsbogd:


lib/src/metta/types.rs
```diff
@@ -631,4 +640,157 @@ mod tests {
...
+            (: atomF (-> Atom R))
+            (: exprF (-> Expression R))
+            (: gndF (-> Grounded R))
+            (: symF (-> Symbol R))
+            (: varF (-> Variable R))
+        ");
+
+        assert!(check_type(&space, &atom("(aF a)"), type_r));
+        assert!(!check_type(&space, &atom("(aF b)"), type_r));
+        // TODO: (aF b) is incorrectly typed, but it is an Atom and check_type
```
vsbogd, line 664: @Necr0x0Der please look at this and below examples marked by TODO. Should we handle them differently?

Necr0x0Der (approved):

lib/src/metta/types.rs
Necr0x0Der: We need some real use cases in order to understand if this is the intended behavior or not. Let us keep it as is for now.

PR #189, opened 2022-10-31
Proposal for collaboration

h555yand:
Hi, my name is Aleksei. [About me.](https://www.linkedin.com/in/alexey-isaev-820b4035/)
I'm very curious in General Artificial intelligence researches and follow OpenCog projects and events.
I’m sure that OpenCog team is doing great job to develop general AI and I would be happy to help.

I have ideas how to develop Artificial Intelligence as a child from birth to youth and I believe that Hyperon is the best framework for trying to implement it.  [Details here](https://github.com/h555yand/hyperon-experimental/blob/main/python/tests/child_ai/README.md)
I’ve created a simplified Python example of sensory perception exercise for child (Catching the toy from the hands of an adult) in  [child_ai.py](https://github.com/h555yand/hyperon-experimental/blob/c13628b3447e7d1ff0704cba6649abb39bb71923/python/tests/child_ai/child_ai.py) file.
To use Hyperon  Space / Atoms functionality I’ve created several functions in Rust (source code in [graph_kb.rs](https://github.com/h555yand/hyperon-experimental/blob/c13628b3447e7d1ff0704cba6649abb39bb71923/lib/src/child_ai/graph_kb.rs) file). They duplicate similar functions in Python child_ai.py and have the same names. At the moment there is no integration between Python and Rust for these functions. 

I understand that general AI is a very long way and there are many another ideas.
I don’t insist to implement my idea and ready to help with other tasks that you think have more potential and priority.

Now I’m looking for new opportunities as my employer SAP winds down business in Russia. If you have any open position I would be happy to apply and support your projects in any format.

Looking forward for your feedback.

Best regards, Aleksei


Necr0x0Der:
@h555yand , please, consider raising this PR in https://github.com/trueagi-io/metta-examples/

vsbogd:
Hi Aleksei, thank you for your interest and for sharing your code.

As far as I understand you have implemented the idea in Python and a Rust part is to demonstrate how it could be implemented in the core library. And as far as I see the Python code don't use Metta, so it is an independent implementation.

I would also suggest raising PR to the https://github.com/trueagi-io/metta-examples/ and then it can be discussed how your idea can be implemented using Metta or what can be added to Metta to make implementation possible.

> Now I’m looking for new opportunities as my employer SAP winds down business in Russia. If you have any open position I would be happy to apply and support your projects in any format.

I will write you an email on this.

h555yand:
Hi Vitaly, thank you very much for the quick response.

Your understanding is correct. Python code is independent at the moment and Rust functions are commented and not used there.
OK, I will create a new PR in https://github.com/trueagi-io/metta-examples/

Best regards, Aleksei

h555yand:
Hi Alexey, Vitaly,
As agreed I’ve made updates for Expression Atoms for Graph knowledge base task.
The task is to demonstrate possible approaches how to use Graph knowledge base data with MeTTa.
Typical  Graph knowledge base consists of Edges with Nodes, and they can have features to store information of different types (sparse distributed vectors for code and weights, IDs, Names, Types etc).
Use case is to enter Edges with Nodes in MeTTa and then parse / interpret them in Graph knowledge base format that used in Child_AI idea https://github.com/h555yand/hyperon-experimental/blob/main/python/tests/child_ai/README.md

For Edges with Nodes the expression is used “((LinksEdge Link1) (MemoryNode Node1) (MemoryNode Node2))”.
This is similar to OpenCog classic format expression "(EvaluationLink(PredicateNode eats) (ListLink (ConceptNode Sam)(ConceptNode flies)))"
In future we can change it to Type format like (: LINK Type).

I’ve modified the Rust file [graph_kb.rs](https://github.com/h555yand/hyperon-experimental/blob/dd1b7b02f58ce1a072ab0f3834c6032b4fc5ff18/lib/src/child_ai/graph_kb.rs) and have not changed any Python code. As I understand there are parse / interpret functions that are in the process of transferring from Python into Rust.
I’ve created several new functions ( GraphKB::parse_graph and others) in addition to the existing SexprParser / interpreter::interpret functions. In future, if we decide to include such functionality in MeTTa, we need to adapt the existing functions.

The existing Rust functions parse / interpret the expression “((LinksEdge Link1) (MemoryNode Node1) (MemoryNode Node2))” as Expression Atom with children Symbol Atoms. To use it with Graph knowledge base we need to convert them into Expression Atom with children Grounded Atoms (Link Edge, Source and Destination Nodes) . 
The first step is to try to find Grounded Atoms by Symbol Link and Node Names. If no success, we create new edge and nodes, using Name from MeTTa Expression and default values for other fields.

If no need in interpretation we just add Expression Atom in Space.
To use interpretation we need to add special functions to work with Graph, similar to existing functions in Python self.add_atom(r"get-type", newGetAtomTypeAtom(self)).
We can return to MeTTa requested  information from Grounded Atoms in addition to Symbol Names and execute requested actions (to find linked nodes for example).

There are two test cases: (1) when Grounded Atoms (Link1, Node1, Node2) exist,
(2) when Grounded Atoms (Link11, Node11, Node22) don't exist.

Best regards, Aleksei.

vsbogd:
Hi Alexei,

When I was talking about more tight integration with Metta I thought about something like the following program:
```
(bind! &memory (new-space))

(= (predict-next obj)
    (select-max-weight
        (collapse (match &memory (next-link obj $next $w) ($w $next)))
    )
)

(= (add-next-observation $prev $next)
    (case (match &memory (next-link $prev $next $w) $w)
        (%void% (call:add &memory (next-link $prev $next initial-weight)))
        ($w (call:replace &memory
            (next-link $prev $next $w)
            (next-link $prev $next (increment-weight $w))))
    )
)

!(add-next-observation "a" "b") ; adds (next-link "a" "b" <initial-weight>) into &memory kb
!(add-next-observation "a" "a") ; adds (next-link "a" "a" <initial-weight>) into &memory kb
!(add-next-observation "a" "b") ; increments weight of the (next-link "a" "b" <initial-weight>)
!(add-next-observation "a" "b") ; increments weight of the (next-link "a" "b" <initial-weight>)

!(predict-next "a") ; returns prediction with a maximum weight "b"
```

I wrote it to demonstrate how one can implement prediction of the next object (one of the examples I found in README.md file you provided). Basically we have `&memory` database which contains information about next objects and weights. Each possible next object memoized from experience is represented by `(next-link <prev-object> <next-object> <weight>)` where `next-link` is a symbol, `<prev-object>` and `<next-object>` are some atoms, and `<weight>` is grounded atom which contains instance of `numpy` array or any other convenient low-level structure.

I have added `predict-next` and `add-next-observation` functions to demonstrate how one can implement such functionality in Metta. Of cause operations on grounded weights cannot be implemented in pure Metta so one need implement few grounded functions in addition:
 - `select-max-weight` which gets list of pairs `(<weight> <object>)` and returns the object with maximum weight;
 - `initial-weight` which is a symbol to assign initial weight to the link;
 - `increment-weight` which is a function to increment weight when some instance of next link is created from experience.

The purpose of the example is to demonstrate how one of your examples could be implemented in Metta like style with minimal necessary grounded objects and operations. Here only operations on weights should be implemented manually. All other functions (`bind!`, `new-space`, `match`, `call`, `case`, `collapse`) are part of the Metta library. And operations on weights can be implemented even in Python.

I am not sure actually how far this example from what you have implemented in the PR though. May be it is simpler to discuss it on a call and then continue discussing here.

vsbogd:
One can also write this code in pure Python if it is more convenient like code below. But still there is no need to create custom Link and Node instance, one can use built-in Metta expressions and symbols:
```py
from hyperon import *

memory = GroundingSpace()

class Weight:

    def __init__(self):
        self.w = 1

    def __lt__(self, other):
        return self.w < other.w

    def inc(self):
        self.w += 1

def next_link(prev, next, weight):
    return E(S('next-link'), prev, next, weight)

def predict_next(obj):
    candidates = memory.subst(next_link(obj, V('next'), V('w')), E(V('w'), V('next')))
    if len(candidates) == 0:
        return None
    max_weight_expr = max(candidates,
                          key=lambda e: e.get_children()[0].get_object().value)
    return max_weight_expr.get_children()[1]

def add_next_observation(prev, next):
    existing = memory.query(next_link(prev, next, V('w')))
    if len(existing) == 0:
        memory.add_atom(next_link(prev, next, ValueAtom(Weight()))) 
    else:
        assert(len(existing) == 1)
        weight = existing[0]['w']
        new_weight = weight.get_object().value
        new_weight.inc()
        memory.replace_atom(next_link(prev, next, weight), next_link(prev, next, ValueAtom(new_weight)))

add_next_observation(S('a'), S('b')) # adds (next-link "a" "b" <initial-weight>) into &memory kb
add_next_observation(S('a'), S('a')) # adds (next-link "a" "a" <initial-weight>) into &memory kb
add_next_observation(S('a'), S('b')) # increments weight of the (next-link "a" "b" <initial-weight>)
add_next_observation(S('a'), S('b')) # increments weight of the (next-link "a" "b" <initial-weight>)

print(predict_next(S('a'))) # prints prediction with a maximum weight "b"
```

vsbogd:
@h555yand as we discussed I closing this, please move this PR into https://github.com/trueagi-io/metta-examples/.
It makes sense to write a small README.md with instructions how to run this project before hyperon library is published.

PR #199, opened 2022-11-14
small PR to accept metta files without ./

Necr0x0Der:


Necr0x0Der:
It's convenient while running Metta from terminal and should harm nothing.

vsbogd (approved):

PR #201, opened 2022-11-16
Return BadType when argument of an expression has incorrect type, remove MeTTaPy implementation

vsbogd:
@Necr0x0Der please take a look at changes in [d5_auto_types.metta](https://github.com/trueagi-io/hyperon-experimental/compare/main...vsbogd:metta-repl-in-rust-2?expand=1#diff-9e4aebe5db37a74cb87c6580aa2514ae0fc9380a6f276c8935627cf59263929b) and [python/tests/scripts/e1_kb_write.metta](https://github.com/trueagi-io/hyperon-experimental/compare/main...vsbogd:metta-repl-in-rust-2?expand=1#diff-3fe68ea6a373886149d6c50d49fc82aa925e208539b7fd1fad4a8466d8e8a336)


python/tests/scripts/e1_kb_write.metta
```diff
@@ -10,9 +10,10 @@
 (= (green $x)
    (frog $x))
 
+(: ift (-> %Undefined% Atom Atom))
```
Necr0x0Der, line 13: Doesn't it work with `Bool` instead of `%Undefined%`?

Necr0x0Der (approved):

Necr0x0Der:
Well, it's a little bit sad that only the internal atom with type mismatch is shown in the error message, but it's ok for now. As for "autotypes", it seems that these tests have less sense now, because they don't check auto types, but execute expressions, so these are runtime type checks (and it didn't really work earlier at all). So, the PR looks great, but some some things should be sorted out after it.

python/tests/scripts/e1_kb_write.metta
vsbogd, line 13: Sorry forgot that True is Bool because it is a token.
```suggestion
(: ift (-> Bool Atom Atom))
```

vsbogd:
> Well, it's a little bit sad that only the internal atom with type mismatch is shown in the error message, but it's ok for now.

Yes, unfortunately current logic find the error in context where is no full expression. And at the moment I don't see how to fix it properly without significantly changing the code of interpreter.

PR #202, opened 2022-11-16
car, cdr, cons on expressions

Necr0x0Der:
* `null-atom?` is not added: it is not needed too much, since `== () ...` can be used, and `Bool` is yet to be discussed;
* `expression?` is not added, it can be added later once needed;
* `*-atom` can be renamed to `*-expr` since these are operations on expressions, but `atom` sounds more distinguishable...


lib/src/metta/runner/stdlib.rs
```diff
@@ -283,6 +284,89 @@ impl Grounded for GetAtomsOp {
...
+    }
+}
+
+impl Grounded for ConsAtomOp {
+    fn type_(&self) -> Atom {
+        Atom::expr([ARROW_SYMBOL, ATOM_TYPE_ATOM, ATOM_TYPE_EXPRESSION, ATOM_TYPE_EXPRESSION])
+    }
+
+    fn execute(&self, args: &mut Vec<Atom>) -> Result<Vec<Atom>, ExecError> {
+        let arg_error = || ExecError::from("cons-atom expects two argument: atom and expression");
```
vsbogd, line 356:
```suggestion
        let arg_error = || ExecError::from("cons-atom expects two arguments: atom and expression");
```

vsbogd (approved):

vsbogd:
Removed some parenthesis in `expr!`, please take a look. Otherwise looks good to me.

PR #206, opened 2022-11-18
the first step of refactoring python library after moving runner to rust

Necr0x0Der:


python/hyperon/atoms.py
```diff
@@ -0,0 +1,180 @@
...
+        else:
+            result = self.op(*args)
+            if not isinstance(result, list):
+                raise RuntimeError("Grounded operation `" + self.name + "` should return list")
+            return result
+
+    def __eq__(self, other):
+        return isinstance(other, OperationObject) and self.name == other.name
+
+def type_sugar(type_names):
```
vsbogd, line 166: Minor and I know it is not directly related to the PR description, but I would prefix `type_sugar` and `call_execute_on_grounded_atom` by `_` to remove them from public API of the module.

python/hyperon/runner.py
```diff
@@ -0,0 +1,65 @@
...
+    def __del__(self):
+        hp.metta_free(self.cmetta)
+
+    def space(self):
+        return GroundingSpace._from_cspace(hp.metta_space(self.cmetta))
+
+    def tokenizer(self):
+        return Tokenizer._from_ctokenizer(hp.metta_tokenizer(self.cmetta))
+
+    def add_token(self, regexp, constr):
```
vsbogd, line 24: It probably makes sense to remove `add_token` from `MeTTa` and use `metta.tokenizer().register_token(...)` instead. The similar to `add_atom` we could move it into `Tokenizer` and use via `metta.tokenizer()`.

vsbogd:
BTW, feel free to merge it and make changes in future PRs if you like.

vsbogd (approved):

python/hyperon/atoms.py
Necr0x0Der: `call_execute_on_grounded_atom` is used in `hyperonpy.cpp`, so it should be in the public API

Necr0x0Der: I've change `type_sugar`. It might be used outside, though...

vsbogd: Nothing should prevent usage of `_call_execute_on_grounded_atom` from `hyperonpy.cpp`, it just will say to users that they should not use it directly.

PR #207, opened 2022-11-22
loading metta extensions in python from scripts

Necr0x0Der:
This is the first step for introducing MeTTa extensions in Python in a standardised way loadable from MeTTa scripts. Additional features as well as descriptions will be provided in further PRs


python/tests/test_extend.py
```diff
@@ -0,0 +1,18 @@
...
+
+from hyperon import *
+
+class ExtendTest(unittest.TestCase):
+
+    def test_extend(self):
+        metta = MeTTa()
+        self.assertEqual(
+            metta.run('''
+              !(extend-py! extention)
```
vsbogd, line 11: Two items here:
1. I would suggest getting string with a module name as an argument instead of symbol.
2. Why `extend-py` and not `import-py`?

python/hyperon/ext.py
```diff
@@ -0,0 +1,11 @@
+def add_atoms(func):
```
vsbogd, line 1: `add_atom` can be confused with adding atoms into atomspace. May be some other name, for example: `add_atom_tokens`?

vsbogd: Or may be 'register_atom_tokens` instead of `add_atoms` and `register_tokens` instead of `add_tokens`?

python/tests/test_extend.py
Necr0x0Der: > I would suggest getting string with a module name as an argument instead of symbol.

It works both with strings and symbols. I simply use `repr` if you noticed. The issue was that I first tried to debug this functionality on importing `stdlib`, which introduces strings. For this case, it is not possible to use strings at all. Also, I noticed that `import stdlib` without quotes looks natural in Python. And third, we discussed the possibility that strings will be implemented in Rust, and it is not guaranteed that Python strings will continue working. Strings can be used in this examples (although I don't think it is more useful than using a symbol), but I'm quite strongly against removing support for changing the current behavior with `repr` (unless something equivalent but safer is proposed).

Necr0x0Der: > Why `extend-py` and not `import-py`?

Well, I wanted to emphasize the current difference between importing expressions by `import` and extending MeTTa itself by `extend-py` (which will not just add grounded atoms, but also may tweak the runner behavior itself). However, since we've discussed that `extend-py` will also be able to act as `import` (add expressions to a space, although the question is what space to use and whether `extend-py` will have a space as its argument), then it may make sense to rename it. My conclusion is: if `extend-py` gets an additional argument (space) and its signature will be the same as of `import` (and it will import expressions to the space), then it can be renamed. While the signature remains different, it should not be renamed.

python/hyperon/ext.py
Necr0x0Der: I agree that `add_atom` is somewhat confusing. However, it is not the problem of this PR. It simply uses naming in the runner class. I agree that it makes sense to rename it, but it is used in too many tests, so it is a separate refactoring.

python/tests/test_extend.py
vsbogd: > I simply use repr if you noticed.

Yes, I have noticed and it is the reason why I am asking. Because `repr` allows using anything which is representable as a string instead of module name, and it is not clean from my perspective. After you comment I realized that strings cannot be used because they are supported by stdlib, but then I would use symbols only. But it is my personal preference probably.

vsbogd: > difference between importing expressions by import and extending MeTTa itself by extend-py

I don't see why current PR doesn't include adding atoms into the MeTTa atomspace?

python/hyperon/ext.py
vsbogd: Yes, I agree this is not the problem of this PR. Just thought it is fixed before adding import because I added such comment on previous PR.

Necr0x0Der: 1) It has not yet been implemented (see the initial comment on the PR).
2) It is not clear for me if this should really be done by `ext` module. As I mentioned before, I expect this module to be used to extend `MeTTa` in Python. Importing expressions is already done by `import`. This is not the main functionality of `ext`, and I'm not sure it should be duplicated. Of course, if it appears to be convenient for users, I'll be happy to add some decarators to `ext` to wrap up `import` or to add expressions. Once again, `extend-py` extends the MeTTa runner, so if we include an additional argument to it, I'm not sure if it should be space or runner. Thus, type signatures can remain different for `extend` and `import` even if we add functionality to insert MeTTa code in the runner's space by `extend`. Having similar names (`import!`, `import-py!`) with different type signatures might be confusing. For now, I'm not sure if we should extend only the current runner (but for possibly a different space), or we should extend another runner (adding tokens to a tokenizer shouldn't be global?). I didn't test if another runner with its own tokenizer is created by `import` right now, and how `extend` within `import`'ed file will work...
So, I still prefer to keep different naming for now.

Necr0x0Der: `add_[atom, token]` is renamed

vsbogd:
Ok, now I see that you are trying to separate two things:
- extending the tokenizer
- adding atoms into atomspace
in this context `extend-py` and `import` are different.
I asked about it because I thought that may be it doesn't make sense to introduce two separate operators to extend tokenizer and fill atomspace and do both things in a single operation. My feeling it that most of libraries should require both things and making them by two lines of code can be bothering.

PR #208, opened 2022-11-25
move test_case to Rust

Necr0x0Der:


lib/tests/case.rs
```diff
@@ -0,0 +1,75 @@
+use hyperon::metta::text::SExprParser;
+use hyperon::metta::runner::Metta;
+use hyperon::common::shared::Shared;
+use hyperon::space::grounding::GroundingSpace;
+
+#[test]
+fn test_case_operation() {
```
vsbogd, line 7: Do you mind splitting these long programs on smaller ones and keep assert just after each separate statement? It would be simpler for reading to me at least.

Necr0x0Der: In the case of this test, it makes sense, yes.

vsbogd (approved):

PR #210, opened 2022-11-28
Add a decorator to pass metta to token registration functions

Necr0x0Der:


vsbogd (commented):
I would add the flag `pass_metta` to the `register_tokens` and `register_atoms` instead. Otherwise the incorrect order of decorators or additional decorators in between may break the functionality.

vsbogd (approved):

PR #212, opened 2022-12-01
custom match for space redirecting matching to its content

Necr0x0Der:


Necr0x0Der:
One comment: import of files with grounded symbols doesn't work now, but it seems the problem of `ImportOp`:
https://github.com/trueagi-io/hyperon-experimental/blob/main/lib/src/metta/runner/stdlib.rs#L88
it creates Rust metta runner without inheriting (Python) tokens from the tokenizer of the current runner. So, it is a separate issue.

vsbogd (approved):
I have changed part of implementation. @Necr0x0Der , if you are happy with it lets merge it?

PR #219, opened 2022-12-11
C and Python API to define custom matching for grounded atoms

h555yand:
Hi,
as agreed sending a draft PR.

Best regards,
Aleksei.


c/src/atom.rs
```diff
@@ -272,7 +310,25 @@ impl Grounded for CGrounded {
...
+
+                func(self.get_ptr(), (other as *const Atom).cast::<atom_t>(), CGrounded::match_callback, context);
+
+                let res_vec = unsafe{ &mut *context.cast::<Vec<Bindings>>() };
+                let len_ = res_vec.len();
+                let res_bnd = res_vec.get(len_-1).unwrap();
+                Box::new(std::iter::once(res_bnd.clone()))
+            },
+            None => match_by_equality(self, other)
+        }
```
vsbogd, line 331: 1. `mut` is not needed at line 321, and compiler warns about it
2. Pushing something into the `vec_temp` doesn't make sense, but I believe it is test code, not sure why it is not inside callback or even C caller
3. It also no need to convert the vector again from `context` and returning the results code looks incorrect, what is required is just returning iterator over `vec_temp` content
4. I would also rename `vec_temp` to `results` or something more meaningful

Overall function should look like this:
```Rust
    fn match_(&self, other: &Atom) -> matcher::MatchResultIter {
        match self.api().match_ {
            Some(func) => {
                let mut results: Vec<Bindings> = Vec::new();
                let context: *mut c_void = (&mut results as *mut Vec<Bindings>).cast::<c_void>();
                func(self.get_ptr(), (other as *const Atom).cast::<atom_t>(), CGrounded::match_callback, context);
                Box::new(results.into_iter())
            },
            None => match_by_equality(self, other)
        }
    }
```

lib/src/atom/mod.rs
```diff
@@ -214,6 +214,10 @@ impl VariableAtom {
         Self{ name: name.into(), id: 0 }
     }
 
+    pub fn new_id<T: Into<String>>(name: T, id: usize) -> Self {
```
vsbogd, line 217: Please document the function.

c/src/atom.rs
```diff
@@ -303,3 +359,46 @@ impl Drop for CGrounded {
     }
 }
 
+#[cfg(test)]
+mod tests {
+use std::sync::Arc;
+
+use super::*;
+    #[test]
+    pub fn test_match_callback(){
```
vsbogd, line 368: Doesn't look like unit test - it has no asserts, but probably it is temporary.

python/hyperon/atoms.py
```diff
@@ -99,6 +99,15 @@ def call_execute_on_grounded_atom(gnd, typ, args):
     args = [Atom._from_catom(catom) for catom in args]
     return gnd.execute(*args, res_typ=res_typ)
 
+def call_match_on_grounded_atom(gnd, typ):
+    # placeholder, please implement your custom functionaluty here
```
vsbogd, line 103: As with execute it should call some custom match on grounded atom.

python/hyperonpy.cpp
```diff
@@ -5,14 +5,15 @@
 
 namespace py = pybind11;
 
-template<class T, size_t N>
+template <class T, size_t N>
```
vsbogd, line 8: Please remove space and code formatting changes as they make reviewing code difficult.

vsbogd:
Also it throws segfault when I am trying to run Python tests. Probably the reason is inside `hyperonpy.cpp` file.
Did you look at it?

h555yand:
Hi,
what Python test throws error? I will check on my side.

vsbogd:
Test output on my machine:
```
63% tests passed, 6 tests failed out of 16

Total Test time (real) =   1.55 sec

The following tests FAILED:
	  2 - check_atom (Failed)
	  4 - check_sexpr_parser (Failed)
	  8 - test_examples.py (SEGFAULT)
	 10 - test_minecraft.py (SEGFAULT)
	 14 - test_pln_tv.py (SEGFAULT)
	 15 - test_run_metta.py (SEGFAULT)
Errors while running CTest
make[3]: *** [CMakeFiles/check.dir/build.make:70: CMakeFiles/check] Error 8
make[2]: *** [CMakeFiles/Makefile2:158: CMakeFiles/check.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:165: CMakeFiles/check.dir/rule] Error 2
make: *** [Makefile:179: check] Error 2
```

Git status:
```
$ git status
On branch test_task
Your branch is up to date with 'h555yand/test_task'.

nothing to commit, working tree clean
```

h555yand:
I've repeated this problem on test_examples.py.
Crash happens on: fritz_frog = metta.run('!(if (and ($x croaks) ($x eats_flies)) (= ($x frog) True) nop)')[0]
Then in runner.py: results = hp.metta_run(self.cmetta, parser.cparser)

Do you know, is everything fine on any branch without py_match_ function?

vsbogd:
CI is always green on main branch - it is a merge policy. I have checked tests on your branch before your changes (commit 91a8858a041a6e56620838afd37a944fd61f79a6) it is green.

h555yand:
Hi All,
I've made several changes and current test log is:
The following tests FAILED:
  8 - test_examples.py (Failed)
 10 - test_minecraft.py (SEGFAULT)
 14 - test_pln_tv.py (Failed)
 15 - test_run_metta.py (Failed)

SEGFAULT is not because of crashes in match_ functions.
The reason is endless loop in 'runner.py' in 'run' function with 'results = hp.metta_run(self.cmetta, parser.cparser)'
It calls match_ functions and probably expects some results to complete action that differs from the returning results. After many iterations (several minutes running) it crashes. I expect due to memory overflow.

Debugging other errors I think they are linked with expected returning results as well:
Second list contains 1 additional elements.
First differing element 1:
First differing element 0:
First list contains 1 additional elements.

It's difficult to fix such errors without business logic knowledge of using the tests and existing functions.
Current default match_ logic (def match_(self, catom, *args):) is similar to default logic from match_by_equality(self, other).

May be it's more efficient if somebody experienced in existing code and logic have a look at these errors.

Best regards,
Aleksei

Necr0x0Der:
As far as I can see, `fn match_(&self, other: &Atom)` executes grounded matching only if it is defined in API of a concrete grounded object (similar to `execute`). And this is right. For example, `execute` is not called for any `ValueAtom`, because it doesn't have `execute`. However, after this, you add `match_` to `GroundedObject`. This is a bad idea. Your `GroundedObject::match_` acts differently from `match_by_equality`, when multiple matches appear (in particular, in non-deterministic scenarios). But you don't need to know "business logic" to fix this. The idea to have `match_` for `GroundedObject` is bad, because it replicates the already available functionality in Rust, and if we decide to change it, we will have to look into two places. And also, doing this in Python via numerous gateways is less efficient. The solution is to remove `match_` from `GroundedObject` completely. You don't need to make its functionality similar to Rust. Your PR should absolutely not affect the current interpretation of the existing code, so you won't have to worry about this at all. What you need is to introduce something like `MatchableAtom` class which can inherit from `ValueAtom` for now and which will have `match_` method (which can delegate matching to a python function in the same fashion as `OperationAtom` does it for `execute`), and make sure that `call_match_on_grounded_atom` is really called only when `match_` is in the object API.
In `hyperonpy.cpp`, there is a check for `if (py::hasattr(pyobj, "execute"))`. You should extend it with checking for `match_` as well.

Necr0x0Der:
PS apparently, if we just remove `match_` from `GroundedObject` and from `PY_VALUE_API`, everything is green.  Custom matching functions should be written by MeTTa developers. And since your `GroundedObject::match_` works somehow, the API for such custom matching is already mostly created. There might be some tricky problems, but they will become clear when we start using it in practice. And what is already implemented seems not bad for starters.

Necr0x0Der:
PPS your PR is kinda broken. Your `hyperonpy.cpp` contains `metta_evaluate_atom`, `metta_clone` and new signature for `metta_new`, which were introduced recently, but Python and Rust code is old and don't have them. It just doesn't compile. (It is easily fixable, but it would be better to have a compilable code for convenience).

vsbogd:
@h555yand, you need to merge `trueagi/main` into your branch to build as @Necr0x0Der said. You can do it manually in your local repo and then push, or by pressing `Update branch` button in GitHub GUI.

vsbogd:
I agree with https://github.com/trueagi-io/hyperon-experimental/pull/219#issuecomment-1363372262.

The first issue in the `hyperonpy.cpp` is that it expects that `match_` is always implemented in Python. But it is not what we are trying to achieve. Our goal is keep default behavior (calling `match_by_equality` in Rust) until Python API user defined some custom `match_`. This is why `match_` is optional in C API: https://github.com/h555yand/hyperon-experimental/blob/de1b8eb5300349d690597c0a818e0231912b6c76/c/src/atom.rs#L38. And it is the reason why `execute` is null by default in `hyperonpy.cpp`: https://github.com/h555yand/hyperon-experimental/blob/de1b8eb5300349d690597c0a818e0231912b6c76/python/hyperonpy.cpp#L65. So `GroundedObject` constructor in `hyperonpy.cpp` need to be fixed to pass `py_match_` into C API iff Python grounded object defines `match_` and pass `nullptr` in other cases.

Second issue which is probably causes segfault is that `std::string` instance created inside the conversion loop at https://github.com/h555yand/hyperon-experimental/blob/de1b8eb5300349d690597c0a818e0231912b6c76/python/hyperonpy.cpp#L128 is freed before `callback` is called. It is a temporary object which is created after conversion of the Python string into C++ string. The code keeps pointer to the internal buffer of the `std::string` but instance itself is freed after loop iteration ends. To eliminate this one can create an array of the `std::string` on the stack and keep strings inside. Then the pointers are valid during call of the `callback` and memory is freed after `py_match_` function ends.

h555yand:
Hi All,
many thanks for the comments.
I've made the corresponding changes.
Now it passes the tests.
Could you please have a look from your side? 
Best regards, Aleksei

c/src/atom.rs
```diff
@@ -303,3 +349,36 @@ impl Drop for CGrounded {
     }
 }
 
+#[cfg(test)]
+mod tests {
+use super::*;
+    #[test]
+    pub fn test_match_callback(){
```
vsbogd, line 356: Test can be made more compact:
```rust
    #[test]
    pub fn test_match_callback(){
        let var = str_as_cstr("var#1");
        let atom = atom_to_ptr(Atom::sym("atom_test"));
        let vec = vec![ binding_t{ var: var.as_ptr(), atom } ];
        let bindings = (&vec).into();
        
        let mut results: Vec<Bindings> = Vec::new();
        let context = addr_of_mut!(results).cast::<c_void>();

        CGrounded::match_callback(bindings, context);

        assert_eq!(results, vec![Bindings::from(vec![
                (VariableAtom::new_id("var", 1), Atom::sym("atom_test"))])]);
        unsafe{ atom_free(atom) };
    }
```
Also I would add C API test for the custom `match_` implementation instead of Rust one. The C API test will check all the flow front-to-back.

vsbogd (commented):
One thing which is still absent is a Python unit test for the custom `match_` implementation which would demonstrate the code works. 
I would also add C API unit test separately to ensure that C API works if internal implementation is changed in future.

python/hyperon/atoms.py
```diff
@@ -164,6 +168,16 @@ def execute(self, *args, res_typ=AtomType.UNDEFINED):
     def __eq__(self, other):
         return isinstance(other, OperationObject) and self.name == other.name
 
+class MatchableObject(ValueObject):
+    def match_(self, catom, *args):
```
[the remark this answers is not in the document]
Necr0x0Der: My idea was to pass a method for matching the atom to the `MatchableObject` constructor similar to `OperationObject`, so we don't have to define new classes with `match_` in order to create new atoms with custom matching. But maybe the situation with operations is different, and custom matching is relevant for complex objects with internal structure, and it doesn't necessary to have a shortcut with a generic class for them. For this PR, I agree that this is definitely ok to have `MatchableObject`, which raises  "MatchableObject::match_() is not implemented" exception (so one may just inherit from this class and see that `match_` should be implemented), and sort this out later.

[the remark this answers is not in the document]
h555yand: Done

python/hyperon/atoms.py
```diff
@@ -100,6 +100,10 @@ def call_execute_on_grounded_atom(gnd, typ, args):
     args = [Atom._from_catom(catom) for catom in args]
     return gnd.execute(*args, res_typ=res_typ)
 
+def call_match_on_grounded_atom(gnd, catom):
+    args=[]
+    return gnd.match_(gnd, catom, *args)
```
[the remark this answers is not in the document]
h555yand: Done

python/hyperonpy.cpp
```diff
@@ -110,6 +117,30 @@ exec_error_t *py_execute(const struct gnd_t* _cgnd, struct vec_atom_t* _args, st
...
+    {
+        py::dict dict_ = py_list[i];
+        for (auto item : dict_){
+            string_array[i] = std::string(py::str(item.first));
+            c_binding_t[i].var = string_array[i].c_str();
+            c_binding_t[i].atom = item.second.attr("catom").cast<CAtom>().ptr;
+        }
+    }
+    data = {c_binding_t, size};
+    callback(data, context);
```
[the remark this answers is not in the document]
h555yand: Done

h555yand:
>

c/src/atom.rs
h555yand: Done for Rust part.

h555yand:
Hi, I have questions on unit tests.
For C part as I understand one should call Rust function "Grounded::fn match_(&self, other: &Atom) ". Then automatically will be called C++ py_match_ , then Python call_match_on_grounded_atom and finally Matchableobject match_. I saw test C files but I could find unit test for Grounded::execute that I can use as a reference. What should one put as a callback function argument in match_ call in C? I see realization of callback for grounding_space_query in C but match_ has a different callback logic. I need some advice here.

For Python unit test, should I create a new test py file and what test can be used as a good reference? As I understand one should create Matchableobject and then call "metta.run" similar to fritz_frog = metta.run('!(if (and ($x croaks) ($x eats_flies)) (= ($x frog) True) nop)')[0]?

vsbogd:
The example of unit test for the custom `match_` implementation in Rust is https://github.com/trueagi-io/hyperon-experimental/blob/4afa4256d4a105bb9c8423552dcb17cb9144837c/lib/src/atom/matcher.rs#L922-L990

Similar scheme should be used to implement an unit test in C or Python. But for testing this in C or Python I would not use `match_atoms` directly but use `GroundingSpace::query` to receive the result. Also test above is a kind of complex I would use much simpler `match_` definition. For example `match_` could return the type of the passed atom (Variable, Symbol, Expression or Grounded) as a symbol atom. 

Basically the plan is:
- write Python object with custom `match_` definition;
- put the atom into an instance of `GroundingSpace` (lets call it `space`);
- call `space.query()` with different kinds of atoms and check that type is returned correctly.

A proper place for such simple test is https://github.com/trueagi-io/hyperon-experimental/blob/main/python/tests/test_atom.py. You can find simple tests for `execute` there as well.

Lets start from Python unit test and skip C unit test for now to not make the task too complex.

python/hyperon/atoms.py
```diff
@@ -164,6 +167,23 @@ def execute(self, *args, res_typ=AtomType.UNDEFINED):
     def __eq__(self, other):
         return isinstance(other, OperationObject) and self.name == other.name
 
+class MatchableObject(ValueObject):
+    def match_(self, atom):
+        raise RuntimeError("MatchableObject::match_() is not implemented")
+        # Below example of returning results
```
vsbogd, line 173: This example is not necessary when Python unit test is implemented.

h555yand:
Hi Vitaly,
I've implemented Python test and during running usually Segmentation fault arises, but sometimes it works.
Log is below (source files with log calls in the my branch):
.............test_match_ start
fn match_ self.api().match_: self: CGrounded(0x555a1c349f58), other: test
py_match_ _gnd: 0x555a1c349f58, _atom: 0x555a1c34bce0
MatchableObject match_ atom:  test
MatchableObject match_ py_bind_arr:  [{'atom_type': SYMBOL}]
py_match_ c_binding_t[i].var: atom_type
fn match_ func results: [{$atom_type: SYMBOL}], ptr: 0x555a1c33b2f0
query result:  [{'atom_type': <hyperonpy.CAtom object at 0x7ff6e268e030>}]
query ret_res:  [{'atom_type': SYMBOL}]
Segmentation fault (core dumped)

I understand that this is a memory problem but don't know what exactly. 
All log chain looks fine except of last one: result = space.query(S('test'))
        print('test_match_ result:', result)

Could you please give some advises from your side?

Regards, Aleksei

python/hyperonpy.cpp
```diff
@@ -110,6 +117,46 @@ exec_error_t *py_execute(const struct gnd_t* _cgnd, struct vec_atom_t* _args, st
...
+
+    py::dict py_dict = py_list[0];
+    binding_array_t data;
+    binding_t c_binding_t[1];
+    std::string string_array[1];
+    for (auto item : py_dict) {
+        string_array[0] = std::string(py::str(item.first));
+        c_binding_t[0].var = string_array[0].c_str();
+        std::cout << "py_match_ c_binding_t[i].var: " <<  c_binding_t[0].var << std::endl;
+        c_binding_t[0].atom = item.second.attr("catom").cast<CAtom>().ptr;
```
vsbogd, line 137: The memory issue is here. Python object contains the pointer to the Rust atom. But this instance of the Rust atom is owned by Python object. The pointer is copied inside `binding_t` structure then inside `match_callback` function Rust program behaves as an owner (here https://github.com/h555yand/hyperon-experimental/blob/b71ea42abf2fbd415f2ecf179d51e759c3293236/c/src/atom.rs#L276). I.e. Rust program keeps pointer to the atom and behaves like an owner. At the same time Python atom keeps pointer to the atom and behaves like owner. When space- is finishing processing a query Python atom is cleaned up and frees Rust atom memory. After that Python test tries to print the results and gets access to the memory through the Rust atom which leads to segmentation fault.

Actually there are two ways to fix it. We could copy atom inside `match_callback`. But I consciously used in C API the approach  when caller passes ownership of the atom into the function. For C API it looks logical because on the side of the caller you can choose whether you need additional copy or not. It is also a bit more compact because you don't need freeing created atom when you passed it into the callback. We can remove this additional copying on the Python side, but we need to modify code of Atoms to explicitly encode two states: Rust object is owned and Rust object is not owned. I don't propose doing this now.

Instead I would propose taking the second approach which is used inside `hyperonpy.cpp`: cloning atom before passing it to the C API function. This code fixes the issue:
```suggestion
        c_binding_t[0].atom = atom_clone(item.second.attr("catom").cast<CAtom>().ptr);
```

h555yand: Hi All, thanks for the advice, It fixed segfault.

I've implemented Python unit test.
It works fine for Symbol, Expr, Grounded types but does not work as expected for Variable atom.
For Variable type custom match_ function is not called and I don't is it a bug or a feature.
Could you please advise?

vsbogd: Ok, it is normal for a `VariableAtom` because when it is matched with anything else it just gets the other atom as a value. You can check the equality to the added atom instead.

I would not add a code to a `MatchableAtom` itself for the purpose of unit test. I would write a separate class instead and wrap it into the test atom (like it is done in `test_grounded_no_copy` for example.

And last thing: before merging this PR please clean up the code:
- remove prints and replace by logging when it is appropriate
- remove commented code
- remove all shortcuts like this one https://github.com/h555yand/hyperon-experimental/blob/00446b1eb5c1e7d83da87a60dbfbdd73f015af16/python/tests/test_atom.py#L4

h555yand: Hi All,
thanks for the comments.
I've made corresponding changes. Hope it works now.
Regards, Aleksei.

python/hyperonpy.cpp
```diff
@@ -110,6 +117,31 @@ exec_error_t *py_execute(const struct gnd_t* _cgnd, struct vec_atom_t* _args, st
     }
 }
 
+void py_match_(const struct gnd_t *_gnd, const struct atom_t *_atom, lambda_t_binding_array_t callback, void *context) {
+    py::object hyperon = py::module_::import("hyperon");
+    py::function call_match_on_grounded_atom = hyperon.attr("call_match_on_grounded_atom");
+    py::object pyobj = static_cast<GroundedObject const *>(_gnd)->pyobj;
+    CAtom catom = (atom_t *)_atom;
```
vsbogd, line 124: We cannot wrap the atom owned by Rust into the Python object as it will be freed second time by Python object:
```suggestion
    CAtom catom = atom_clone(_atom);
```

python/hyperon/atoms.py
```diff
@@ -114,7 +117,7 @@ def __repr__(self):
 
     def copy(self):
         return self
-
+
```
vsbogd, line 120:
```suggestion

```

vsbogd:
@h555yand please review the suggested changes and commit them or suggest another way. 
After that I am ready to merge it.

h555yand:
Hi Vitaly, thanks for the review.
I've accepted the suggested changes, sorry for some mistypes in the last commit.
Happy New Year!
Aleksei

vsbogd (approved):

PR #225, opened 2022-12-19
first step of introduction of stdlib code

Necr0x0Der:


lib/src/metta/runner/stdlib.rs
```diff
@@ -873,12 +885,17 @@ pub fn register_tokens(metta: &Metta, cwd: PathBuf) {
...
 
-    metta.tokenizer.borrow_mut().move_front(&mut stdlib_tokens);
+    metta.tokenizer.borrow_mut().move_front(&mut rust_tokens);
+}
 
-    // &self should be updated
-    // TODO: adding &self might be done not by stdlib, but by MeTTa itself
-    let space_val = Atom::gnd(metta.space.clone());
-    metta.tokenizer.borrow_mut().register_token(regex(r"&self"), move |_| { space_val.clone() });
+pub fn metta_code() -> &'static str {
```
vsbogd, line 891: I would suggest to rename it to `stdlib_metta`.

Necr0x0Der: Since it is defined in `stdlib`, and represented as `stdlib::...` outside, `stdlib::stdlib_metta` looks a little bit tautological, while `stdlib::metta` would look unclear since `metta` is a structure in the code, so I chose `stdlib::metta_code`, but I don't have strong preferences (since it will be moved to `stdlib.metta` later, and I got the idea that `stdlib_metta` resembles `stdlib.metta`).

vsbogd: Yes, ok, I didn't pay attention on module name. I agree it is ok from this perspective, thanks!

PR #234, opened 2023-01-12
Use modern way to find Python interpreter

vsbogd:
Use FindPython3 when it is possible. Upgrade Conan version to 1.57. Upgrade pybind11 version to 2.10.1 and support Python 3.11.


Necr0x0Der:
Seems to work for me. Thanks!

Necr0x0Der (approved):

Issue #244, opened 2023-03-02
Replace `bindings_t` implementation in C and Python API


vsbogd:
MeTTa code heavily uses matching of atoms. Let's say we have two atoms: `(A $x $x)` and `(A B $y)`. Result of the matching is a set of variable bindings or variable values (for both sides) which makes the given expressions equal. For the example above it is `{ $x = $y = B }`.

There is a special data structure to represent variable bindings: `hyperon::atom::matcher::Bindings`. Matching is a fundamental part of the atoms processing so `Bindings` structure is a part of C API. For example it is used in [c/src/space.rs](https://github.com/trueagi-io/hyperon-experimental/blob/main/c/src/space.rs) module to return results of a query to the caller. 

Most of the structures from the Hyperon core library are represented in C API by thin wrappers. For example `atom_t` is a wrapper of the `Atom`. All public functions of `Atom` are represented by `atom_...` functions in [c/src/atom.rs](https://github.com/trueagi-io/hyperon-experimental/blob/main/c/src/atom.rs). It is not a case for `Bindings`. It is represented by totally another kind of structure `bindings_t`. Which is a vector of pairs variable name/atom. This was done to simplify C API and provide more natural C interface. But at the moment `Bindings` API is going to be changed. It is the reason why it is better to replace `bindings_t` by simple wrapper of `Bindings` as it is done for most of other structures.

The task is to replace current `bindings_t` implementation by thin wrapper of the `hyperon::atom::matcher::Bindings`. It requires replacing the `bindings_t` implementation in the Hyperon library C API and adding `bindings_...` functions for each public `Bindings` function. It is also required to add C++ proxy calls into [python/hyperonpy.cpp](https://github.com/trueagi-io/hyperon-experimental/blob/main/python/hyperonpy.cpp) file. And using this calls to represent `Bindings` by class in Python API. Finally all C and Python API unit tests need to be fixed to support new API.

vsbogd:
Leave it here as a first contribution task for @e-gorodnitskiy.

vsbogd:
`check_grounding_space.c` is a test which is implemented by custom printing of variables and values into the character buffer. But another way (and may be it is even more proper) is to compare the list of `Bindings` received with the expected one. I expect this test will be re-implemented as a part of the change by @e-gorodnitskiy.

[...]

PR #245, opened 2023-03-03
Replace Bindings implementation by one which is able to keep variable equalities

vsbogd:
New implementation of the `match_atoms()` treats both atoms symmetrically. It returns variables from both left and right arguments. If caller doesn't need this then `Bindings::narrow_vars()` can be called to leave only variables which are required. Such implementation of the `match_atoms()` is going to be reused in type checker logic.

To return variables new implementation of `Bindings` is done. This implementation keeps both variable equalities and variable values. Variable values are kept without full resolution of sub-variables. When instance of `Bindings` is returned via C API values for all variables are resolved. Separate change in C API is required to keep `Bindings` as is (see #244). 

Also separate change is required to implement `Bindings` editing logic correctly. Right now code expects that adding a binding for the value doesn't multiply number of `Bindings` instances. But there are cases when it is not true. For example we are trying to match two expressions `($x $x)` and ` ((A $y) {space})`, where `space` is a atomspace containing following atoms: `(A B)`, `(A C)`, `(D E)`. Then algorithm first matches `$x` with `(A $y)` and adds this match as bindings for variable `$x`: `{ $x = (A $y) }`. But second match of the `$x` and `{space}` will require checking whether `{space}` and `(A $y)` can be matched. And indeed they can be matched but there are more then one result of such match: `(A B)` and `(A C)`. Thus the final match result should contain two instances of `Bindings`: `{ $x = (A $y), $y = B }` and `{ $x = (A $y), $y = C }`. Current API of `Bindings` structure doesn't allow representing it. It is explained in `TODO` comments before corresponding functions. Taking into account such case is relatively rare I postponed it for the future PRs.


vsbogd:
@e-gorodnitskiy FIY, it is a PR which contains new `Bindings` API.

Necr0x0Der:
@vsbogd , I skimmed through the PR. I supposed there should be something like this to track variables in consequent bindings, but understanding the implementation in detail would be time-consuming and would unlikely lead to any useful input, so I wonder what you expect from the review. Do you think it is necessary to check if this PR solves the issues like https://github.com/trueagi-io/hyperon-experimental/blob/b94274904cea7c96a619381a14283d68835ee594/python/tests/scripts/b2_backchain.metta#L50 , or it is not intended to solve all the problems and doesn't require such checks before merging?

vsbogd:
@Necr0x0Der, first goal is to pay attention to new `Bindings` representation. In particular I suggest closing #127 because new representation allows representing correctly result of matching `(x x)` and `(y z)` as `{ x = y = z }`.

Second goal is to agree on change in unit tests behavior because new logic changes some of them.

In fact I already have second PR with changes to type checker which fixes incorrect type checker behavior which was found by @ngeiswei, but I wanted to merge this PR first to unblock #245.

vsbogd:
Reviewing `Bindings` implementation is a super goal :-) But I agree it is time consuming and not sure how much value it has. But I think we need to agree on API which `Bindings` has and on relations it represents. To me it is important because before this all implementations we used was more or less kind of map `Variable -> Value` and it is not true for this one.

vsbogd:
BTW your example in `b2_backchain.metta`  still doesn't work as far as I can see.

Necr0x0Der (approved):

PR #250, opened 2023-03-13
Replace bindings_t implementation in C and Python API #244

### LINKED ISSUE #244: Replace `bindings_t` implementation in C and Python API

e-gorodnitskiy:
Fixes #244

### DIFF
--- a/c/src/atom.rs
+++ b/c/src/atom.rs
@@ -40,13 +40,14 @@ pub struct bindings_t {
 }
 
 pub type bindings_callback_t = lambda_t<*const bindings_t>;
+pub type bindings_mut_callback_t = lambda_t<*mut bindings_t>;
 
 #[repr(C)]
 pub struct gnd_api_t {
     // TODO: replace args by C array and ret by callback
     // One can assign NULL to this field, it means the atom is not executable
     execute: Option<extern "C" fn(*const gnd_t, *mut vec_atom_t, *mut vec_atom_t) -> *mut exec_error_t>,
-    match_: Option<extern "C" fn(*const gnd_t, *const atom_t, bindings_callback_t, *mut c_void)>,
+    match_: Option<extern "C" fn(*const gnd_t, *const atom_t, bindings_mut_callback_t, *mut c_void)>,
     eq: extern "C" fn(*const gnd_t, *const gnd_t) -> bool,
     clone: extern "C" fn(*const gnd_t) -> *mut gnd_t,
     display: extern "C" fn(*const gnd_t, *mut c_char, usize) -> usize,
@@ -106,7 +107,7 @@ pub extern "C" fn bindings_eq(bindingsa: *const bindings_t, bindingsb: *const bi
 }
 
 #[no_mangle]
-pub extern "C" fn bindings_traverse(cbindings: * const bindings_t, callback: lambda_t<* const var_atom_t>, context: *mut c_void) {
+pub extern "C" fn bindings_traverse(cbindings: *const bindings_t, callback: lambda_t<* const var_atom_t>, context: *mut c_void) {
     let bindings = unsafe{&(*cbindings).bindings};
     bindings.iter().for_each(|(var, atom)|  {
             let name = string_as_cstr(var.name());
@@ -373,13 +374,11 @@ impl CGrounded {
         (self.api().free)(self.get_mut_ptr());
     }
 
-    extern "C" fn match_callback(cbindings: *const bindings_t, context: *mut c_void) {
-        // todo: check for optimality
-        let bindings = unsafe{ (*cbindings).bindings.clone() };
-        mem::forget(cbindings);
-
+    extern "C" fn match_callback(cbindings: *mut bindings_t, context: *mut c_void) {
+        let bindings = ptr_into_bindings(cbindings);
         let vec_bnd = unsafe{ &mut *context.cast::<Vec<Bindings>>() };
         vec_bnd.push(bindings);
+        mem::forget(cbindings);
     }
 
 }
@@ -458,13 +457,12 @@ use std::ptr;
     #[test]
     pub fn test_match_callback() {
 
-        let bindings= bind! {var: expr!("atom_test")};
-        let cbindings = bindings_t{ bindings };
+        let cbindings = bindings_into_ptr(bind!{var: expr!("atom_test")} );
 
         let mut results: Vec<Bindings> = Vec::new();
         let context = ptr::addr_of_mut!(results).cast::<c_void>();
 
-        CGrounded::match_callback(&cbindings, context);
+        CGrounded::match_callback( cbindings, context);
 
         assert_eq!(results, vec![Bindings::from(vec![
                 (VariableAtom::new("var"), Atom::sym("atom_test"))])]);
--- a/c/src/atom.rs
+++ b/c/src/atom.rs
@@ -1,17 +1,14 @@
-use std::env::var;
 use hyperon::*;
 
 use crate::util::*;
 
 use std::os::raw::*;
 use std::fmt::Display;
-use std::ptr::null;
 use std::sync::atomic::{AtomicPtr, Ordering};
-use log::debug;
 
 use hyperon::matcher::Bindings;
 use std::mem;
-//use std::slice;
+
 
 // Atom
 
@@ -40,14 +37,6 @@ pub struct var_atom_t {
 pub struct bindings_t {
     pub bindings: Bindings,
 }
-
-
-fn fff(bin: bindings_t) -> bool {
-    //bin.bindings.into_iter().for_each();
-
-    return false;
-}
-
 
 #[repr(C)]
 pub struct gnd_api_t {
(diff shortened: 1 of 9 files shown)


c/src/atom.rs
```diff
@@ -69,6 +72,101 @@ pub extern "C" fn exec_error_free(error: *mut exec_error_t) {
...
+            callback(&var_atom, context);
+        }
+    )
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn bindings_add_var_binding(bindings: * mut bindings_t, atom: *const var_atom_t) -> bool {
+    let bindings = &mut(*bindings).bindings;
+    let var = VariableAtom::new(cstr_into_string ((*atom).var));
+    let atom = (*(*atom).atom).atom.clone();
```
vsbogd, line 121: Here we should be able to remove `clone` and use `ptr_into_atom` instead:
```rs
    let atom =ptr_into_atom(atom);
```
Atom should be already cloned before calling `bindings_add_var_binding` in `hyperonpy.cpp`.

python/hyperonpy.cpp
```diff
@@ -275,6 +280,29 @@ PYBIND11_MODULE(hyperonpy, m) {
...
 
+    py::class_<CBindings>(m, "CBindings");
+    m.def("bindings_new", []() { return CBindings(bindings_new()); }, "New bindings");
+    m.def("bindings_free", [](CBindings bindings) { bindings_free(bindings.ptr);}, "Free bindings" );
+    m.def("bindings_clone", [](CBindings bindings) { return CBindings(bindings_clone(bindings.ptr)); }, "Deep copy if bindings");
+    m.def("bindings_merge", [](CBindings left, CBindings right) { return CBindings(bindings_merge(left.ptr, right.ptr));}, "Merges bindings");
+    m.def("bindings_eq", [](CBindings left, CBindings right){ return bindings_eq(left.ptr, right.ptr);}, "Compares bindings"  );
+    m.def("bindings_add_var_bindings",
+          [](CBindings bindings, char const* varName, CAtom atom) {
+              var_atom_t var_atom{.var = varName, .atom = atom.ptr};
```
vsbogd, line 291: Here we should clone atom, because it is held by Python interpreter. If we pass pointer to the Rust and Rust move atom, Python interpreter will access to the memory which is already freed.
```suggestion
              var_atom_t var_atom{.var = varName, .atom = atom_clone(atom.ptr)};
```

python/hyperonpy.cpp
```diff
@@ -275,6 +280,29 @@ PYBIND11_MODULE(hyperonpy, m) {
...
+    m.def("bindings_is_empty", [](CBindings bindings){ return bindings_is_empty(bindings.ptr);}, "Returns true if bindings is empty");
+    m.def("bindings_resolve", [](CBindings bindings, char const* varName){ return CAtom(bindings_resolve(bindings.ptr, varName));}, "Resolve" );
+    m.def("bindings_resolve_and_remove", [](CBindings bindings, char const* varName){ return CAtom(bindings_resolve_and_remove(bindings.ptr, varName));}, "Resolve" );
+    m.def("bindings_to_str", [](CBindings bindings) {
+        std::string str;
+        bindings_to_str(bindings.ptr, copy_to_string, &str);
+        return str;
+    }, "Convert bindings to human readable string");
+    // todo: how to pass callback?
+    //pub unsafe extern "C" fn bindings_traverse(bindings: * const bindings_t, callback: lambda_t<* const var_atom_t>, context: *mut c_void) {
```
vsbogd, line 304: Not sure what is the question, because below in line 321 you already have an example of the  `bindings_traverse` call.

python/tests/test_bindings.py
```diff
@@ -0,0 +1,94 @@
...
+from copy import deepcopy
+
+
+class BindingsTest(unittest.TestCase):
+
+    def setUp(self):
+        self.emptyBindings = Bindings()
+
+        self.bindings = Bindings()
+        self.bindings.add_var_bindings("a", SymbolAtom(hp.atom_sym("b")))
```
vsbogd, line 13: 1. There is a shortcut for `SymbolAtom(hp.atom_sym("b"))`: `S("b")`, it is better use it for better readability.
2. `hyperonpy` or `hp` is not intended to be used from the user code, usually there is a Python wrapper to call corresponding function. Low level functions may have some non-obvious invariants.

python/tests/test_bindings.py
```diff
@@ -0,0 +1,94 @@
...
+
+        merged_self = Bindings.merge(self.bindings, self.bindings)
+        self.assertEqual(merged_self, self.bindings)
+
+    def test_bindings_is_empty(self):
+        self.assertTrue(self.emptyBindings.is_empty())
+        self.assertFalse(self.bindings.is_empty())
+
+    def test_bindings_resolve(self):
+        # todo: how to perfom negative tests?
```
vsbogd, line 71: Minor but it is better using `TODO` in upper case, because it is more easy to search it in a codebase.

vsbogd: Regarding negative tests. From the Python API perspective `Bindings.resolve` should return `None` when key is absent. To not repeat the code may be it would be simpler to handle it in `Bindings.resolve` and check for the null pointer. If there are more specific questions in your mind we can discuss it.

vsbogd (changes requested):

c/src/space.rs
```diff
@@ -55,20 +54,11 @@ pub unsafe extern "C" fn grounding_space_get(space: *const grounding_space_t, id
...
-                // put C string into collection which is external to closure
-                // to prevent its deallocation before callback is called
-                vars.push(string_as_cstr(k.name()));
-                var_atom_t{
-                    var: vars.last().unwrap().as_ptr(),
-                    atom: atom_into_ptr(v),
-                }
-            }).collect();
-        callback((&vec).into(), context);
+        let b = bindings_into_ptr(result.clone());
```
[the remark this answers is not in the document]
e-gorodnitskiy: done

python/hyperonpy.cpp
```diff
@@ -54,7 +60,7 @@ py::object get_attr_or_fail(py::handle const& pyobj, char const* attr) {
 
 extern "C" {
     exec_error_t *py_execute(const struct gnd_t* _gnd, struct vec_atom_t* args, struct vec_atom_t* ret);
-    void py_match_(const struct gnd_t *_gnd, const struct atom_t *_atom, lambda_t_bindings_t callback, void *context);
+    void py_match_(const struct gnd_t *_gnd, const struct atom_t *_atom, lambda_t______bindings_t callback, void *context);
```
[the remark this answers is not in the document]
e-gorodnitskiy: done, but let us discuss const/mut

python/hyperonpy.cpp
e-gorodnitskiy: done

python/hyperonpy.cpp
e-gorodnitskiy: I am sorry for unprecise wording. It is about how to pass callback from python and express it here. Below binding_traverse is called from the c/cpp code. The question is how to call it from the python and express with pybind.

c/src/atom.rs
```diff
@@ -69,6 +74,104 @@ pub extern "C" fn exec_error_free(error: *mut exec_error_t) {
...
+            callback(&var_atom, context);
+        }
+    )
+}
+
+#[no_mangle]
+pub extern "C" fn bindings_add_var_binding(bindings: * mut bindings_t, atom: *const var_atom_t) -> bool {
+    let bindings = unsafe{ &mut(*bindings).bindings };
+    let var = VariableAtom::new(cstr_into_string (unsafe{(*atom).var}));
+    let atom = ptr_into_atom(unsafe{(*atom).atom});
```
vsbogd, line 126: In fact this should not work, the proper `ptr_into_atom` usage is:
```suggestion
    let atom = ptr_into_atom(atom);
```

vsbogd: I am sorry, didn't get that `atom` is in fact a pair  `(name, atom)`. Could you please rename a parameter to make it clear.

e-gorodnitskiy: done

python/tests/test_bindings.py
e-gorodnitskiy: done

python/hyperon/atoms.py
```diff
@@ -187,4 +189,54 @@ def ValueAtom(value, type_name=None, atom_id=None):
...
+        return Bindings(hp.bindings_merge(left.cbindings, right.cbindings))
+
+    def add_var_bindings(self, var_name:str, atom: Atom) -> bool:
+        return hp.bindings_add_var_bindings(self.cbindings, var_name, atom.catom)
+
+    def is_empty(self) -> bool:
+        return hp.bindings_is_empty(self.cbindings)
+
+    def resolve(self, var_name: str) -> Atom:
+        return Atom(hp.bindings_resolve(self.cbindings, var_name))
```
vsbogd, line 239: One more thing, the `Atom._from_catom` factory method should be used to correctly create new atom from C structure here and below.

vsbogd (approved):
Looks good, thanks!

ngeiswei:
I'm getting the following error from `metta.py` after this PR
```
Traceback (most recent call last):
  File "/home/nilg/Work/TrueAGI/hyperon-experimental/python/tests/metta.py", line 4, in <module>
    from hyperon import MeTTa
  File "/home/nilg/Work/TrueAGI/hyperon-experimental/python/hyperon/__init__.py", line 1, in <module>
    from .atoms import *
  File "/home/nilg/Work/TrueAGI/hyperon-experimental/python/hyperon/atoms.py", line 195, in <module>
    class Bindings:
  File "/home/nilg/Work/TrueAGI/hyperon-experimental/python/hyperon/atoms.py", line 197, in Bindings
    def __init__(self, bindings: hp.CBindings | None = None):
TypeError: unsupported operand type(s) for |: 'pybind11_type' and 'NoneType'
```
Is it normal?

My version of Python 3.9.2.

vsbogd:
> Is it normal?
> My version of Python 3.9.2.

Thanks for reporting Nil, I will check it asap

vsbogd:
The root cause is new union notation `hp.CBindings | None` was introduced from Python 3.10, and GitHub CI uses ubuntu-latest which is also on Python 3.10. I think I will fix the code to support 3.9 or 3.8.

vsbogd:
Fixed in #254

PR #251, opened 2023-03-14
Fix index memory overflow

vsbogd:
Fixes:
- memory issue for an atoms with large number of subexpressions;
- cleanup empty tree nodes after value is removed.

Documentation to the public API and overall index logic is added.

The idea is to have almost standard prefix tree with additional information about shortcuts to the ends of expressions. These shortcuts then can be used to continue search after matching the whole expression by wildcard key. `MultiTrieNode::add` function adds the key into the prefix tree, and setups shortcuts. `MultiTrieNode::remove` is rewritten via recursion to remove the child node when it becomes empty. `MultTrieNode::get` is only operation which requires an iterator through unexplored paths  so iterator is rewritten without macro.

### DIFF
--- a/lib/src/common/multitrie.rs
+++ b/lib/src/common/multitrie.rs
@@ -25,7 +25,7 @@
 //! `[ Exact(A), LeftPar, Exact(B), RightPar ]` keys.
 //!
 //! Wildcard can be used not only for getting value but also as a key for
-//! keeping value. Thus a singly key in the [MultiTrie] can match many different keys
+//! keeping value. Thus a single key in the [MultiTrie] can match many different keys
 //! for retrieve value. In the example above we could put two
 //! values with `[ Exact(A), Exact(B) ]` and `[ Exact(A), LeftPar, Exact(B), RightPar ]`
 //! keys into the trie and then get both of them using `[ Exact(A), * ]` key.
--- a/lib/src/common/multitrie.rs
+++ b/lib/src/common/multitrie.rs
@@ -1,6 +1,7 @@
 use std::fmt::{Debug, Display};
 use std::collections::{HashMap, HashSet, VecDeque};
 use std::hash::Hash;
+use std::marker::PhantomData;
 use crate::common::shared::Shared;
 
 #[derive(PartialEq, Eq, Clone, Debug, Hash)]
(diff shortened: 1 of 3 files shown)


lib/src/common/multitrie.rs
```diff
@@ -1,244 +1,487 @@
...
+//!
+//! The [TrieToken::Wildcard] is a last kind of token which matches the exact value, the whole
+//! sub-expression or another wildcard. It can recognize anything but when dealing
+//! with subexpression it matches with the whole subexpression only. It cannot
+//! be matched with a part of the subexpression or with the left or right parenthesis.
+//! For example `[ Exact(A), * ]` key matches both `[ Exact(A), Exact(B) ]` and
+//! `[ Exact(A), LeftPar, Exact(B), RightPar ]` keys.
+//!
+//! Wildcard can be used not only for getting value but also as a key for
+//! keeping value. Thus a singly key in the [MultiTrie] can match many different keys
```
ngeiswei, line 28: singly -> single

ngeiswei (approved):
I've only reviewed the top comment, the rest is too involved for me, at that point.  There's a minor typo.

vsbogd:
> I've only reviewed the top comment, the rest is too involved for me, at that point.

I have raised this Draft PR because I have no further ideas at the moment. I tried to replace `Shared<MultiTrieNode>` by `*mut MultiTrieNode` it eliminates the lifetime complexities but (1) surprisingly it works a bit slower (2) it segfaults on freeing. I don't really like the implementation but it works and fixes the issue, so I think it makes sense to merge it.

I think I will add the description of the internal logic before merge.

lib/src/common/multitrie.rs
vsbogd, line 28:
```suggestion
//! keeping value. Thus a single key in the [MultiTrie] can match many different keys
```

ngeiswei:
At least it fixes the PLN formula overflow.  Before it would take a couple minutes and eat up all my 64GB of RAM, while now it is instantaneous!

vsbogd:
I have added documentation for most interesting internal functions and structures.
I have rewritten `MultiTrieNode::insert` function using recursion to make it as clear as possible.

vsbogd:
@ngeiswei I am not sure you have time to review this, but I have tried to make code as clean as possible. I believe the current state should be definitely more clear than before.

lib/src/common/multitrie.rs
vsbogd: Thanks!

ngeiswei:
I won't have time to review it today, but I can give it a try early next week.

vsbogd:
Well, as far as I know your are the primary person who is affected by this, so I can just wait until you look at it.

vsbogd:
@ngeiswei , do you mind if I merge it?

ngeiswei:
Sure, @vsbogd, feel free to merge, I did not have time to look into it this week, and there's always something more important coming up it seems.

PR #255, opened 2023-03-17
Ensure GitHub actions are executed with minimal supported Python version 3.7

vsbogd:
Python version 3.6 is not supported by the actual GitHub actions runners. The oldest runner is `ubuntu-20.04` and the oldest Python version is 3.7.

Makefiles still support 3.6 or later but this support is done on the best effort basis. Guarantees are given for the Python 3.7 and later.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,12 @@ on:
 
 jobs:
   tests:
-    runs-on: ubuntu-latest
+
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        python-version: ["3.7"]
+
     steps:
       - name: Check out repository code
         uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


vsbogd:
I am not sure does it makes sense to support 3.6? May be we should switch to the 3.8 or later.
From the common sense I feel it is better to support for 3.6 in order to support Ubuntu 18.04 users.

vsbogd:
Tried to run Python 3.6 using `pyenv` but it doesn't work for me.

Necr0x0Der:
> Tried to run Python 3.6 using `pyenv` but it doesn't work for me.

Maybe we can "officially" abandon support of Python 3.6 then

vsbogd:
Yes, it what I am proposing. Support 3.6 on a best effort basis.
As an alternative somebody can help with using Python 3.6 inside GitHub actions script.

ngeiswei (approved):
I don't know well enough github ci to help unfortunately.

noskill:
it's reasonable to switch to python 3.7, it's already around 5 years old

noskill (approved):

Issue #188, opened 2022-10-24
Failed to catch multiple type definition


tanksha:
Working with the Dependent types, I unintentionally assigned two types for a given term and MeTTa didn't catch it as a type error, instead I got multiple results. Consider the following code, `p` has assigned of two different types,
```
!(pragma! type-check auto)
(: = (-> $t $t Type))
(: implies (-> $p $r Type))
(: testDTL (-> $p $r (implies $p $r)))
(: PR_prf (-> (implies P R)))
(: p P)
(: p X)
(: r R)
(: q Q)
(= (PR_prf) (testDTL p r))
!(PR_prf)
```
The code executes fine and I got two results `[(testDTL p r), (testDTL p r)]`. 

But, if we exchange the order of the line for (: p P) and (: p X), 
```
....
(: p X)
(: p P)
....
```
then we got a type error at `(= (PR_prf) (testDTL p r))` and that's expected because the left and right side of the equality don't have the same type. That means, even if `p` has assigned the correct type during the second type definition, it has not been updated (I think it shouldn't be) but why did I get two results in the first example? I think it is a bug.

vsbogd:
After running this code on new version of the type checker I see that both versions (with direct and exchanged order) sometimes work and sometimes not.

luketpeterson:
Tracing backwards through the code, in `types.rs`, the function `query_has_type` returns `[{ $X#223 = X }, { $X#223 = P }]` in the error case, and it returns `[{ $X#223 = P }, { $X#223 = X }]` in the non-error case.  Note the reversal in ordering.

The intermittency is caused by the fact that `query_has_type` ultimately calls into `GroundingSpace::single_query`, which, in turn causes the iteration of `GroundingSpace.index`, which is a `MultiTrie`, backed by a `HashMap`.

`HashMap` iteration order is subject to random variation on account of `RandomState` being Rust’s default `Hasher`.  See https://doc.rust-lang.org/std/collections/hash_map/struct.RandomState.html  There are many good explanations as to why this is the case but it's not relevant here.

The more interesting question is what is the right fix?

I don’t have enough understanding of MeTTa yet to make the decision by myself.  Specifically I don’t know if / when it’s allowed to have an atom with multiple assigned types.  The comment by @tanksha makes me think perhaps it should be illegal to ascribe multiple types to the same atom, however the functions return `Vec<>` for types, and I see explicit logic for super-types, etc.

[...]

PR #256, opened 2023-03-18
Unit tests to demonstrate the issue with Bindings API

vsbogd:
Add three unit tests in the ignored state to demonstrate the issue with `Bindings` API.
Change logs for the `check_types` function to demonstrate #188


vsbogd:
@luketpeterson FYI

PR #259, opened 2023-03-20
Changes to handle multiple matching results.  Issue 257

luketpeterson:
There are several issues that need to be discussed.

1. This issue required changing many function signatures in matcher.rs to return `Vec<Bindings>` instead of  `bool` or `Option<Bindings>`.  Many of these functions are `Bindings` methods, so they formerly took `&mut self` as input.

Therefore some situations can cause a single `Bindings` input to create multiple `Bindings` outputs.  In this case, the `&mut self` pattern doesn't make sense.

In order to keeping the current code structure as much as possible, I changed the function signatures to take ownership of the input Bindings.  However this inherently leads to a strange API usage semantic.  See all the ugly `pop()` calls I added to the tests to keep them functioning as before.

For example: 
https://github.com/luketpeterson/hyperon-experimental/blob/e3a2a6c8174dff176e41f55e4157d0c4f5633937/lib/src/atom/matcher.rs#L1073

One solution is to create another abstraction, perhaps `BindingsSet` that is a more ergonomic wrapper around `Vec<Bindings>`, and implement many of the methods on `BindingsSet` instead of on `Bindings`.

It is also possible with a more fundamental refactor it would be possible to reduce the number of places where one-to-many signatures exist.

2. I added the functions `add_var_binding_v2` and `merge_v2` with the new signatures, and kept the original `add_var_binding` and `merge` functions available with the original behavior to ease the transition to the new API.

However, when updating the signatures, the functions `with_var_equality` and `with_var_binding` became identical to the more fundamental functions they each call.  Therefore they can probably be deleted altogether.

3. There was a few places where the logic was tricky and I don't have the experience with MeTTa to be sure I didn't mess up anything subtle.  All the tests continue to pass, but I think a second set of eyes is prudent.

Here: https://github.com/luketpeterson/hyperon-experimental/blob/e3a2a6c8174dff176e41f55e4157d0c4f5633937/lib/src/atom/matcher.rs#L329

And here: https://github.com/luketpeterson/hyperon-experimental/blob/e3a2a6c8174dff176e41f55e4157d0c4f5633937/lib/src/atom/matcher.rs#L733

### DIFF
--- a/lib/src/atom/matcher.rs
+++ b/lib/src/atom/matcher.rs
@@ -150,13 +150,6 @@ impl Bindings {
         }
     }
 
-    // TODO: There is no longer a functional difference between `with_var_equality` and `add_var_equality`
-    //   so one of them is probably unnecessary.
-    // Also, Should this function be made public?  The previous comment said that was the idea.
-    fn with_var_equality(self, a: &VariableAtom, b: &VariableAtom) -> Vec<Bindings> {
-        self.add_var_equality(a, b)
-    }
-
     fn add_var_equality(mut self, a: &VariableAtom, b: &VariableAtom) -> Vec<Bindings> {
         match (self.id_by_var.get(a).copied(), self.id_by_var.get(b).copied()) {
             (Some(a_var_id), Some(b_var_id))  =>
@@ -216,12 +209,6 @@ impl Bindings {
         self.next_var_id = self.next_var_id + 1;
         next_var_id
     }
-
-    // TODO: There is no longer a functional difference between `with_var_binding` and `add_var_binding_v2`
-    //   so one of them is probably unnecessary.  (Commenting out pending deletion)
-    // fn with_var_binding<T1: RefOrMove<VariableAtom>, T2: RefOrMove<Atom>>(mut self, var: T1, value: T2) -> Vec<Bindings> {
-    //     self.add_var_binding(var, value)
-    // }
 
     /// Tries to insert `value` as a binding for the `var`. If `self` already
     /// has binding for the `var` and it is not matchable with the `value` then
@@ -330,7 +317,7 @@ impl Bindings {
                 
                 for (result, mut b_vars_merged) in results {
                     let new_results = if let Some(first_var) = b_vars_merged.get(&var_id) {
-                        result.with_var_equality(first_var, var)
+                        result.add_var_equality(first_var, var)
                     } else {
                         b_vars_merged.insert(*var_id, var.clone());
                         if let Some(value) = b.value_by_id.get(var_id) {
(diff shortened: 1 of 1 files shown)


vsbogd (commented):
Thanks @luketpeterson, overall change looks good, but I agree with your points above.

> 3. There was a few places where the logic was tricky and I don't have the experience with MeTTa to be sure I didn't mess up anything subtle. ...

I have looked at the places you doubt about. `match_result_product()` looks good, but `Bindings::merge_v2()` has an issue. I have added a comment with unit tests and root cause analysis.

> 2. ... the functions with_var_equality and with_var_binding became identical to the more fundamental functions they each call.

Sure, we can delete them, I have commented locally this as well. 

> 1. ... However this inherently leads to a strange API usage semantic. See all the ugly pop() calls I added to the tests to keep them functioning as before.

Yes, this one I have overlooked during the initial analysis. It is an interesting question what to do with it however. One point is we definitely need `Bindings` as a separate type. Looks like adding `BindingsSet` will lead to duplication of `add_var_binding()`, `add_var_equality()` functions. I mean we still need `Bindings::add_var_binding()` but also need `BindingsSet::add_var_binding()` which applies `Bindings::add_var_binding()` to the each item of the `BindingsSet`. Please correct me if I didn't get your idea properly.

If we look at it from the use-cases perspective calling `add_var_binding()` on `Bindings` in almost all cases (except call from `Bindings::merge_v2()`) is used to construct a new instance of  `Bindings`. And in all such cases the caller expects that only result is returned. The same is true for `add_var_equality()` (or `with_var_equality()`). So may be we need somehow separate API for creating new instances of `Bindings` when dividing `Bindings` on few instances is not expected (btw may be it should be expressed via returning `Result` rather then `panic!`). Current versions of methods should be still used for merging `Bindings`. In such case `Bindings::merge_v2()` is only public function which divides `Bindings`.

One way of doing it is introducing two versions of functions (we still has duplication but second version is private):
- one public which always expects `Bindings` is not divided `add_var_binding()` and used externally (btw may be we should rename it into `add_var_value()`);
- one private which can divide `Bindings` and used to implement `Bindings::merge_v2()` (we could call it `merge_with_var_binding()`  or `merge_with_var_value()` for instance).

We could also make a step forward and  move public functions into separate structure. For example we could introduce `BindingsBuilder` which defines `with_var_value` and `with_var_equality()` and `with_var()` functions which return `Self`. Then `BindingsBuilder::build()` function applies these calls. We could also provide possibility to a caller to expect single `Bindings` instance from `build()` or expect `Vec<Bindings>`.

luketpeterson:
I'm prototyping an interface based on the `BindingsBuilder` idea.  In the meantime, does it make sense to merge this PR (which preserves the existing public interface), and then implement the new interface as we transition the rest of the code to use the "v2" versions?

vsbogd (approved):

vsbogd:
Thanks, sure, I also thought about asking you if you feel it can be merged.

PR #262, opened 2023-04-05
Fix (let ...) pattern/template var mapping

vsbogd:
The variables in pattern part of `let` expressions are unique and visible only inside `let` expression. Variables from the matched part vice versa can be visible outside of `let` expression and thus should be moved into the template untouched.
Another part of the issue is the possible variable loop between internal variable in pattern and external variable in the matched atom. Current code doesn't match such patterns.

As a side change logging logic is improved to spot the loops inside bindings. Also `make_variables_unique()` function is rewritten using `Atom` iterator. It allows removing cloning in one usage at least.

This PR fixes the behavior and introduces Rust unit tests for the behavior.

It makes the following original example work correctly:
```
;; Knowledge
(→ P Q)
(→ Q R)

;; Rule
(= (rule (→ $p $q) (→ $q $r)) (→ $p $r))

;; Query (does not work as expected)
(= (query $kb)
   (let* (($pq (→ $p $q))
          ($qr (→ $q $r)))
     (match $kb
       ;; Premises
       (, $pq $qr)
       ;; Conclusion
       (rule $pq $qr))))

;; Call
!(query &self)
;; [(→ P R)]
```

### DIFF
--- a/lib/src/atom/matcher.rs
+++ b/lib/src/atom/matcher.rs
@@ -955,8 +955,8 @@ mod test {
     fn match_variable_with_unique_itself() {
         let x_uniq = Atom::Variable(VariableAtom::new_id("x", 1));
         assert_match(
-            make_variables_unique(&expr!(("A" x) ("B" x))),
-                                   expr!(("A" x)    z   ),
+            make_variables_unique(expr!(("A" x) ("B" x))),
+                                  expr!(("A" x)    z   ),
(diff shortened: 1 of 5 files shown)


vsbogd:
Adding similar fix for the `case` operation.

Necr0x0Der (approved):

ngeiswei (approved):

PR #266, opened 2023-04-07
Use repr() to display Python grounded atoms in logs

vsbogd:
One liner fixes issue with displaying both number 42 and string "42" as 42.


vsbogd:
@bessveta FYI

Necr0x0Der (approved):

PR #271, opened 2023-04-10
Fix superpose to interpret arguments independently

vsbogd:
This PR is to fix the following MeTTa script:
```
(: f (-> Number String))
!(superpose ((get-type (f 42)) (get-type (f "42"))))
```
which returns
```
[(Error ((get-type (f 42)) (get-type (f '42'))) NoValidAlternatives)]
```
This result shows interpreter cannot interpret the whole `superpose` expression because second argument `(get-type (f "42"))` returns empty result.

The PR adds unit tests to fix current behavior. `SuperposeOp` implementation is changed to interpret each sub-expression of expression independently and return combined result. In addition type for the `Error` is added into `stdlib` in order to prevent interpretation of the `Error` expression returned by `superpose`.

### DIFF
--- a/lib/src/metta/runner/stdlib.rs
+++ b/lib/src/metta/runner/stdlib.rs
@@ -612,7 +612,15 @@ impl Grounded for CollapseOp {
 }
 
 #[derive(Clone, PartialEq, Debug)]
-pub struct SuperposeOp { }
+pub struct SuperposeOp {
+    space: Shared<GroundingSpace>,
+}
+
+impl SuperposeOp {
+    fn new(space: Shared<GroundingSpace>) -> Self {
+        Self{ space }
+    }
+}
 
 impl Display for SuperposeOp {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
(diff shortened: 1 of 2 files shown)


lib/src/metta/runner/stdlib.rs
```diff
@@ -622,16 +630,22 @@ impl Display for SuperposeOp {
...
-        let expr = atom_as_expr(args.get(0).ok_or_else(arg_error)?).ok_or(arg_error())?;
+        let atom = args.get(0).ok_or_else(arg_error)?;
+        let expr = atom_as_expr(&atom).ok_or(arg_error())?;
 
-        Ok(expr.children().clone())
+        let mut superposed = Vec::new();
+        for atom in expr.children() {
+            match interpret_no_error(self.space.clone(), atom) {
+                Ok(results) => { superposed.extend(results); },
+                Err(message) => { return Err(format!("Error: {}", message).into()) },
```
Necr0x0Der, line 645: Should `Err` be returned here, or should it be added to `superposed`? What are implications of these two approaches?

vsbogd: In current version this branch will not be executed because `interpret` never returns `Err` in fact. May be it is more logical to replace it with `panic!`. I returned `Err` here as I would expect in future `interpret` may use `Err` to return some errors which are cannot be handled on a MeTTa code level. But may be it doesn't make sense.

vsbogd: I would say it is a separate task to understand whether we need `Err` in `interpret`.

vsbogd: To summarize: I would leave it as is or add a `panic!` here, but anyway we need to revisit this code while revisiting `interpreter` API.

Necr0x0Der: OK, got it. I thought `Err` can appear because of badly typed expressions or other errors in MeTTa code, but it seems that MeTTa `Err` expressions will already be produced instead of Rust `Err`. So, let's leave it as is.

Necr0x0Der (approved):

PR #275, opened 2023-04-14
Low-impact changes to adopt new BindingsSet API

luketpeterson:
Changing some tests & documentation to call newer "_v2" functions instead of compatibility shims
Changing GroundingSpace & Space trait `query` method to return BindingsSet instead of Vec<Bindings>

Still remaining to migrate are: FFI Exports & Interpreter

### DIFF
--- a/lib/benches/grounding_space.rs
+++ b/lib/benches/grounding_space.rs
@@ -5,6 +5,7 @@ extern crate test;
 use test::Bencher;
 
 use hyperon::*;
+use hyperon::matcher::BindingsSet;
 use hyperon::space::grounding::*;
 
 fn space(size: isize) -> GroundingSpace {
@@ -24,7 +25,7 @@ fn query_x10(bencher: &mut Bencher) {
     let space = space(10);
     bencher.iter(|| {
         let res = space.query(&expr!("=" ("func-9" "arg") X));
-        assert_eq!(res, vec![bind!{ X: Atom::sym("arg") }]);
+        assert_eq!(res, BindingsSet::from(bind!{ X: Atom::sym("arg") }));
     })
 }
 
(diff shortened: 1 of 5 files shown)


lib/src/atom/matcher.rs
```diff
@@ -897,7 +914,8 @@ pub fn apply_bindings_to_atom(atom: &Atom, bindings: &Bindings) -> Atom {
 pub fn apply_bindings_to_bindings(from: &Bindings, to: &Bindings) -> Result<Bindings, ()> {
     // TODO: apply_bindings_to_bindings can be replaced by Bindings::merge,
     // when Bindings::merge are modified to return Vec<Bindings>
-    Bindings::merge(to, from).filter(|bindings| !bindings.has_loops()).ok_or(())
+    //LP QUESTION: unclear whether behavior to filter out Bindings with loops continues to make this function relevant
```
vsbogd, line 917: No, filtering loops doesn't make this function unique. In fact I mechanically moved this check from a previous version which implemented custom algorithm and this check was needed. Now we have `Bindings::merge` and all these checks are inside. So we can remove this function safely.

lib/benches/grounding_space.rs
```diff
@@ -24,7 +25,7 @@ fn query_x10(bencher: &mut Bencher) {
     let space = space(10);
     bencher.iter(|| {
         let res = space.query(&expr!("=" ("func-9" "arg") X));
-        assert_eq!(res, vec![bind!{ X: Atom::sym("arg") }]);
+        assert_eq!(res, BindingsSet::from(bind!{ X: Atom::sym("arg") }));
```
vsbogd, line 28: I see this pattern is frequently met in code. May be we need adding another macro `bind_set!` to construct `BindingSet` from `bind!` arguments?

vsbogd (dismissed):
Thanks Luke, looks good to me. Feel free to merge and fix comments or fix comments and re-review.

lib/benches/grounding_space.rs
luketpeterson: Adding bind_set! macro & documentation.  Updated to use where appropriate.

lib/src/atom/matcher.rs
luketpeterson: Ok.  I eliminated the call from GroundingSpace::query, as it appeared redundant.  I did not touch Interpreter.

vsbogd (approved):

vsbogd:
Thanks!

PR #278, opened 2023-04-25
Bindings set migration

luketpeterson:


vsbogd (commented):
Luke, change looks fine to me. I have added couple of suggestions please let me know what do you think.

c/src/atom.rs
```diff
@@ -171,6 +192,80 @@ pub extern "C" fn bindings_resolve_and_remove(bindings: *mut bindings_t, var_nam
...
+    bindings_set_into_ptr(BindingsSet::empty())
+}
+
+#[no_mangle]
+pub extern "C" fn bindings_set_single() -> *mut bindings_set_t {
+    bindings_set_into_ptr(BindingsSet::single())
+}
+
+#[no_mangle]
+pub extern "C" fn bindings_set_from_bindings(bindings: *const bindings_t) -> *mut bindings_set_t {
```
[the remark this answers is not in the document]
luketpeterson: Done.

c/src/atom.rs
```diff
@@ -171,6 +192,80 @@ pub extern "C" fn bindings_resolve_and_remove(bindings: *mut bindings_t, var_nam
...
+    let value = unsafe{ &(*value).atom };
+
+    let mut owned_set = BindingsSet::empty();
+    core::mem::swap(&mut owned_set, set);
+    let mut result_set = owned_set.add_var_binding(TryInto::<&VariableAtom>::try_into(var).unwrap(), value);
+    core::mem::swap(&mut result_set, set);
+}
+
+#[no_mangle]
+pub extern "C" fn bindings_set_merge(a: *const bindings_set_t, b: *const bindings_set_t) -> *mut bindings_set_t {
```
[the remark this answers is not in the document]
luketpeterson: Done.  My fear was that the user might not realize they were handing off ownership, leading to a use-after-free error.  But I think the C API users just need to be very careful.  I documented this for every function in the safer_ffi branch, but it's still the duty of the caller to read and obey the documentation.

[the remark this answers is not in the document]
luketpeterson: Also applied the same change to bindings_merge_v2, to eliminate the clone() there.

c/tests/check_atom.c
```diff
@@ -10,6 +10,50 @@ void setup(void) {
...
+    bindings_add_var_binding(bindings_c, &var_atom_c);
+
+    bindings_set_t* set_2 = bindings_set_from_bindings(bindings_c);
+    bindings_set_t* result_set = bindings_set_merge(set_1, set_2);
+
+    bindings_set_add_var_equality(result_set, atom_var("a"), atom_var("a_prime"));
+
+    bindings_set_add_var_binding(result_set, atom_var("d"), atom_sym("D"));
+
+    bindings_set_iterate(result_set, &print_bindings, NULL);
```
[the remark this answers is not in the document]
luketpeterson: Added additional checks to the test.

c/src/atom.rs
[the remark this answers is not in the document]
vsbogd: Well, yes C API users should always know what they do. The `*mut` pointer is on hint that argument is moved here. We could also use more verbose parameter names example adding `_moved` postfix or `move_` prefix.

c/src/atom.rs
```diff
@@ -171,6 +196,106 @@ pub extern "C" fn bindings_resolve_and_remove(bindings: *mut bindings_t, var_nam
...
+    let b = unsafe{ &(*b).atom };
+
+    let mut owned_set = BindingsSet::empty();
+    core::mem::swap(&mut owned_set, set);
+    let mut result_set = owned_set.add_var_equality(a.try_into().unwrap(), b.try_into().unwrap());
+    core::mem::swap(&mut result_set, set);
+}
+
+#[no_mangle]
+pub extern "C" fn bindings_set_add_var_binding(set: *mut bindings_set_t, var: *const atom_t, value: *const atom_t) {
```
vsbogd, line 275: Playing with API I have found an issue. If one uses `BindingsSet::empty()` to create new `BindingsSet` instance then it is impossible to add something into it. It always ends up with an empty `BindingsSet` instance.

python/hyperonpy.cpp
```diff
@@ -330,6 +339,56 @@ PYBIND11_MODULE(hyperonpy, m) {
...
+    // code base, all the way back to Rust. Then I'd like to unify the behavior between a BindingsSet with a single empty
+    // Bindings and a BindingsSet with no Bindings whatsoever.  I believe this is one step towards a future where
+    // BindingsSets can represent lazily-evaluated bindings, and ultimately infinite and explosive sets
+    m.def("bindings_set_is_empty", [](CBindingsSet set) {
+        if (bindings_set_len(set.ptr) == 0) return true;
+        bindings_set_t* test_set = bindings_set_single();
+        bool val = bindings_set_eq(set.ptr, test_set);
+        bindings_set_free(test_set);
+        return val;
+    }, "Returns true if BindingsSet contains no Bindings");
```
vsbogd, line 357: 1. I would wait for the use-case in code first. The current logic doesn't treat `BindingsSet` like a better version of `Bindings`. It treats it as a set of `Bindings` and always goes through this set to work with each `Bindings` instance. For instance it doesn't make sense to introduce `BindingsSet::resolve()` because all of the variables values in some `Bindings` instance form the inference context and one value is not useful out of this context. `BindingsSet::single` in matcher means that two atoms are equal, and `BindingsSet::empty` means they cannot be matched. Thus I don't see the use-case for such semantics of `_empty()` yet. And if this semantics is not used then name can be confusing because one may think it returns true iff no `Bindings` instances are inside.

2. If you are agree with first comment we have two options here: introduce `_len` function in `hyperonpy.cpp` and map it to the C API function, or leave `_empty` in `hyperonpy.cpp` then probably it makes sense to replace `_len` by `_empty` in C API.

python/hyperon/atoms.py
```diff
@@ -250,3 +253,62 @@ def iterator(self):
...
+            hp.bindings_set_free(self.c_set)
+            self.c_set = None
+
+    def clone(self):
+        return BindingsSet(hp.bindings_set_clone(self.c_set))
+
+    def is_empty(self) -> bool:
+        return hp.bindings_set_is_empty(self.c_set)
+
+    def add_var_binding(self, var_name:str, atom: Atom) -> bool:
```
vsbogd, line 296: I would made it an Atom, as far as you introduced it on the C API level. Same for the `add_var_equality`.

luketpeterson: Done.  I also added a union type argument to Bindings.add_var_bindings() in the Python code, so it's possible to interact with Bindings and BindingsSets in a consistent way from Python.

python/hyperonpy.cpp
luketpeterson: There is a conceptual subtlety here that I don't fully appreciate.  My preference is to eliminate bindings_set_len() from the C API (and also from Rust), but you raise an interesting question about how to express (in code and also in documentation) the difference between a BindingsSet with one empty Bindings, and a BindingsSet with no Bindings.

As you say, one is the absence of constraints, and the other is fully constrained.  I think another word is needed, as opposed to empty.

luketpeterson: I filled this issue to keep the discussion in one place:  https://github.com/trueagi-io/hyperon-experimental/issues/281

c/src/atom.rs
luketpeterson: Good point.  I actually can't think of any good reason to expose the BindingsSet::empty() constructor to the user.  As you point out, in light of the other functionality we've chosen to expose, BindingsSet::empty is actually useless.  Maybe this should be a private method in Rust, and not exposed to C & Python.

luketpeterson: I removed if from the C & Python APIs.  I left it in Rust because it provided a handy way to make an object to provide to mem::swap() to take ownership from a mutable pointer to a BindingsSet in the C API.

I want to find a better way to describe the meaning of the different states of a BindingsSet; and when we agree on something I will harmonize the code.

luketpeterson: Thanks to your comment, I now see the value in keeping the empty constructor, for composing the return value from a grounded atom match.  So I've re-added it.

python/hyperonpy.cpp
luketpeterson: In my opinion this PR can be merged - even without finalized names for empty & single.  It will be a simple matter to rename them.  And I would hate to hold back other work on a conceptual discussion that might take longer to come to a conclusion.

vsbogd (approved):
Thanks Luke, I agree we can discuss `BindingsSet` in a background. Looks good to me, thank you very much for finishing it.

PR #286, opened 2023-05-04
Fix for gracefully returning on parsing errors.

### LINKED ISSUE #273: Gracefully return on parsing errors
Currently parsing an invalid expression like `(Q 2` raises something like
```
thread '<unnamed>' panicked at 'Unexpected end of expression', /home/adamv/PycharmProjects/hyperon-experimental/lib/src/metta/text.rs:129:9
```

cowboyphilosopher:
Fixes #273  
There are two parsing functions that were throwing panic errors. text.rs has parse and parse_exp functions with unhandled panics. This fix returns a Result object for those 2 functions. The Result/Error is then handled when called in the mod.rs files. Also marked a test as ignored since the function no longer panics. test_panic_on_unbalanced_brackets should be deprecated.


lib/src/metta/text.rs
```diff
@@ -77,19 +77,27 @@ impl<'a> SExprParser<'a> {
...
+                    return Ok(Some(Atom::var(token)));
                 },
                 '(' => {
                     self.it.next();
-                    return Some(self.parse_expr(tokenizer))
+                    if let Ok(expr) = self.parse_expr(tokenizer) {
+                        return Ok(Some(expr));
+                    } else {
+                        return Err("Error parsing expression".to_string());
+                    }
```
vsbogd, line 88: Can be simplified to:
```suggestion
                    return self.parse_expr(tokenizer).map(Some);
```

lib/src/metta/text.rs
```diff
@@ -77,19 +77,27 @@ impl<'a> SExprParser<'a> {
...
                 },
-                ')' => panic!("Unexpected right bracket"),
+                ')' => return Err("Unexpected right bracket".to_string()),
                 _ => {
-                    return self.parse_atom(tokenizer);
+                    if let Some(atom) = self.parse_atom(tokenizer) {
+                        return Ok(Some(atom));
+                    } else {
+                        return Err("Error parsing atom".to_string());
+                    }
```
vsbogd, line 96: Can be simplified to:
```suggestion
                    return Ok(self.parse_atom(tokenizer));
```

lib/src/metta/text.rs
```diff
@@ -253,10 +266,11 @@ mod tests {
     }
 
     #[test]
+    #[ignore]
```
vsbogd, line 269: We could just check that a proper error is returned instead of expecting a panic. It also makes sense to add unit test for other possible errors.

lib/src/metta/runner/mod.rs
```diff
@@ -115,49 +115,50 @@ impl Metta {
     pub fn run(&self, parser: &mut SExprParser) -> Result<Vec<Vec<Atom>>, String> {
         let mut mode = Mode::ADD;
         let mut results: Vec<Vec<Atom>> = Vec::new();
-
+    
         loop {
-            let atom = parser.parse(&self.tokenizer.borrow());
-            match atom {
-                Some(atom) => {
+            let atom_result = parser.parse(&self.tokenizer.borrow());
```
vsbogd, line 120: We could use `?` here to re-throw error:
```rs
            let atom_result = parser.parse(&self.tokenizer.borrow())?;
```
Then unpacking result in `match` branches is not needed.

lib/src/metta/text.rs
```diff
@@ -216,8 +229,8 @@ mod tests {
 
         let mut parser = SExprParser::new("ab");
 
-        assert_eq!(Some(expr!("ab")), parser.parse(&tokenizer));
-        assert_eq!(None, parser.parse(&tokenizer));
+        assert_eq!(Some(expr!("ab")), parser.parse(&tokenizer).unwrap());
+        assert_eq!(None, parser.parse(&tokenizer).unwrap());
```
vsbogd, line 233: I would use full result as a first argument of `assert_eq!` here and below then if test fails it will show a difference between actual and expected result instead of panic.
```suggestion
        assert_eq!(Ok(Some(expr!("ab"))), parser.parse(&tokenizer));
        assert_eq!(Ok(None), parser.parse(&tokenizer));
```

vsbogd (commented):
Build fails because `SExprParser::parse` is available through C API. Workaround could be replacing the lines: https://github.com/trueagi-io/hyperon-experimental/blob/244f7e698f513be6738d644e7db7d05d5a361b2a/c/src/metta.rs#L80-L81
by the following code:
```rs
    (*parser).borrow_mut().parse(&(*tokenizer).borrow())
        .unwrap()
        .map_or(std::ptr::null_mut(), |atom| { atom_into_ptr(atom) })
```
but in a long term we need to change `sexpr_parser_parse` function signature to return both `*mut atom_t` and error.

vsbogd:
Thanks @cowboyphilosopher , last comment left, so I am not sure if PR should be merged or should I wait for your fix?

lib/src/metta/text.rs
```diff
@@ -253,10 +258,12 @@ mod tests {
...
-    #[should_panic(expected = "Unexpected right bracket")]
-    fn test_panic_on_unbalanced_brackets() {
-        let mut parser = SExprParser::new("(a))");
-        while let Some(_) = parser.parse(&Tokenizer::new()) {}
+    fn test_unbalanced_brackets() {
+        let mut parser = SExprParser::new("(a)))");
+        let x = parser.parse(&Tokenizer::new());
+        //println!("x: {:?}", x);
+
+        assert_eq!(Err(String::from("Unexpected right bracket")), parser.parse(&Tokenizer::new()));
```
vsbogd, line 266: This test looks ok. I would reuse the instance of the tokenizer, remove the commented code and check the value of the first atom though:
```suggestion
    fn test_unbalanced_brackets() {
        let tokenizer = Tokenizer::new();
        let mut parser = SExprParser::new("(a)))");
        
        assert_eq!(parser.parse(&tokenizer), Ok(Some(expr!(("a")))));
        assert_eq!(parser.parse(&tokenizer), Err(String::from("Unexpected right bracket")));
```

lib/src/metta/text.rs
```diff
@@ -101,13 +101,13 @@ impl<'a> SExprParser<'a> {
         }
     }
 
-    fn parse_atom(&mut self, tokenizer: &Tokenizer) -> Option<Atom> {
-        let token = next_token(&mut self.it);
+    fn parse_atom(&mut self, tokenizer: &Tokenizer) -> Result<Option<Atom>, String> {
```
vsbogd, line 104: I know you just kept the previous API, but it looks like `Option` is not needed here, so the return result could be just `Result<Atom, String>`.

vsbogd (approved):

vsbogd:
Merged, thanks @cowboyphilosopher.

PR #292, opened 2023-05-09
Fast fix for nested let with variables equalities

### LINKED ISSUE #290: Variable equality is incorrectly applied in case of nested `let` expression
The following expression `!(let $f $x (let $f f $x))` being executed should return `[f]` but it returns `[$f#1]` instead.

To calculate the result property we should apply the equality of variables stated on the outer `let` (`{ $x = $f }`) to the nested `let` expression. Thus the nested `let` becomes `(let $x f $x)` or `(let $f f $f)` and finally evaluated to `f`.

The root cause is that `let` is implemented as a grounded operation. Grounded operation doesn't return `Bindings` and thus can only return values of the variables in the atoms. It does it applying `{ $x = $f }` to the `(let $f f $x)` but the result is `(let $x f $f)` because `Bindings` returns `$x` as a value for `$f` and `$f` as a value for `$x`. I omitted variable renaming here which leads to `$f` becomes `$f#1`. Finally `(let $x f $f)` is evaluated to `$f` because $f is absent in the pattern `$x`.

I would suggest fixing it through introducing a proper `let` implementation as a part of minimal interpreter implementation.

vsbogd:
Fixes #290. When two variables are equal we replace one from the pattern by one from the matched atom and not vice versa.

### DIFF
--- a/lib/src/atom/matcher.rs
+++ b/lib/src/atom/matcher.rs
@@ -1386,6 +1386,19 @@ mod test {
     }
 
     #[test]
+    fn bindings_narrow_vars_keeps_vars_equality() -> Result<(), &'static str> {
+        let bindings = Bindings::new()
+            .add_var_equality(&VariableAtom::new("x"), &VariableAtom::new("y"))?
+            .add_var_equality(&VariableAtom::new("x"), &VariableAtom::new("z"))?;
+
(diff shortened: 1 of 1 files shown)


ngeiswei:
Thanks, @vsbogd, please take your time and do the right thing.  I'm using a workaround which is to duplicate assignments in the `let*` statement, so that issue is not blocking my progress for now.

PR #293, opened 2023-05-10
Add Char primary type

### LINKED ISSUE #289: Introduce Char primary type
The following
```
!(== 'A' "A")
!(get-type 'A')
!(get-type "A")
!("A")
```
outputs
```
[False]
[%Undefined%]
[String]
[('A')]
```
Indicating that even though "A" is displayed as 'A', the latter is neither a String nor a Char.

There are several fixes for that, the first one being to displaying "A" as "A".  In addition I would suggest to introduce a Char primary type that would be parsed with single quotes like traditional functional programming languages such as Haskell, Idris and Racket.

If you agree to that I don't mind taking care of it.

ngeiswei:
Fixes #289

### DIFF
--- a/python/hyperon/atoms.py
+++ b/python/hyperon/atoms.py
@@ -140,6 +140,12 @@ class GroundedObject:
         self.id = id
 
     def __repr__(self):
+        # Overwrite Python default representation of a string to use
+        # double quotes instead of single quotes.
+        if isinstance(self.content, str):
+            return f'"{self.content}"'
+
+        # Use default representation for everything else
         return repr(self.content) if self.id is None else self.id
 
     def copy(self):
--- a/python/hyperon/atoms.py
+++ b/python/hyperon/atoms.py
@@ -2,31 +2,6 @@ import hyperonpy as hp
 from hyperonpy import AtomKind
 from typing import Union
 
-
-class Char:
-    """Emulate Char type as in a traditional FPL"""
-    def __init__(self, char):
-        if len(char) != 1:
-            raise ValueError("A Char object must be initialized with a single character.")
-        self.char = char
-
-    def __str__(self):
-        return self.char
-
-    def __repr__(self):
-        return f"'{self.char}'"
-
-    def __eq__(self, other):
-        if isinstance(other, Char):
-            return self.char == other.char
-        elif isinstance(other, str):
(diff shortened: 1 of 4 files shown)


python/hyperon/atoms.py
```diff
@@ -2,6 +2,31 @@
...
+
+    def __eq__(self, other):
+        if isinstance(other, Char):
+            return self.char == other.char
+        elif isinstance(other, str):
+            return self.char == other
+        return False
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
```
vsbogd, line 27: I believe it should be the default behavior of `__ne__`, so we probably don't need it?
```suggestion

    def __ne__(self, other):
        return not self.__eq__(other)
```

vsbogd, line 27: I believe it should be the default behavior of `__ne__`, so we probably don't need a separate implementation:
```suggestion

    def __ne__(self, other):
        return not self.__eq__(other)
```

python/hyperon/atoms.py
```diff
@@ -2,6 +2,31 @@
 from hyperonpy import AtomKind
 from typing import Union
 
+
+class Char:
```
vsbogd, line 6: I would move this class into `stdlib.py` because it is not a part of a core library but a part of the MeTTa `stdlib` instead.

vsbogd:
Added @Necr0x0Der to review as well, because he uses Python library more intensive.

ngeiswei:
@vsbogd, I took into account your feedback (thank you!).  In addition I had to move some unit tests from `test_atoms.py` to `test_examples.py`.

python/hyperon/atoms.py
vsbogd: Yes, I missed this, but you are right.

python/hyperon/atoms.py
vsbogd: Yes, I don't think we need separate file, it is a tiny class.

vsbogd (approved):
Looks good to me.

Necr0x0Der (approved):

Necr0x0Der:
I've merged this PR. The only thing I want to mention is that we may want to have this type in Rust version of `stdlib` as well. What do you think?

PR #294, opened 2023-05-11
Statically link hyperonc library to the hyperonpy.so library

vsbogd:
This change is required to simplify packaging the Python hyperon library into a wheel package.
While it is possible to pack shared libraries the method is cumbersome and requires manual library loading from the custom path.

@wenwei-dev FYI


c/CMakeLists.txt
```diff
@@ -37,20 +43,28 @@ add_custom_target(build-hyperonc ALL
...
+set_target_properties(hyperonc-static PROPERTIES
+    IMPORTED_LOCATION "${HYPERONC_STATIC_LIB_PATH}"
+    INTERFACE_INCLUDE_DIRECTORIES "${RUST_TARGET_DIR}"
+    )
+add_dependencies(hyperonc-static build-hyperonc)
 
 add_subdirectory(tests)
 
-install(FILES ${HYPERONC_LIB_PATH} DESTINATION lib)
+install(FILES ${HYPERONC_SHARED_LIB_PATH} ${HYPERONC_STATIC_LIB_PATH} DESTINATION lib)
```
luketpeterson, line 67: Do we want to be installing both the hyperonc dylib and the static lib, for outside clients?

It is probably a non-issue at this point because hyperon is pretty self-contained.  But in principle we probably want to decide on the recommended way for C apps to link to hyperon.

luketpeterson (commented):
On the whole I think this makes good sense, and I don't see any problems.

My only question concerns whether it makes sense to direct external C clients (aside from Hyperononpy which is under our control) to statically link hyperonc, or whether we'd prefer they use the dylib.

luketpeterson (dismissed):

luketpeterson:
I'll leave it up to you whether to merge as is, or to edit the install command

c/CMakeLists.txt
vsbogd: I think the user may need either static or shared library according to his use-case. As an example Python native packages usually use static linkage. I would prefer shared library by default to have a possibility of reusing it. If we would have standard cmake build we could use `BUILD_SHARED_LIBS` option to control which artifact to build. But in our case `cargo` builds both at the same time.

But I agree we can start from installing a shared library for now and if we need a static library we modify cmake scripts to do it properly.

PR #297, opened 2023-05-17
Improve parsing Number

ngeiswei:
This add cases for negative and exponential numbers.


ngeiswei:
Unit tests are still missing.  Do you know what is the best place to add them?

vsbogd:
Thank you for fixing it.

> Unit tests are still missing. Do you know where is the best place to add them?

I would add a separate file `test_stdlib.py` and put unit tests inside.

vsbogd (dismissed):

ngeiswei:
@vsbogd, I'll write the unit tests on Friday, so please leave it as it is until then.

ngeiswei:
I've added the unit tests for number parsing, @vsbogd.  Note however that I had to use an operation, such as `!(+ 1 2)`, cause metta isn't able to just evaluate "!3".  It's not a big deal by any means, but I thought I'd mention it.

vsbogd (approved):

vsbogd:
> cause metta isn't able to just evaluate !3

The root cause is that the parser handles `!3` as a single symbol `Atom::sym("!3")` and as it is a single atom and not the pair `Atom::sym("!")`, `Atom::value(3)` it just adds it into the space without trying interpreting it.

vsbogd:
I have opened https://github.com/trueagi-io/hyperon-experimental/issues/303 to track the issue with non-expression atoms execution.

PR #299, opened 2023-05-18
Add filter_type() function into AtomIter to filter sub-atoms by types

vsbogd:
This removes a code duplication and adds a standard way of filtering vars as it is a frequent operation.
Addresses https://github.com/trueagi-io/hyperon-experimental/pull/298#discussion_r1197338902


luketpeterson (approved):
Looks good to me.

PR #300, opened 2023-05-18
Add ReplacingMapper utility to reuse variable mapping code

vsbogd:
Kind of experimental refactoring to reuse the variable mapping code which was written in three different places.

After writing this I think that may be the better way is to allow collecting `AtomIter` back into the atom, then one can use traditional `Iterator::map` instead of `Iterator::for_each(...)`. But it requires keeping the structure of the expression inside the iterated items, so I think we can do it in the future if we want.

### DIFF
--- a/lib/src/atom/mod.rs
+++ b/lib/src/atom/mod.rs
@@ -268,7 +268,7 @@ impl Display for VariableAtom {
     }
 }
 
-/// Returns a copy of `atom` with all variables replaced by unique instances.
+/// Returns `atom` with all variables replaced by unique instances.
 pub fn make_variables_unique(mut atom: Atom) -> Atom {
     let mut mapper = ReplacingMapper::new(VariableAtom::make_unique);
     atom.iter_mut().filter_type::<&mut VariableAtom>().for_each(|var| mapper.replace(var));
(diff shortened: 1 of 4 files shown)


lib/src/atom/mod.rs
```diff
@@ -269,17 +270,8 @@ impl Display for VariableAtom {
 
 /// Returns a copy of `atom` with all variables replaced by unique instances.
```
luketpeterson, line 271:
```suggestion
/// Returns `atom` with all variables replaced by unique instances.
```

luketpeterson (approved):

luketpeterson:
I wasn't sure if, from your message, you meant the original code is better without ReplacingMapper factored out into a separate object, or if you meant you tried some additional experimental decoupling, which didn't work out.  Feel free to merge or discard this PR.

vsbogd:
> I wasn't sure if, from your message, you meant the original code is better without ReplacingMapper factored out into a separate object, or if you meant you tried some additional experimental decoupling, which didn't work out. Feel free to merge or discard this PR.

I mean I have tried an additional experiment with splitting `ReplacingMapper` onto `CachingMapper` which maps `T -> R` and `ReplacingMapper` without `HashMap` inside. But this decoupling makes the code more cumbersome, so I don't add it to the PR.

luketpeterson (approved):

PR #306, opened 2023-05-24
Canonicalize path before loading module to make sure it is unique

ngeiswei:
Two syntactically different paths pointing to the same destination will not be wrongly qualified as different after being canonicalized.


ngeiswei:
I'm closing, it is not handling with grace the case where canonicalizing fails.  And it's missing unit tests as well.

Issue #305, opened 2023-05-23
Some imports are redundant


ngeiswei:
## Overview

Some metta file imports lead to extraneous non-determinism. 

## How to reproduce

Unpack the following archive

[folder.tar.gz](https://github.com/trueagi-io/hyperon-experimental/files/11545740/folder.tar.gz)

which contains a folder called `folder` with `A.metta`, `B.metta`, `C.metta`, `D.metta` and `E.metta` under it.

`A.metta` contains the definition of a function `foo`, while `B.metta` to `E.metta` contains various ways to import `A.metta` and call `foo`.

`B.metta` and `C.metta` are just here to show that `import!` works as expected.  Then `D.metta` fails, seemingly because of a convoluted import path of `A.metta`.  `E.metta` also fails, possibly for a different reason.

## What should normally be expected

In `D.metta` and `E.metta`, `!(foo 41)` should output only one result, `[42]`.

## What is observed instead

`D.metta` and `E.metta` outputs the result multiple times `[42, 42, 42, 42, 42, 42, 42, 42]`, indicating redundancy in the non-deterministic interpretation.

ngeiswei:
@vsbogd, I need to have this fixed in order to move forward, so I have assigned myself to it, but if you happen to have already started to look into it and have some insights to share, please feel free to do so.

Necr0x0Der:
> @vsbogd, I need to have this fixed in order to move forward, so I have assigned myself to it, but if you happen to have already started to look into it and have some insights to share, please feel free to do so.

Great, thanks! The core implementation is in `impl Grounded for ImportOp {` (stdlib.rs).
Also, take a look at `modules: Shared<HashMap<PathBuf, Shared<GroundingSpace>>>` in metta/runner/mod.rs .
 Feel free to ask questions if the logic behind the code is not clear.

PR #307, opened 2023-05-24
Canonicalize path before loading module to make sure it is unique

ngeiswei:
Two syntactically different paths pointing to the same destination will not be wrongly qualified as different after being canonicalized.

### DIFF
--- a/lib/src/metta/runner/stdlib.rs
+++ b/lib/src/metta/runner/stdlib.rs
@@ -66,6 +66,7 @@ impl Grounded for ImportOp {
         // TODO: replace Symbol by grounded String?
         if let Atom::Symbol(file) = file {
             path.push(file.name());
+            path = path.canonicalize().unwrap_or(path);
             log::debug!("import! load file, full path: {}", path.display());
         } else {
             return Err("import! expects a file path as a second argument".into())


ngeiswei:
OK, that version handles a failure of `canonicalize` "gracefully".  Well, it just ignores the canonicalizing if it fails, which is OK because it is then caught by `load_module_space`.

Although I believe it is an OK fix, I wonder if I should have moved that canonicalizing inside `load_module_space`.  I'm neutral on that, but if you feel it belongs there, let me know.

Also, that is only a partial fix for #305, it does not handle the situation where `!(import! A.metta)` is consecutively called twice.  I'll see if I can figure that one out as well.

ngeiswei:
Oh, I forgot to add a unit test!  I suppose it could be within `stdlib.rs`, on the other hand it requires to create metta files, so maybe it should be under `python/tests` instead, what do you think?

ngeiswei:
I'm seeing that `f1_imports.metta` runs as test, so I suppose a unit test could be added in there.

ngeiswei:
Sorry for impulsively creating then closing issues.  I believe I now know what to do I prefer to close that one and create a new one when I have a complete solution with tests.

vsbogd:
No problem Nil :+1:

PR #314, opened 2023-05-30
Init logger in a Python API library

vsbogd:
PR fixes issue with logs absence in debug releases. The root cause is that `ctor` is not called when C API static library is built in debug configuration. I am calling `init_logger` manually from C library to fix it.

 I opened https://github.com/mmastrac/rust-ctor/issues/280 to ask `ctor` author for help in investigation.

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -9,7 +9,7 @@ mopa = "0.2.2"
 regex = "1.5.4"
 log = "0.4.0"
 env_logger = "0.8.4"
-ctor = "0.1.22"
+ctor = "0.2.0"
 smallvec = "1.10.0"
 
 [lib]
--- a/python/hyperonpy.cpp
+++ b/python/hyperonpy.cpp
@@ -236,9 +236,6 @@ struct CAtomType {};
 PYBIND11_MODULE(hyperonpy, m) {
     m.doc() = "Python API of the Hyperon library";
 
-    // TODO: integrate Rust logs with Python logger
-    m.def("init_logger", &init_logger, "Initialize Hyperon library logger");
-
     py::enum_<atom_type_t>(m, "AtomKind")
         .value("SYMBOL", atom_type_t::SYMBOL)
         .value("VARIABLE", atom_type_t::VARIABLE)
@@ -496,5 +493,11 @@ PYBIND11_MODULE(hyperonpy, m) {
         metta_load_module(metta.ptr, text.c_str());
     }, "Load MeTTa module");
 
+}
+
+__attribute__((constructor))
+static void init_library() {
+    // TODO: integrate Rust logs with Python logger
+    init_logger();
 }
 


luketpeterson:
This change looks reasonable enough.  But I would like to hear what @mmastrac (ctor author) has to say about the root cause.  It looks like he's very responsive so my guess is we'll hear back soon.

luketpeterson:
Does it make sense to invent a persistent object for all Hyperon clients?  I.e. a runtime object that must be created to use the library?

I think we might need something like this to own the thread pool for a multi-threaded interpreter, and I was considering turning all symbol atoms into hashes for performance, and keeping the strings themselves in a central store.  Things like that.

If we had such an object, that would give us a place to initialize the logger without relying on ctor.

vsbogd:
> Does it make sense to invent a persistent object for all Hyperon Rust clients?

Luke, do you have any examples of this used in mature Rust libraries? I am not crazy of having global singletons except some specific cases. Even with logger it definitely has drawbacks, for example one cannot properly configure logger from code from the very beginning. But maybe in Rust there is no other way when one need to initialize some global constant object before program started. For example it is the reason why reserved symbol atoms are `const` instead of `static`: because I didn't found an easy way to initialize them beforehand.

Anyway I would not create some global object which keeps all what we need at once. I would instead introduce specific APIs for each case and use it to initialize storages. For example address of the string could probably be used instead of its hash and it would not require a separate store. Thread pool could be a part of interpreter and initialized with its instance. Maybe there are cases where we cannot eliminate the global storage. So it is also interesting to see how this problem solved in mature Rust libraries.

ngeiswei (approved):
I cannot comment on design of that fix but I can confirm that it works for me.

luketpeterson:
TLDR: this change is good to merge, IMO.  But longer-term, I believe we should remove the logger init from the ctor hook, and try to put it somewhere else.

Many embedded language crates use a top-level runtime state object (often a singleton):

- JavaScript (V8 engine): https://docs.rs/v8/0.73.0/v8/V8/fn.initialize.html
- Python: https://pyo3.rs/main/features.html#features-for-embedding-python-in-rust
- Lua https://docs.rs/mlua/0.8.9/mlua/#the-lua-object

However, I agree with you that a singleton might not be the best design.  All of those above examples are rust wrappers around projects with non-rust implementations, and their decision might have been involuntary, rather than a well-considered design choice.

I think the Lua case is the closest to what I have in mind.  There is a top-level object but it’s not a singleton.

Hyperon-experimental has the `runner::Metta` type, but the big difference is that it isn’t created first.  Instead, it is created with a `Space` (and a `Tokenizer`), which might involve a lot of work to create them in advance.  More generally, the current design lets the caller create Atoms and Spaces by themselves without reference to any runner object. (‘static lifetime)

This might end up limiting the optimization that are available to us - although we could solve it like V8 solves it, by creating wrappers for objects under the management of a runner (ie. something like https://docs.rs/v8/0.73.0/v8/trait.Handle.html)

Anyway, this is a bigger discussion, and I apologize for getting us off-topic with regard to logger initialization.  If we ultimately moved in this direction, top-level object creation would be a logical place to init the logger.

My final point about the logger is that running the initialization from ctor means that another app that wants to embed the MeTTa interpreter can’t initialize their own logger from their own main (if they wanted to use `fern` (https://docs.rs/fern/0.6.2/fern/) for example), because the MeTTa logger would be initialized first.  Therefore, my preference would be to give the caller an option to do stuff from main() before MeTTa installs its logger.

vsbogd:
Thanks Luke, on the first glance Lua example is clear, but I think I need to look at the API of the library closer to understand how it works in details.

I am curious why Rust has a such restricted conception of a static read-only object.As far as I remember static can be initialized by `const` functions only and `const` functions mean one cannot allocate memory.

vsbogd:
Merging it as a quickfix as this functionality is important, but I am ready to discuss a proper fix further.

PR #315, opened 2023-05-30
Return variable from Bindings::resolve() when no value is found

### LINKED ISSUE #313: The MeTTa interpreter crashes potentially leaky evaluation
## Overview

The MeTTa interpreter crashes seemingly when calling a function with a variable in argument, and unification with that function's definition and is not sufficient to carry out the reduction.

## How to reproduce

Run the following metta code

```
(= (foo (A $x)) $x)
!(foo $y)
```

## What should be expected

One possible output could be `[$x#1]`, but I'm not sure if that would be permitted.

## What do we get instead

The interpreter crashes with the error message

```
thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', /home/nilg/Work/TrueAGI/hyperon-experimental/lib/src/metta/interpreter.rs:573:67
Aborted (core dumped)
```

vsbogd:
Quick fix for #313. Return None from Bindings::resolve() when variable is absent. Return variable itself when variable has no value.

### DIFF
--- a/lib/src/atom/matcher.rs
+++ b/lib/src/atom/matcher.rs
@@ -166,7 +166,7 @@ impl Bindings {
                     let replacing_var = self.var_by_id(var_id, |alt| *alt != *var);
                     match  replacing_var {
                         Some(var) => VarResolutionResult::Some(Atom::Variable(var.clone())),
-                        None => VarResolutionResult::None,
+                        None => VarResolutionResult::Some(Atom::Variable(var.clone())),
                     }
                 },
             };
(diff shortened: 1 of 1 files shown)


vsbogd:
@ngeiswei , could you please check it it works for the rest of your code. I am afraid it can break something. 
I need to write down the better long term fix for the `Bindings` logic, but it will take more time.

ngeiswei (approved):
> I need to write down the better long term fix for the Bindings logic, but it will take more time.

@vsbogd, how different would be your long term fix compared to this one?  From what I can tell it fixes the problem without introducing undesirable behaviors.

vsbogd:
Better fix I am talking about is more a refactoring of `Bindings` API mentioned in https://github.com/trueagi-io/hyperon-experimental/pull/298#issuecomment-1551995900.

More specifically I think we could add a method to convert `Bindings` into a strict map of variables to atoms and back (opposite direction conversion is already implemented as `Bindings::from(Vec<(Var, Atom)>)`). For example `{ $x = $y = $z <- A }` could be converted into `{ $x <- $y, $y <- $z, $z <- A }'. This conversion solves the ambiguities which appear inside `Bindings` which contain variable equalities. For example in `{ $x = $y }` what should we return as a value for `$x`? MeTTa code expects that for both `$x` and `$y` the same variable is returned. Converting it into `{ $x <- $y }` may make the resolution unambiguous. Also this conversion could at the same time filter out all extra variables (which can be done using `Bindings::narrow_vars`).

But this refactoring doesn't contradict with the #315, so I think we can merge it.

PR #318, opened 2023-06-02
Changing c_str api to follow the snprintf() pattern

luketpeterson:
Background:  @luketpeterson wrote:

>>  ## The `c_str_callback_t` pattern is less efficient than returning a malloced buffer.
>>
>> Regardless of the safer_ffi migration, I think we should eliminate the c_str_callback_t callback.
>>
>> My reasoning goes like this: The Rust &str & String types don’t include a trailing terminator character because the Rust types track a size.  The callback gets a non-owned read-only c_str argument, which the implementation created by allocating buffer and copying the string into the buffer, plus the terminator.  Then, the callback needs to do something with the string, and the most common thing is to allocate yet another buffer and copy the string again.
>>
>> We can save one of these copies by just returning an allocated buffer - not to mention making the API simpler to use.

To which @vsbogd replied:

> The only issue with returning the allocated string from Rust is that a C API user will be obliged to free it with the Rust specific function. Thus as far as I understand one cannot just get the `[u8]` buffer from Rust and use it as a usual `char*` (at least cannot free it by `free()`). Please correct me if I am wrong here. But I don't object, we can implement `string_t = *mut c_char` in C API and add `string_free` function.

Anyhow, it is possible to use the `libc` Rust crate, and get access to exactly the same allocator functions used by the code.  For example: https://docs.rs/libc/latest/libc/fn.malloc.html  So we could allocate a pointer in Rust that the C side could free with an ordinary `free()`.

However, I thought about it a little more, and returning an allocated buffer is an improvement over two allocations in every case.  But a bigger improvement would be to let the C side create the buffer (on the stack) and then have Rust write into it.  The `snprintf()` API pattern.  That's what this PR does.

It's not a big deal, performance-wise.  But it's also not a very big change and it simplifies the C API from the caller's perspective.

### DIFF
--- a/c/src/atom.rs
+++ b/c/src/atom.rs
@@ -97,11 +97,14 @@ pub extern "C" fn bindings_clone(bindings: *const bindings_t) -> *mut bindings_t
     bindings_into_ptr(unsafe{ &(*bindings) }.bindings.clone())
 }
 
+/// Writes a text description of the bindings_t into the provided buffer and returns the number of bytes
+/// written, excluding the string terminator.
+///
+/// Returns 0 if the buffer is too small
 #[no_mangle]
-pub extern "C" fn bindings_to_str(bindings: *const bindings_t, callback: c_str_callback_t, context: *mut c_void) {
+pub extern "C" fn bindings_to_str(bindings: *const bindings_t, buf: *mut c_char, buf_len: usize) -> usize {
     let bindings = unsafe{ &(*bindings).bindings };
-    let s = str_as_cstr(bindings.to_string().as_str());
-    callback(s.as_ptr(), context);
+    write_into_buf(bindings, buf, buf_len)
 }
 
 #[no_mangle]
@@ -262,11 +265,14 @@ pub extern "C" fn bindings_set_eq(set: *const bindings_set_t, other: *const bind
     set == other
 }
 
+/// Writes a text description of the bindings_set_t into the provided buffer and returns the number of bytes
+/// written, excluding the string terminator.
+///
+/// Returns 0 if the buffer is too small
 #[no_mangle]
-pub extern "C" fn bindings_set_to_str(set: *const bindings_set_t, callback: c_str_callback_t, context: *mut c_void) {
+pub extern "C" fn bindings_set_to_str(set: *const bindings_set_t, buf: *mut c_char, buf_len: usize) -> usize {
     let set = unsafe{ &(*set).set };
-    let s = str_as_cstr(set.to_string().as_str());
-    callback(s.as_ptr(), context);
+    write_into_buf(set, buf, buf_len)
 }
 
 #[no_mangle]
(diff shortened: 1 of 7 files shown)


vsbogd:
Thanks for `libc` Luke, I didn't know about it but the answer was on a surface.

The problem with `sprintf` like interface: one need to get a size of the buffer which fits the string before calling it. Usually it requires two calls of the `sprintf`-like function: first with `NULL` buffer to get the size of the string, second with real buffer and real length. Thus using `get_str_ptr()` API should be a bit more compact in general.

For example we could write the function which adapts `get_str_ptr` to `sprintf` like looks like the following:
```c
void write_buf(char* buf, size_t len, char* src) {
  strncpy(buf, str, len);
  free(src);
}
```

The advantage of the current implementation: the buffer allocation is totally on a caller side and thus the caller knows how to free it.

luketpeterson:
I agree with you that the `snprinf()` pattern isn't the most convenient for the caller.  However, it is the only pattern that I know of that allows the caller to speculatively over-allocate or make use of a stack-allocated buffer to avoid allocation altogether.

I wanted to see how expensive small allocations are, in practice, so I quickly coded the following benchmark:
```rust
extern crate test;

use test::{Bencher, black_box};

use hyperon::*;

use core::slice;
use std::io::{Write, Cursor};

fn lots_of_atoms(num: usize) -> Vec<Atom> {
    let mut atom_vec = vec![];
    for i in 0..num {
        let new_atom = Atom::sym(format!("{i:10}")); //10 character atom names
        atom_vec.push(new_atom);
    }
    atom_vec
}

#[bench]
fn atom_name_to_string(bencher: &mut Bencher) {
    let atoms = lots_of_atoms(10000); //Ten thousand atoms
    bencher.iter(|| black_box({
        for atom in atoms.iter() {
            if let Atom::Symbol(sym) = atom {
                let name_str = sym.name().to_string();
                assert_eq!(name_str.len(), 10)
            }
        }
    }))
}

#[bench]
fn atom_name_into_buf(bencher: &mut Bencher) {
    let atoms = lots_of_atoms(10000); //Ten thousand atoms
    let mut buf = [0u8; 2048];
    bencher.iter(|| black_box({
        for atom in atoms.iter() {
            let slice = unsafe{ slice::from_raw_parts_mut(&mut buf as *mut u8, buf.len()) };
            let mut cursor = Cursor::new(slice);

            if let Atom::Symbol(sym) = atom {
                write!(cursor, "{}", sym.name()).unwrap();
            }
            assert_eq!(cursor.position() as usize, 10);
        }
    }))
}
```
Here are my results:
```
running 2 tests
test atom_name_into_buf  ... bench:     161,257 ns/iter (+/- 15,464)
test atom_name_to_string ... bench:     553,583 ns/iter (+/- 102,728)
```
So the allocation approximately 3x the cost.  Proportionally that's big, but overall the numbers here are tiny so it's going to make absolutely no difference in overall runtime.

The reason I made the change at all is part of a larger philosophy that unnecessary allocations are very costly and any software that runs optimally needs to control where and how the memory management is happening.  This becomes doubly true in a multi-threading environment since there is usually a mutex around certain allocator operations and that can end up causing otherwise parallel code to stall.

luketpeterson:
I realized my benchmarks weren't quite fair...  That is, the atom name is in a &str already so there is no need to call through the `fmt` code.  So after replacing:
```rust
write!(cursor, "{}", sym.name()).unwrap();
```
with
```rust
cursor.write(sym.name().as_bytes()).unwrap();
```
Here are the new numbers:
```
running 2 tests
test atom_name_into_buf  ... bench:      37,628 ns/iter (+/- 7,056)
test atom_name_to_string ... bench:     507,757 ns/iter (+/- 138,932)
```

vsbogd:
I agree that when we don't need passing the string out of the function we can keep the data on the stack and achieve zero cost allocation. In practice I think the input/output from C code could be implemented by this way.

But in `hyperonpy.cpp` second allocation happens anyway. When the buffer is allocated on a stack then it should be reallocated when `std::string` instance is created. And when the pre-allocated buffer is not enough, then formatting is called two times which can be much worse for the complex objects. Thus I would suggest simplify the `func_to_string` code to:

```c++
std::string func_to_string(write_to_buf_func_t func, void* arg) {
    size_t len = func(arg, nullptr, 0);
    std::string new_string = std::string();
    new_string.resize(len+1); //Room for the terminator in the allocated buffer
    func(arg, new_string.data(), len+1);
    new_string.resize(len); //But the C++ string doesn't include the terminator in its len
    return new_string;
}
```

Also looking at `func_to_string` code I see that we can use it as a conversion from `sprintf` to `get_str_ptr` semantics in C (unfortunately the number of arguments should be known in advance). Thus we can use pre-allocated buffer when it is possible and use `_ptr` when pre-allocated buffer is not possible. Maybe it is worth to add it as an API function. But I would change the name to something like `write_to_buf_into_ptr` to emphasize its role:

```c
typedef size_t (*write_to_buf_func_t)(void*, char*, size_t);

char* write_to_buf_as_ptr(write_to_buf_func_t func, void* arg) {
    size_t len = func(arg, nullptr, 0);
    char* str = (char*)malloc(len + 1);
    if (str) {
        func(arg, str, len + 1);
    }
    return str;
}
```

Looks like we cannot reuse it in order to implement `func_to_string` because it is not possible to create an instance of `std::string` from pre-allocated buffer, but maybe we could use it to create a Python string instead.

luketpeterson:
> But in hyperonpy.cpp second allocation happens anyway...

I don't have a strong opinion here.  On one hand, stack allocation (and `alloca`) are practically free, compared with the cost of dynamic allocation through `malloc`.  On the other hand, starting with a buffer, partly filling it up, and then starting over, will be slower than just doing a first pass to determine the size needed.

The approach that would be faster overall comes down to what proportion of the calls fit on the fast path (simple stack buffer) vs. what proportion have to go to the slow path.

But there is a larger philosophical point that actually makes me not care - Python itself is a very slow language.  So, unlike C, when a user is using Python they've already surrendered the performance battle.  So I'll fight a lot harder for performance through the C & Rust interfaces than through the Python interface.

vsbogd (approved):
One additional argument from Luke in my words: even while  `func_to_string` allocates the memory it still can be a bit faster in a small string case because it processes formatting only once versus guaranteed double formatting in case when size of the buffer is not enough.

I agree that we keep `func_to_string` in a current state. I also agree `sprintf`-like API can be used to optimize a code in specific situations. `sprintf` like API can be converted into `_get_ptr` API using `func_to_string` like function. I am ready to merge it. Thanks Luke.

Issue #308, opened 2023-05-25
Add interface to create custom Space implementation in Python


luketpeterson:
Building on the CSpace, allow Python code to define a new Space.

vsbogd:
One example of similar API which is already implemented is a Python `GroundedObject` class. It has a C++ `GroundedObject` adapter class in `hyperonpy.cpp` file which adapts Python successors of the `GroundedObject` to the C and Rust API: https://github.com/trueagi-io/hyperon-experimental/blob/4f9b774c90445bf5de8c50b4cc74d9c874d6970a/python/hyperonpy.cpp#L78-L95

PR #322, opened 2023-06-07
Adding glue to allow spaces implemented in Python. 

luketpeterson:
https://github.com/trueagi-io/hyperon-experimental/issues/308


python/hyperon/base.py
```diff
@@ -2,24 +2,57 @@
 
 from .atoms import Atom
 
-class GroundingSpace:
+class PySpace:
```
vsbogd, line 5: I would suggest allow calling methods of the Python space implementations directly without proxying them through C API. Then current `Space` class could be renamed to the `RustSpace` to designate that it proxies the spaces implemented in Rust. `Space` could be new abstract class which has all methods throwing "not implemented" exception. `GroundingSpace` could be a successor of `RustSpace`. And `PySpace` is a successor of `Space`. Please see the code example here: https://github.com/vsbogd/hyperon-experimental/commit/44093d299c5fbf4ba99cf2731aba29a4d9ced086

In such implementation we don't need `get_payload` because user can get data directly from the specific class. `copy` is also not needed. `query` should return `BindingsSet` instead of map in previous implementation. This is not implemented in the commit above and it is why one of the tests fails.  Also methods of the `RustSpace` could be renamed in order to have same names as in Rust and in `Space`.

vsbogd: @Necr0x0Der , could you please also take a look at the Python API.

python/hyperonpy.cpp
```diff
@@ -188,6 +188,125 @@ void py_free(struct gnd_t* _cgnd) {
...
+    call_add_on_python_space(pyobj, catom);
+}
+
+bool py_space_remove(const struct space_params_t *params, const struct atom_t *atom) {
+    py::object hyperon = py::module_::import("hyperon");
+    py::function call_remove_on_python_space = hyperon.attr("call_remove_on_python_space");
+    py::object pyobj = static_cast<PySpace const *>(params->payload)->pyobj;
+    CAtom catom = atom_clone(atom);
+    py::object result = call_remove_on_python_space(pyobj, catom);
+    return py::str(result).is(py::str(Py_True)); //Really?? The best way to test bools is using strings??
```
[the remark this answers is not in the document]
luketpeterson: I read a blog that said Python can represent `True` and `False` several different ways each, and the safe way is to stringify it.  But now that I think about that again, it seems very dumb.  Given we care about performance (at least as much as Python lets us care) we can make the Python programmers use the one true `True`.

python/hyperonpy.cpp
```diff
@@ -387,18 +518,25 @@ PYBIND11_MODULE(hyperonpy, m) {
...
     m.def("space_eq", [](CSpace a, CSpace b) { return space_eq(a.ptr, b.ptr); }, "Check if two spaces are equal");
-    m.def("space_len", [](CSpace space) { return space_atom_count(space.ptr); }, "Return number of atoms in space, or -1 if the space is unable to determine the value");
+    m.def("space_atom_count", [](CSpace space) { return space_atom_count(space.ptr); }, "Return number of atoms in space, or -1 if the space is unable to determine the value");
     m.def("space_list", [](CSpace space) -> std::optional<pybind11::list> {
         pybind11::list atoms_list;
         if (space_iterate(space.ptr, atom_copy_to_list_callback, &atoms_list)) {
             return atoms_list;
         } else {
-            return pybind11::none();
+            return std::nullopt;
```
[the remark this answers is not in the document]
luketpeterson: Actually the original code had a bug, that was masked by the details of the test.

My intention with this code was that space.get_atoms() would return None if that wasn't a supported operation.  But with the code being the way it was, it was attempting to iterate the None value inside of `get_atoms` and throwing an exception before control ever got back to the caller.

python/hyperonpy.cpp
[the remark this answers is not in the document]
luketpeterson: I committed your suggestion in 364ce13

python/hyperon/base.py
luketpeterson: Ultimately it is your (and Alexey's) choice for big architectural decisions like this.  However I feel that the system will be more maintainable and have fewer bugs if we channel and centralize functional pathways as much as possible.

By calling through the Core (Rust) interfaces, we will get full interoperability between spaces regardless of which language they are implemented in.  I.e. a Rust client wouldn’t even need to care if a space is implemented in Python or C, assuming the module-loading is done.

Also this gives us the freedom to handle certain functionality at lower levels where appropriate.  For example, if we call through the core, the Python code doesn't need to handle observer registration tracking because it’s done in the base layer.  As we evolve the Space implementation, I believe this will become increasingly important.

luketpeterson: Am I misunderstanding your proposal?

python/hyperonpy.cpp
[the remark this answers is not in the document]
vsbogd: Hm, interesting information about Python, but I don't think I ever faced it. Let's try using a clean simple way and improve it if we find an issue.

Necr0x0Der:
I see no problem with API, but I have had a particular issue with the previous API, which might hopefully be solvable with this new one. When we are coding in Python and creating various spaces, we have to run `(new-space)` from MeTTa. It is not just possible to use
```
m = MeTTa()
s = GroundingSpace()
m.space().add_atom(E(S("my-space"), s)) # or G(s)
```
`GroundingSpace` is not a grounded atom by itself. An attempt to wrap it into `G` doesn't work, because it doesn't implement `copy`. But even if it did implement `copy`, then it would not be possible to use such wrapped `GroundingSpace` in `match` expressions in MeTTa, because `match` expects not just a grounded atom, but a Rust grounding space atom. With implementing `Space` as a trait of grounded atoms, we are much closer to calling `match` on `PySpace`.  It can work out of the box or needs just a minor wrapper, it would be nice to know. If there are still some problems with this, the approach to mitigate them should be discussed.

python/hyperon/base.py
vsbogd: > By calling through the Core (Rust) interfaces, we will get full interoperability between spaces regardless of which language they are implemented in.

This point is true for my suggestion as well. I just would like to simplify the chain of calls when Python Space is called from Python:
- in current code it is: `client -> Space -> hyperonpy -> C API -> hyperonpy -> PySpace`
- in my suggestion it is: `client -> Space`

When call is made from Rust it is identical in both cases (when space is implemented in Python):
`client -> C API -> hyperonpy -> PySpace`

> Also this gives us the freedom to handle certain functionality at lower levels where appropriate. For example, if we call through the core, the Python code doesn't need to handle observer registration tracking because it’s done in the base layer.

I would keep Rust and Python implementations similar here. In Rust we have `Space` trait and notifying observers is implemented by embedding `SpaceCommon` and calling it from methods.  I think Python spaces could be in the same position here.

If we would like to automatically notify observers when add/remove/replace are called we probably should embed custom space implementation into `SpaceCommon` instead (as it is done in this PR). And in this case I still don't see the necessity of calling PySpace methods via `client -> Space -> hyperonpy -> C API -> hyperonpy -> PySpace` chain. We can implement notifications in `SpaceCommon::add` after calling the `PySpace::add` method. If words are confusing I can write a snippet bit later.

vsbogd: Basically the idea I tried to express in a last paragraph in Python:
```py
class Space:

    def __init__(self, space_obj):
        self.cspace = hp.space_new_custom(space_obj)

    def __del__(self):
        hp.space_free(self.cspace)

    # other methods

    def add_atom(self, atom):
        self.space_obj.add(atom)
        hp.space_notify_observers(...)

    # all other methods
```

luketpeterson: I think I see what you have in mind.  With the current code structure, it will be no problem to change things to the way you propose.  However, looking to the future, It means the space management logic needs to remain very thin at the Rust level if Python clients will be allowed to shortcut past it.

What if the benefit of providing the shortcut calling path?  Is it for performance, or some other functionality / design feature you have in mind?

This is a philosophical choice.  If it were me, I would prefer not to allow the shortcut.  My reason is based on a belief that we might want to create a tighter coupling between a top-level host object and the Spaces in the future.  But if spaces remain self-contained and fully owned by the code that instantiates them, then your proposal should be fine.

Necr0x0Der:
PS basically, what I mean is that this PR is lacking an example of putting a custom space atom into self or another space, so we can invoke matching for it from MeTTa itself in addition to calling directly `kb.query(E(S("A"), V("x")))` in Python

vsbogd:
@Necr0x0Der, am I right that the problem you are talking about can be caught by the following test:
```py
    def test_match_custom_space(self):
        nested = GroundingSpace()
        nested.add_atom(E(S("A"), S("B")))
        gnd = ValueAtom(nested)

        runner = MeTTa()
        runner.space().add_atom(gnd)
        runner.tokenizer().register_token("nested", lambda token: gnd)

        result = runner.run("!(match nested (A $x) $x)")
        self.assertEqual([[S("B")]], result)
```
The output of the test:
```console
======================================================================
FAIL: test_match_custom_space (__main__.MettaTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test_metta.py", line 56, in test_match_custom_space
    self.assertEqual([[S("B")]], result)
AssertionError: Lists differ: [[B]] != [[(Error (match <hyperon.base.GroundingSpac[80 chars]nt)]]

First differing element 0:
[B]
[(Error (match <hyperon.base.GroundingSpac[79 chars]ent)]

- [[B]]
+ [[(Error (match <hyperon.base.GroundingSpace object at 0x7fcf100969d0> (A $x) $x) match expects a space as the first argument)]]

----------------------------------------------------------------------
Ran 1 test in 0.011s

FAILED (failures=1)
```
If so this issue is probably not directly related to this PR, while this PR for sure makes it more important. The problem is that `MatchOp` implementation in `stdlib.rs` expects a specific class and I not sure how we could fix it in Rust.

Necr0x0Der:
> @Necr0x0Der, am I right that the problem you are talking about can be caught by the following test:
> 

Yes, this one way to show this. The problem is not that this specific test fails, but that it's unclear how to pass Python grounded space (possibly, custom) as an atom to `match` in MeTTa. E.g. `GroundingSpace` might be already a grounded atom acceptable by `match` without wrapping into `ValueAtom`, or there could be `SpaceAtom`, in which we could wrap `GroundingSpace` or anything else.

vsbogd (approved):
After discussing the API both options has strong and weak sides. I am ready to merge it and see if in future we have better arguments for one or another solution.

vsbogd:
Raised https://github.com/trueagi-io/hyperon-experimental/issues/328 to address the nested spaces issue.

PR #331, opened 2023-06-15
Consolidating MeTTa type definitions in one place for Atom Types and Grounded Atom Types

luketpeterson:
Originally, I had put the Python definition of ATOM_TYPE_GROUNDED_SPACE in atom.rs (precursor for the C API) and exported GroundedAtomType.SPACE to the Python API.

My thinking was to keep a separation from types returned by `atom_get_type()` and types returned by `atom_get_grounded_type()`.  However, it does make things a little more confusing.

Vitaly suggested putting the MeTTa types in the same place: https://github.com/trueagi-io/hyperon-experimental/pull/329#discussion_r1229642538

That is the purpose of this PR.

### DIFF
--- a/c/src/atom.rs
+++ b/c/src/atom.rs
@@ -49,8 +49,6 @@ pub struct bindings_set_t {
 pub type bindings_callback_t = lambda_t<*const bindings_t>;
 pub type bindings_mut_callback_t = lambda_t<*mut bindings_t>;
 
-#[no_mangle] pub extern "C" fn ATOM_TYPE_GROUNDED_SPACE() -> *mut atom_t { atom_into_ptr(rust_type_atom::<DynSpace>()) }
-
 #[repr(C)]
 pub struct gnd_api_t {
     // TODO: replace args by C array and ret by callback
--- a/c/src/metta.rs
+++ b/c/src/metta.rs
@@ -1,9 +1,11 @@
+use hyperon::space::DynSpace;
 use hyperon::Atom;
 use hyperon::metta::text::*;
 use hyperon::metta::interpreter;
 use hyperon::metta::interpreter::InterpretedAtom;
 use hyperon::common::plan::StepResult;
 use hyperon::metta::runner::Metta;
+use hyperon::rust_type_atom;
 
 use crate::util::*;
 use crate::atom::*;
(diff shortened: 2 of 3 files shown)


luketpeterson:
Agh... I just saw this after I pushed this PR.  https://github.com/trueagi-io/hyperon-experimental/pull/329#discussion_r1230870089

luketpeterson:
In my opinion it makes sense to keep separation between GroundedAtom types (sub-types within grounded atoms) and Atom types.  (Sym, Var, Expr, & Grounded).  However I will defer to your opinions because you have thought more deeply about this.

Necr0x0Der:
> Agh... I just saw this after I pushed this PR. [#329 (comment)](https://github.com/trueagi-io/hyperon-experimental/pull/329#discussion_r1230870089)

No problem. If you have already done this, then it will be easier to see if this is the right choice, and if everything else is ok

Necr0x0Der:
> In my opinion it makes sense to keep separation between GroundedAtom types (sub-types within grounded atoms) and Atom types. (Sym, Var, Expr, & Grounded). However I will defer to your opinions because you have thought more deeply about this.
> 
> However. Spaces are a little bit special as Grounded atoms go, because they are exposing a fundamental MeTTa construct in a different way - as opposed to exposing an external interface.

I agree that having Grounded Space as a normal type (rather than meta-language type like `Sym`, `Var`, etc.) may make sense. That's why I wanted to analyze this more carefully. While grounded space is special indeed, it doesn't look like a meta-type. because a space atom cannot have any other ordinary type besides of space type. At the same time, we can treat it is a singleton meta-type. The question is if there are any consequences to MeTTa after merging this PR. Are there some expressions, which will become valid or invalid after this? I'm not precisely sure.

vsbogd:
Well, sorry for the confusion guys. Actually `AtomType` is not a collection of meta-types. It is a collection of types which are used in MeTTa and it was created even before we agreed on not returning meta-types in the list of atom's type. Originally it is an analogue of this list of types in Python: https://github.com/trueagi-io/hyperon-experimental/blob/b5acba1905d9bb47a8ad336ae56285ef93865f9c/lib/src/metta/mod.rs#L15-L21 and thus maybe `atom.py` is not correct place for it, and  `MettaType` would be better name than `AtomType` . From this perspective for the `Space` type is totally correct to be there.

If @Necr0x0Der  is going to add some new types then I think we can merge it and fix it in further PRs.

vsbogd (approved):

PR #343, opened 2023-06-23
Automatically update hyperonpy module when Python hyperon package is installed in editable mode

vsbogd:
At the moment it is required to run `python3 -m pip install -e .[dev]` in order to update the `hyperonpy` module after some native code is changed. This PR changes the target directory for `hyperonpy` to `./python` which means that `hyperonpy` is automatically updated by `make` command.

Also this PR contains:
- add proper CMake package config for `hyperonc` library
- remove Python CMake installation because `pip` should be used instead
- stop uploading artifacts into GitHub internal storage as they are published to Release page
- rename GitHub Actions workflows and jobs to make them look better in GitHub UI

This PR is a step to `manylinux` package building.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -99,7 +99,7 @@ jobs:
         working-directory: ./python
         run: |
           python -m pip install build==0.10.0
-          python -m build
+          python -m build --wheel
           tar zcf dist-${{ inputs.os }}-py${{ inputs.python-version }}.tar.gz dist
 
       - name: Upload Python package to GitHub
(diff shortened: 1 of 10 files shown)


vsbogd:
@wenwei-dev , please take a look as I made some minor changes to workflows you introduced.

luketpeterson (approved):
In my opinion, build & config changes are the hardest to validate because it requires knowledge of the particular gotchas of different environments.  Also I am comparatively weak in my knowledge of Python best-practices.

However, for what it's worth, these changes all look ok to me and I ran them on both Mac OS and Debian, and I didn't hit any problems.

wenwei-dev:
Looking good @vsbogd

Issue #351, opened 2023-06-28
General-Purpose bridging mechanism for Grounded Atoms with native types


luketpeterson:
Grounded Atoms should have a facility to implement translation code into native types in their host language.

For example, consider a grounded atom that represents an integer.  If a Python client creates a grounded atom by wrapping an integer, it will be a native python integer, and therefore an Integer atom implemented in Rust won't be able to easily unwrap it and treat it as a Rust integer type.

Ultimately we probably want a generalized mechanism for allowing modules to implement plugins for translation in multiple languages.  However, that will be a lot of work.

I think we'll get a good proportion of the benefit by implementing our own internal facility for this translation for a useful set of grounded atoms in the MeTTa std lib.

More discussion on this topic is here:
https://github.com/trueagi-io/hyperon-experimental/issues/333

luketpeterson:
The Python and the Rust stdlibs each contain a number of declarations (numbers, True|False) that conflict and this is already starting to cause some pain.  So I estimate this will need to be tackled in the next couple months as the pain will only increase as the stdlib does.

vsbogd:
Well it is the reason why we didn't want to allow calling Python code from Rust REPL and vice versa before releasing Alpha and do universal REPL after solving an issue with passing the data between Python and Rust.

luketpeterson:
If the current repl is built with Python support, then the Python types are loaded, otherwise the Rust types take precedence.  It's definitely not a great solution, but it is enough to pass the tests for now.

vsbogd:
Following up the offline discussion about bridging Python/Rust/MeTTa types. Lets say we need to write a MeTTa stdlib `plus` function which gets a pair of integers and returns their sum as integer. As this is a grounded function we would like to implement it in Rust and reuse it for the Python grounded numbers. Which types do we need to support in stdlib and how can we call this function on Python integers?

Python represents numbers as big ints. In Rust platform dependent binary integers are used, and big int implemented by a separate library. It is typical to support both big ints and platform dependent integers as two separate types
because both implementations has own strong sides. We could use polymorphic functions to support both types as an arguments of the `plus` function similarly to the example below:
```
(: to_int (-> Atom int))
(= (to_int $x) (case $x ((a a))))

(: to_bigint (-> Atom BigInt))
(= (to_bigint $x) (case $x ((a A) (B B) (P B))))

(: plusi (-> int int int))
(= (plusi $a $b) (int sum))

[...]

PR #353, opened 2023-06-29
example of a space which query calls gpt-3.5

Necr0x0Der:


Necr0x0Der:
@astroseger , fyi

luketpeterson (approved):
This is very cool as a demonstration of the flexibility of MeTTa!

I don't have much to say about how you're using the API.  That looks good.  I do have some other questions / comments:

1. This comment is more about my own ignorance / lack of visibility into the MeTTa vision, but I am unclear why the GPT interface is implemented as a Space, rather than just inside a Grounded Atom (without the Space)?  I see the way the space's "container-like" property is useful for storing information.  And also how a language model can be conceptually thought of as encompassing a knowledge-base.  However this space couldn't be queried with arbitrarily complex expression atoms to match, which I thought was a conceptual requirement for the Space interface.

2. This is a great demonstration as it is.  But in my experience some more robustness in output parsing is needed if one wanted to productize this.  GPT can sometimes provide unexpected characters (e.g. punctuation, etc.)  However, if this is intended as a demo then parsing robustness code would get in the way so maybe it's best to leave it the way you have it.

3. I feel like the name "neurospace" might not be totally accurate.  Is it better to call it "lmspace" or "gptspace", given we might try and implement other NN algorithms within MeTTa in the future?

4. This is unrelted, but the repl (the one I wrote in Rust) doesn't support the `extend-py!` atom.  So that's something I need to figure out how to handle.

vsbogd:
> 4. This is unrelated, but the repl (the one I wrote in Rust) doesn't support the extend-py! atom. So that's something I need to figure out how to handle.

I believe here we could agree that Python and Rust can have specific parts of the stdlib which can be used only from Python REPL or Rust REPL respectively. At least until we have a good solution for https://github.com/trueagi-io/hyperon-experimental/issues/351

Necr0x0Der:
> 1. This comment is more about my own ignorance / lack of visibility into the MeTTa vision, but I am unclear why the GPT interface is implemented as a Space, rather than just inside a Grounded Atom (without the Space)?

Yeah, I also wondered what is the difference between atoms and spaces. Apparently, spaces respect certain API, and if we redefine `query` instead of `match`, it would be a space. But of course, it can be done via a grounded atom and corresponding grounded query function as well. But any space can be implemented as a grounded atom. So, why do we need Space API at all? This example is intended to show the use of Space API and it supposes further extensions or alternative implementations, e.g. storing additional embeddings of MeTTa expressions, or history of LLM prompts as a part of space extension, or whatever.

> 2. But in my experience some more robustness in output parsing is needed if one wanted to productize this. 

Of course. This is just an example in the sandbox. You should not expect it to be product-ready. If we will work on LLM-MeTTa integration framework, the code will be much more sophisticated.

> 3. I feel like the name "neurospace" might not be totally accurate.  Is it better to call it "lmspace" or "gptspace", given we might try and implement other NN algorithms within MeTTa in the future?

Yes, this example doesn't precisely correspond to "neurospace", and maybe it should have a different name if it would be a stand-alone demonstration. The reason for this name is that we discussed a possibility to create a neural space, which could be based on differentiable embeddings of MeTTa code with graph transformers or something else for queries. And this code is supposed as a maximally simplistic demonstration of using DNNs for querying a space (although without "neurospace" per se).

> 4. This is unrelted, but the repl (the one I wrote in Rust) doesn't support the `extend-py!` atom.  So that's something I need to figure out how to handle.

Well, as we (hopefully) discussed, there should be a common part of REPL implemented in Rust, and REPLs themselves implemented in Rust and Python. While REPL in Rust should not necessarily support `extend-py!`, REPL in Python should.

PR #355, opened 2023-07-06
Adapt CMake files to allow building Python packages under cibuildwheel environment

vsbogd:
In order to build Python package from source under [cibuildwheel](https://cibuildwheel.readthedocs.io/en/stable/) environment we need to split CMake project on two independent projects:
- `hyperonc` which compiles static C API library
- `hyperonpy` which compiles Python native module which depends on `hyperonc`

Python project `setup.py` is modified to copy pre-compiled `hyperonpy` module if it is present or compile it using `CMake` from `hyperonpy.cpp` if it is not found. Copying is used in local dev environment when user installs Python module in dev mode and then recompiles source code using `make. In order to keep compatibility with previous version of the CMake project one additional root CMake config is added to build and test first `hyperonc` and then Python `hyperon` module.

When `cibuildwheel` tool is used to compile the project it copies Python modules source files into the docker environment and initiates Python distribution package construction. To compile Python package successfully we need to prepare environment and install `hyperonc` module in it. It is done using `CIBW_BEFORE_ALL` (`before-all`) parameter and `install-hyperonc.sh` script which is called using this parameter. URL to the repository and revision to get `hyperonc` source code are passed to `install-hyperonc.sh` as arguments. By default the `main` branch of the [hyperon-experimental](https://github.com/trueagi-io/hyperon-experimental) is used. `install-hyperonc.sh` script installs Rust compiler, gets source code of the `hyperonc` library, compiles the library and installs it into the system.

Different tricks are needed to make `install-hyperonc.sh` work in different manylinux containers:
- `hyperonpy.cpp` has to be ported to `C++11` standard in order to be built under manylinux
- `perl-devel` has to be installed in `manylinux2014` environment
- building shared libraries should be disabled for `musllinux` environment as Rust in this environment doesn't support building shared libraries
- `CIBW_BEFORE_ALL` should be different for Linux and MacOSX because in MacOSX shell is not located on `/bin/sh` and in Linux script always called with `sh -c`
- Conan is instructed to build dependencies from source because otherwise they are incorrectly installed under musllinux environment

Even with these tricks `*-musllinux_i686` is not supported because there is no Rust compiler is available for this environment.

Last part of the PR is new GitHub Actions workflow to build and publish Python wheel packages as a part of GitHub release. Results can be found in my fork see [job running results](https://github.com/vsbogd/hyperon-experimental/actions/runs/5475859350) and [published artifacts](https://github.com/vsbogd/hyperon-experimental/releases/tag/v0.1.4.test-all). I have removed previous workflow because new one replaces it completely.

@wenwei-dev please take a look at the changes


vsbogd:
There are different future improvements we could try to make setup more straightforward:
- use CMake [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html#fetchcontent) instead of Conan to build project dependencies from source
- use CMake [ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html#module:ExternalProject) in `hyperonpy` CMake config to build `hyperonc` from source when project sources are available locally or fetch sources from GitHub when it is not available. Thus we could allow building Python module from source with minimal list of the steps to setup the environment. Serious problem here is keeping possibility of installing Python modules in editable mode (`pip install -e .`)

luketpeterson:
I experienced something strange.  I tried a clean build of this branch on MacOS, and the following happened when running the `make` step:

```
CMake Error at CMakeLists.txt:14 (include):
  include could not find requested file:

    /Users/admin/rustProjects/hyperon_upstream/hyperon-experimental/build/hyperonc-prefix/src/hyperonc-build/conanbuildinfo.cmake
```

However, I just re-ran it, changing nothing, and it was successful.

.github/workflows/release-python.yml
```diff
@@ -0,0 +1,64 @@
...
+          # before-all command is also called with "sh -c"
+          CIBW_BEFORE_ALL: ./python/install-hyperonc.sh -u https://github.com/${{github.repository}}.git -r ${{github.ref}}
+        with:
+          package-dir: ./python
+
+      - name: Build wheels on MacOSX
+        if: runner.os == 'macOS'
+        uses: pypa/cibuildwheel@v2.13.1
+        env:
+          # Mac OS X cannot call "./python/install-hyperonc.sh" probably because
```
luketpeterson, line 50: I don't think this is completely accurate.

Mac *has* a shell binary at `/bin/sh` (I'm pretty sure it's installed as part of the base OS, and not the xcode developer tools).  On Mac OS X 12.6, the version is: `GNU bash, version 3.2.57(1)-release`

However, the default shell on MacOS is `zsh 5.8.1` (again on Mac OS X 12.6.  so any environment setup and config likely won't be seen by `bash`

.github/workflows/release-python.yml
```diff
@@ -0,0 +1,64 @@
...
+          # "/bin/sh" is not correct path for the shell there
+          CIBW_BEFORE_ALL: sh -c ./python/install-hyperonc.sh -u https://github.com/${{github.repository}}.git -r ${{github.ref}}
```
luketpeterson, line 52: I can further confirm that `python/install-hyperonc.sh` succeeds on MacOSX.

luketpeterson (commented):
I don't feel I am qualified to approve these changes.  Knowledge of platform-specific best-practices RE install locations, etc. is needed.  For example. some package managers require certain directories to be synchronized with internal databases.  Stuff like that.

That said, I can confirm the changes work, and I can install hyperon and it works on Mac OS and on Ubuntu.

.github/workflows/release-python.yml
vsbogd: Well, truth is if I use "./python/install-hyperonc.sh" as a value of CIBW_BEFORE_ALL, build fails, please see [macosx build results](https://github.com/vsbogd/hyperon-experimental/actions/runs/5471670075/jobs/9963037785)

.github/workflows/release-python.yml
vsbogd: Yes, but it doesn't succeed in cibuildwheel macosx container see https://github.com/trueagi-io/hyperon-experimental/pull/355#discussion_r1254797609

luketpeterson: Looks like it's hitting that conan race condition / sequencing issue.  https://github.com/trueagi-io/hyperon-experimental/pull/355#issuecomment-1624092182

vsbogd:
>  I tried a clean build of this branch on MacOS, and the following happened when running the make step:

It is weird, there should not be any race conditions during the configuration stage I believe. At least usually building conan libraries takes a lot of time and I don't think `include` is executed in parallel with that.

luketpeterson:
> It is weird, there should not be any race conditions during the configuration stage

To be honest it was just a guess that it was due to asynchronous operation.  It did appear some processes continued even after the error was hit and then re-running it was successful.  But I don't have a smoking gun that is what's happening.  I'll work on figuring out a way to reproduce it that's faster than a fresh OS reinstall, and maybe that can point to the root cause.

luketpeterson:
**UPDATE**

The conan cache is definitely to part of the puzzle.  Running

```sh
conan remove '*'
```
causes the error to happen again.  Here is the failure message:

```
Installing (downloading, building) binaries...
cppunit/1.15.1: Configuring sources in /Users/admin/.conan/data/cppunit/1.15.1/_/_/source/src
cppunit/1.15.1: 
cppunit/1.15.1: Copying sources to build folder
cppunit/1.15.1: Building your package in /Users/admin/.conan/data/cppunit/1.15.1/_/_/build/3782cb8c3754bcc4f0ca2283b503d86479835458
cppunit/1.15.1: Generator txt created conanbuildinfo.txt
cppunit/1.15.1: Calling generate()
cppunit/1.15.1: WARN: Using the new toolchains and generators without specifying a build profile (e.g: -pr:b=default) is discouraged and might cause failures and unexpected behavior
cppunit/1.15.1: Aggregating env generators
cppunit/1.15.1: Calling build()
cppunit/1.15.1: Calling:
 > "/Users/admin/.conan/data/cppunit/1.15.1/_/_/build/3782cb8c3754bcc4f0ca2283b503d86479835458/src/configure" '--disable-shared' '--enable-static' '--prefix=/' '--bindir=${prefix}/bin' '--sbindir=${prefix}/bin' '--libdir=${prefix}/lib' '--includedir=${prefix}/include' '--oldincludedir=${prefix}/include' '--enable-debug=no' '--enable-doxygen=no' '--enable-dot=no' '--enable-werror=no' '--enable-html-docs=no' 
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... /Users/admin/.conan/data/cppunit/1.15.1/_/_/build/3782cb8c3754bcc4f0ca2283b503d86479835458/src/install-sh -c -d
checking for gawk... no
checking for mawk... no
checking for nawk... no
checking for awk... awk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking whether make supports nested variables... (cached) yes
checking whether make supports the include directive... yes (GNU style)
checking for gcc... /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc
checking whether the C compiler works... no
configure: error: in `/Users/admin/.conan/data/cppunit/1.15.1/_/_/build/3782cb8c3754bcc4f0ca2283b503d86479835458/build-release':
configure: error: C compiler cannot create executables
See `config.log' for more details
cppunit/1.15.1: 
cppunit/1.15.1: ERROR: Package '3782cb8c3754bcc4f0ca2283b503d86479835458' build failed
cppunit/1.15.1: WARN: Build folder /Users/admin/.conan/data/cppunit/1.15.1/_/_/build/3782cb8c3754bcc4f0ca2283b503d86479835458/build-release
ERROR: cppunit/1.15.1: Error in build() method, line 100
	autotools.configure()
	ConanException: Error 77 while executing "/Users/admin/.conan/data/cppunit/1.15.1/_/_/build/3782cb8c3754bcc4f0ca2283b503d86479835458/src/configure" '--disable-shared' '--enable-static' '--prefix=/' '--bindir=${prefix}/bin' '--sbindir=${prefix}/bin' '--libdir=${prefix}/lib' '--includedir=${prefix}/include' '--oldincludedir=${prefix}/include' '--enable-debug=no' '--enable-doxygen=no' '--enable-dot=no' '--enable-werror=no' '--enable-html-docs=no' 
CMake Error at CMakeLists.txt:14 (include):
  include could not find requested file:

    /Users/admin/rustProjects/hyperon_upstream/hyperon-experimental/build/hyperonc-prefix/src/hyperonc-build/conanbuildinfo.cmake


CMake Error at CMakeLists.txt:15 (conan_basic_setup):
  Unknown CMake command "conan_basic_setup".
```

But then whan I re-run `make`, it says:
```
Installing (downloading, building) binaries...
cppunit/1.15.1: WARN: Build folder is dirty, removing it: /Users/admin/.conan/data/cppunit/1.15.1/_/_/build/3782cb8c3754bcc4f0ca2283b503d86479835458
cppunit/1.15.1: Copying sources to build folder
cppunit/1.15.1: Building your package in /Users/admin/.conan/data/cppunit/1.15.1/_/_/build/3782cb8c3754bcc4f0ca2283b503d86479835458
cppunit/1.15.1: Generator txt created conanbuildinfo.txt
cppunit/1.15.1: Calling generate()
cppunit/1.15.1: WARN: Using the new toolchains and generators without specifying a build profile (e.g: -pr:b=default) is discouraged and might cause failures and unexpected behavior
cppunit/1.15.1: Aggregating env generators
cppunit/1.15.1: Calling build()
cppunit/1.15.1: Calling:
 > "/Users/admin/.conan/data/cppunit/1.15.1/_/_/build/3782cb8c3754bcc4f0ca2283b503d86479835458/src/configure" '--disable-shared' '--enable-static' '--prefix=/' '--bindir=${prefix}/bin' '--sbindir=${prefix}/bin' '--libdir=${prefix}/lib' '--includedir=${prefix}/include' '--oldincludedir=${prefix}/include' '--enable-debug=no' '--enable-doxygen=no' '--enable-dot=no' '--enable-werror=no' '--enable-html-docs=no' 
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... /Users/admin/.conan/data/cppunit/1.15.1/_/_/build/3782cb8c3754bcc4f0ca2283b503d86479835458/src/install-sh -c -d
checking for gawk... no
checking for mawk... no
checking for nawk... no
checking for awk... awk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking whether make supports nested variables... (cached) yes
checking whether make supports the include directive... yes (GNU style)
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
...
```

luketpeterson:
**Update**  Looks very possible that we're hitting this: https://github.com/conan-io/cmake-conan/issues/470

luketpeterson:
I am probably not the best person to debug the failure on Mac, as I know very little about Conan.  Although I'm willing to try.

That being said, I think it's a separate issue from this PR, as the problem reproduces with the current main branch.

vsbogd:
> Looks very possible that we're hitting this: https://github.com/conan-io/cmake-conan/issues/470

@luketpeterson , thank you for finding https://github.com/conan-io/cmake-conan/issues/470 and commenting in it. I agree it doesn't look like the issue was introduced by this PR. I think one of the ways to fix it is to use Conan 2.0 as issue suggests or to get rid of Conan at all and use `FetchContent` instead.

vsbogd:
Btw @luketpeterson, could you please try `conan profile new --detect default` after clearing cache but before build. Does it help?

luketpeterson:
> Btw @luketpeterson, could you please try `conan profile new --detect default` after clearing cache but before build. Does it help?

I never changed / deleted my profile, so even after clearing the cache, it says `ERROR: Profile already exists`

vsbogd:
Merging it to not block C API changes in #361

PR #357, opened 2023-07-07
Fix cdr-atom behavior

### LINKED ISSUE #356: cdr-atom does not handle nullary or unary tuple correctly
## What is the problem?

`cdr-atom` does not handle nullary, `()`, or unary such as `(A)` expression correctly.

## How to reproduce the problem?

Run the following MeTTa codes

```
!(cdr-atom ())
```

or

```
!(cdr-atom (A))
```

## What should normally be output?

`!(cdr-atom ())` should output

```
[(Error (cdr-atom ()) cdr-atom expects non-empty expression)]
```

like `car-atom` does.

And `!(cdr-atom (A))` should output

```
[()]
```

## What is output instead?

They both crash metta.

vsbogd:
Fix #356

### DIFF
--- a/lib/src/metta/runner/mod.rs
+++ b/lib/src/metta/runner/mod.rs
@@ -115,10 +115,10 @@ impl Metta {
     pub fn run(&self, parser: &mut SExprParser) -> Result<Vec<Vec<Atom>>, String> {
         let mut mode = Mode::ADD;
         let mut results: Vec<Vec<Atom>> = Vec::new();
-    
+
         loop {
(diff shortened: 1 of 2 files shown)


vsbogd:
@luketpeterson this commit https://github.com/trueagi-io/hyperon-experimental/pull/357/commits/3b70da12bd0c1fd83eb6443f5f605926afac2c26 fixes incorrect `MettaRunner` behavior and thus affects REPL behaviour.

Necr0x0Der (approved):

PR #364, opened 2023-07-15
Change Grounded::execute function signature so args is an immutable slice, vs. mutable Vec

luketpeterson:
This gets us a tiny bit closer to being able to implement concurrency.  But the immediate reason I did it is so I could unify the C vec_atom_t type with the atom_array_t type.


lib/src/metta/interpreter.rs
```diff
@@ -529,8 +529,8 @@ fn execute_op<'a, T: SpaceRef<'a>>(context: InterpreterContextRef<'a, T>, input:
             let mut expr = expr.clone();
             let op = expr.children().get(0).cloned();
             if let Some(Atom::Grounded(op)) = op {
-                let mut args = expr.children_mut().drain(1..).collect();
-                match op.execute(&mut args) {
+                let args: Vec<Atom> = expr.children_mut().drain(1..).collect();
+                match op.execute(&args[..]) {
```
vsbogd, line 533: In this PR we can eliminate copying in the code above:
```rust
            let op = expr.children().get(0);
            if let Some(Atom::Grounded(op)) = op {
                let args = expr.children();
                match op.execute(&args[1..]) {
```

vsbogd (approved):
Looks good to me, but see comment on implementation of the interpreter.

lib/src/metta/interpreter.rs
luketpeterson: I adopted your suggestion in 4cc22d6.

Issue #348, opened 2023-06-28
Two different atom array types in C API


luketpeterson:
The C API has two different atom array types: `atom_array_t` (used by atom callbacks) and `vec_atom_t` used by the grounded atom, space, and bindings APIs.

I believe there should only be one data structure in the API to represent an array of atoms.

I have a slight preference for vec_atom_t, but I'd be happy with either one - but I don't see a value in having multiple.

vsbogd:
They have slightly different purposes. Historically `vec_atom_t` was introduced as a C API wrapper of the Rust dynamic array `Vec<Atom>` and `array_t` (and `atom_array_t` as its specialization)  was introduced to be transparent representation of a constantly sized array for C API clients (basically it is a pointer to array and size). In case of `*const vec_atom_t` they should be absolutely replaceable. 

I agree that there is some duplication. And I don't have a strong opinion which implementation should be kept. In fact as dynamic array case covers static array case and to implement dynamic array case we anyway need some list of interface  functions, probably we could introduce array type with such functions (or keep `vec_atom_t`) and use any implementation which works.

luketpeterson:
Let's change the interface over to `vec_atom_t`, and as you said, pass `*const vec_atom_t` for the callback functions.

I'll do this if you agree.

vsbogd:
Yes, I think its ok to do this change.

PR #367, opened 2023-07-18
Converge c vec types

luketpeterson:
This PR addresses https://github.com/trueagi-io/hyperon-experimental/issues/348

My aim was to provide a single "best of both worlds" atom array type, because there was no reason for there to be two.  It's freely interoperable with Rust and has performance characteristics in line with the Rust Vec<Atom>.

The only thing that's temporarily sacrificed is the ability to read atoms directly by indexing through the pointer, because the internal types are `Atom`, and not `atom_ref_t`.  This ability will return at a future time when `atom_t` is a mirror of `Atom`.

Finally, I renamed `vec_atom_t` to `atom_vec_t`.  I did this because I kept remembering it as `atom_vec`, and then having to remind myself it was `vec_atom` so I thought `atom_vec` is more natural.  However, It is possible I feel this way because I'm an English speaker, and a speaker of a Romance language / or another noun->adjective language might take issue with this change.  So I am open to revert the final commit of this PR where I implemented the name change.

Thanks.


c/src/atom.rs
```diff
@@ -709,14 +806,8 @@ pub fn ptr_into_bindings(bindings: *mut bindings_t) -> Bindings {
...
 
-fn vec_atom_into_ptr(vec: Vec<Atom>) -> *mut vec_atom_t {
-    Box::into_raw(Box::new(vec_atom_t(vec)))
-}
-
 pub fn return_atoms(atoms: &Vec<Atom>, callback: c_atoms_callback_t, context: *mut c_void) {
-    let results: Vec<atom_ref_t> = atoms.iter()
-        .map(|atom| atom.into()).collect();
-    callback((&results).into(), context);
+    callback(&(&atoms[..]).into(), context);
```
vsbogd, line 810: Except `atom_get_children` case `return_atoms` is used to return owned instance of the `Vec<Atom>` maybe it makes sense split this function on two: `return_owned_atoms`, `return_borrowed_atoms`? Then in turn we should introduce to versions of `c_atoms_callback_t` and modify `hyperonpy.cpp` to clone atoms when borrowed instance is returned and move them when owned instance is returned.

c/src/atom.rs
```diff
@@ -644,49 +649,141 @@ pub unsafe extern "C" fn atom_eq(atoma: *const atom_ref_t, atomb: *const atom_re
...
-    let atom = &(*vec).0[idx];
+pub unsafe extern "C" fn atom_vec_get(vec: *const atom_vec_t, idx: usize) -> atom_ref_t {
+    let vec = &*vec;
+    let atom: &Atom = &vec.as_slice()[idx];
     atom.into()
 }
 
-pub type atom_array_t = array_t<atom_ref_t>;
-pub type c_atoms_callback_t = lambda_t<atom_array_t>;
+pub type c_atoms_callback_t = lambda_t<*const atom_vec_t>;
```
vsbogd, line 786: I would suggest to rename it into `c_atom_vec_callback_t`.

c/src/atom.rs
```diff
@@ -644,49 +649,141 @@ pub unsafe extern "C" fn atom_eq(atoma: *const atom_ref_t, atomb: *const atom_re
...
 
 #[no_mangle]
-pub unsafe extern "C" fn vec_atom_free(vec: *mut vec_atom_t) {
-    drop(Box::from_raw(vec));
+pub unsafe extern "C" fn atom_vec_from_array(atoms: *mut atom_t, size: usize) -> atom_vec_t {
+    let c_arr = std::slice::from_raw_parts_mut(atoms, size);
+    let atoms: Vec<Atom> = c_arr.into_iter().map(|atom| {
+        core::mem::replace(atom, atom_t::null()).into_inner()
+    }).collect();
+    atoms.into()
```
vsbogd, line 741: Just curious, is there a difference with the following implementation:
```rust
pub unsafe extern "C" fn atom_vec_from_array(atoms: *mut atom_t, size: usize) -> atom_vec_t {
    let atoms: Vec<atom_t> = Vec::from_raw_parts(atoms, size, size);
    let atoms: Vec<Atom> = atoms.into_iter().map(|atom| atom.into_inner()).collect();
    atoms.into()
}
```

vsbogd (approved):

c/src/atom.rs
luketpeterson: I don't understand.  The implementation of `atom_get_children` provides a borrowed reference to the atoms, mirroring the Rust interface.  Then when `atom_get_children` is called in hyperonpy.cpp, the returned atoms are already cloned in the `copy_atoms` callback function.

Are you saying you would like to see an additional interface that consumes an expression atom, and returns an owned vector of its children?

c/src/atom.rs
luketpeterson: The issue is that `Vec::from_raw_parts` has a number of safety considerations that we can't guarantee if we're taking the pointer from C.  Namely which allocator was used, the allocation size (it's not enough that the allocation is big enough - it has to be exact), and the pointer alignment.

So basically, the only time `Vec::from_raw_parts` is safe is if the vec started life as a Rust Vec, and we're just letting C hold onto the pointer.

`slice::from_raw_parts` is a lot more forgiving about the pointer details because it doesn't take responsibility for managing the pointer later on. (freeing / growing / etc.)

c/src/atom.rs
luketpeterson: Name changed in 2059746

luketpeterson:
Merged this PR, but opened https://github.com/trueagi-io/hyperon-experimental/pull/368 for the minor points in the code review.

c/src/atom.rs
vsbogd: > Are you saying you would like to see an additional interface that consumes an expression atom, and returns an owned vector of its children?

Sorry I mentioned `atom_get_children` and it is probably confused you. I mean that all usages of `return_atom` (there are about 5 places as far as I remember) except `atom_get_children` actually could pass ownership of the `Vec<Atom>` to the `return...` function. Thus we could introduce another version of the `return_atoms` which gets ownership and use this version everywhere except `atom_get_children`.

c/src/atom.rs
vsbogd: Ah ok, you are right, missed it in `Vec` documentation. Thanks!

PR #370, opened 2023-07-19
Fix Python searching procedure for CMake less than 3.18

vsbogd:
Select correct list of modules to search for CMake versions less than 3.18. `Development.Module` is supported from 3.18, on the other hand `Development` cannot be used for versions 3.18 and greater because it implies `Development.Embed` which in turn is absent in `cibuildwheel` environment.


vsbogd:
@bessveta FYI

luketpeterson (approved):
LGTM

Issue #371, opened 2023-07-19
Fails to build C API after recent updates


Necr0x0Der:
Error under MacOS:

 Building C object tests/CMakeFiles/check_atom.dir/check_atom.c.o
In file included from hyperon-experimental/c/tests/check_atom.c:1:
hyperon-experimental/build/hyperonc-prefix/src/hyperonc-build/cargo/hyperon/hyperon.h:71:33: error: field has incomplete type 'struct Box_RustOpaqueAtom'
      struct Box_RustOpaqueAtom owned;
                                ^
hyperon-experimental/build/hyperonc-prefix/src/hyperonc-build/cargo/hyperon/hyperon.h:32:16: note: forward declaration of 'struct Box_RustOpaqueAtom'
typedef struct Box_RustOpaqueAtom Box_RustOpaqueAtom;
               ^
1 error generated.

@luketpeterson , do you have any idea of the reason and how it can be fixed?

luketpeterson:
Can you please share the hyperon.h file that was generated?  

cbindgen may rely on some linker hooks to decide which symbols to export, and it is possible the Arm version of the Rust toolchain is behaving differently from the x86 version.

Necr0x0Der:
JIC:
arch_build=armv8
build_type=Release
compiler=apple-clang
compiler.libcxx=libc++
compiler.version=14
os_build=Macos


```
#ifndef HYPERON_H
#define HYPERON_H

/* Warning, this file is autogenerated by cbindgen. Don't modify this manually. */

#include <stdarg.h>
#include <stdbool.h>
#include <stdint.h>
#include <stdlib.h>
struct gnd_t;

typedef enum atom_type_t {
  SYMBOL,
  VARIABLE,
  EXPR,
  GROUNDED,
} atom_type_t;

typedef enum space_event_field_t {
  SPACE_EVENT_FIELD_ADD,
  SPACE_EVENT_FIELD_REMOVE,
  SPACE_EVENT_FIELD_REPLACE_PATTERN,
  SPACE_EVENT_FIELD_REPLACE_TEMPLATE,
} space_event_field_t;

typedef enum space_event_type_t {
  SPACE_EVENT_TYPE_ADD,
  SPACE_EVENT_TYPE_REMOVE,
  SPACE_EVENT_TYPE_REPLACE,
} space_event_type_t;

typedef struct Box_RustOpaqueAtom Box_RustOpaqueAtom;

[...]

PR #372, opened 2023-07-19
Changing atom_t internals from Box to *mut

luketpeterson:
Changing atom_t internals from Box to *mut because cbindgen exports them differently on Arm.

https://github.com/trueagi-io/hyperon-experimental/issues/371


Necr0x0Der (approved):
Works for me. Thanks!

PR #377, opened 2023-07-25
Quickfix: propagate release build configuration into external projects

vsbogd:
It turned out that cmake `ExternalProject` functions doesn't propagate the cmake configuration to the external projects. Thus using `cmake -DCMAKE_BUILD_TYPE=Release` doesn't build `hyperonc` in release mode. This is a quickfix to propagate release configuration into the projects included.


vsbogd:
@bessveta FYI

luketpeterson (approved):
These changes look fine, and I have confirmed they don't make anything worse on my setup.

It wasn't clear if the thinking was that this might address the conan problem on Mac (make fails the first time after a conan cache clear), because that still occurs for me even with these changes.

luketpeterson:
> It wasn't clear if the thinking was that this might address the conan problem on Mac (make fails the first time after a conan cache clear), because that still occurs for me even with these changes.

Raised https://github.com/trueagi-io/hyperon-experimental/issues/379 to track the Conan issue

PR #389, opened 2023-08-07
C api docs2

luketpeterson:
Some more progress towards finalizing the C API docs, and cleaning up the Bindings interface a little bit more.


c/src/atom.rs
```diff
@@ -1006,42 +1010,81 @@ impl bindings_set_t {
...
 }
 
-pub type bindings_mut_callback_t = lambda_t<*mut bindings_t>;
+/// @brief Function signature for a callback providing mutable access to individual Bindings frames within a Bindings Set
+/// @ingroup matching_group
+/// @param[in]  bindings  A pointer to the `bindings_t`.  It is ok to call functions that modify the `bindings_t` within the callback
+/// @param[in]  context  The context state pointer initially passed to the upstream function initiating the callback.
+/// @warning The ability for the function to mutate the `bindings_t` may be deprecated in the future, so please use this ability cautiously.
+///
+//TODO: QUESTION FOR VITALY: Is there a good reason to keep this mutable?  I think we would be better off with an immutable accessor, as we update the APIs for the non-determinism
```
vsbogd, line 1019: I believe I made it mutable because in my version of the API it was required when called function gets ownership over passed argument. Thus `* mut` can be replaced by moving the passed instance.

c/src/atom.rs
```diff
@@ -1080,14 +1130,25 @@ pub extern "C" fn bindings_add_var_binding(bindings: *mut bindings_t, var: atom_
...
-/// Returns the atom bound to the supplied variable in the bindings.  Returns a NULL atom_ref if the
-/// variable is not present.
+/// @brief Returns the atom bound to the supplied variable name in the `bindings_t`
+/// @ingroup matching_group
+/// @param[in]  bindings  A pointer to the `bindings_t` to inspect
+/// @param[in]  var_name  A NULL-terminated C-style string containing the name of the variable
+/// @return The `atom_t` representing the atom that corresponds to the specified variable, or a NULL `atom_ref_t` if the variable is not present.
+/// @note The caller must take ownership responsibility for the returned `atom_t`, if it is not NULL
+///
+//TODO: Question for Vitaly: Would be be better off here returnin an actual atom ref, so the caller can clone it if needed?
```
vsbogd, line 1151: It is not always possible to return the reference to the atom from Rust because sometimes atoms are constructed on the fly. One major case is an atom which is constructed by resolving its sub-variables via same instance of `Bindings`.

vsbogd (approved):
Looks good to me.

luketpeterson:
I'm thinking I'll go ahead and merge now, but I have more documentation to add.

I'll finish the docs up and remove the extraneous comments in the next PR.

c/src/atom.rs
vsbogd: Yes, I think it should work for this case.

PR #391, opened 2023-08-09
Comment test scripts

glicerico:
While going through the test scripts, I added comments to help me understand the code more easily.
If you think they may be useful to others, feel free to merge.
A lot of the changes are just decorations that may help readability, which I can remove if preferred.

In the case that this PR is a candidate to merge, it would be a good idea to check the comments in all files for correctness, and in particular files b1 and c3 (I reordered the example bc it makes more sense to me this way, but may be incorrect).


Necr0x0Der:
I wonder what is the test time with and without these additional comments.

glicerico:
@Necr0x0Der Oh, I didn't know comments would slow down execution significantly.
From what I see, these scripts are not included in `make check`, but  I tested running all scripts in the `scripts` folder 10 times, and I see no time difference at all between the 2 branches.

Necr0x0Der:
I hope they shouldn't slow down execution (parsing), but I personally hesitated to add a lot of comments in unit/integration tests, which are executed very often, in order to not increase test time. This was just a good moment to check that they really don't introduce additional expenses. In fact, they are executed here https://github.com/trueagi-io/hyperon-experimental/blob/20a18a04db82e03080b45a2316bac3b85481b663/python/tests/test_run_metta.py#L77
Thus, could you please compare this time for two branches?
12/14 Test #12: test_run_metta.py ................   Passed    0.78 sec

glicerico:
Got it... I ran them with both branches 10 times each, and in average, my branch takes 0.01s longer  to run the script tests.

python/tests/scripts/a1_symbols.metta
```diff
@@ -1,11 +1,14 @@
...
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; `match` searches for all expressions corresponding to
+; the given pattern and produces the output pattern.
+; It doesn't search in subexpressions.
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ; Some expressions to be matched
 ((leaf1 leaf2) leaf3)
 (((leaf0 leaf1) leaf2) leaf3)
-; This one will not be matched
+; This one will not be matched because it's a subexpression
```
Necr0x0Der, line 10: "It's" might be not clear here. Maybe, we need to be explicit here, e.g. "This one contains ((leaf1 leaf2) leaf3) as a subexpression and thus will not be matched"

python/tests/scripts/a3_twoside.metta
```diff
@@ -16,11 +22,16 @@
     ($x might be $P))
   ((Sam might be Frog)))
 
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; Composite queries
+; The comma is used to form multiple queries with shared
```
Necr0x0Der, line 28: Maybe, instead of just "multiple" something more specific could be used, e.g. "conjuncted queries" or "conjunction of queries"?

python/tests/scripts/b0_chaining_prelim.metta
```diff
@@ -1,35 +1,42 @@
-; From matching to rule chaining
-
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; From matching to rule chaining:
```
Necr0x0Der, line 3: Can we replace "rule" with something else here? I'm ok with it, but it sounds not perfect, because these are not necessarily rules. We are talking about either equalities, which matching is chained by the interpreter, or or just arbitrary pattern matching queries. Maybe, we can more verbose here, e.g. "From matching to evaluation of expressions via equality queries chaining" or something like this.

python/tests/scripts/b0_chaining_prelim.metta
```diff
@@ -1,35 +1,42 @@
...
-; First step
+; A single matching step
 !(assertEqualToResult
   (match &self
     (:= (S K K x) $r)
     $r)
   ((K x (K x))))
 
-; Second step
+; Another single matching step
```
Necr0x0Der, line 21: It's not just "another single matching step". It should be underlined that this matching step takes the result of the previous matching step as input. "First" and "Second" provided an order for steps although were not that explicit regarding inputs/outputs. Now, the focus is on the fact that steps are single, but it is less important than the fact that these steps are manually chained.

python/tests/scripts/b1_equal_chain.metta
```diff
@@ -1,43 +1,66 @@
...
 
 !(assertEqual
   (Add (S Z) (S Z))
   (S (S Z)))
 
-; In the current implementation, the child expression will be reduced
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; In the current implementation, child expressions
+; are also reduced
```
Necr0x0Der, line 51: Why "also"?

python/tests/scripts/b1_equal_chain.metta
```diff
@@ -1,43 +1,66 @@
...
   (Something? (Add (S Z) (S Z)))
   (Something? (S (S Z))))
 
-; Note that the same variable can be used multiple time on the left
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; Note that the same variable can be used multiple times on the left
 ; side of expression, and the "function" can be "non-total"
+; (a "non-total" function can be defined only for some inputs;
+; is not required to cover all possible inputs)
```
Necr0x0Der, line 62: These two lines are redundant saying similar things. Moreover, if the function is not total, it really is not defined for all inputs (not just "can" or "not required"). If it is defined for all inputs, it is total.

python/tests/scripts/b2_backchain.metta
```diff
@@ -15,53 +18,85 @@
...
 (= (ift T $then) $then)
 (ift (deduce (Evaluation (mortal $x))) $x)
 
-
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; Another example of backchaining
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; Definition of explain when the desired Evaluation is
+; in  the knowledge base
```
Necr0x0Der, line 80: Excessive whitespace?

python/tests/scripts/b4_nondeterm.metta
```diff
@@ -1,90 +1,122 @@
...
 ; for some combinations of their values, the whole expression
 ; for these values will appear in the set of final results,
 ; but without reduction. However, `match` can be called
 ; directly to filter out non-reducible results
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 (= (foo) red)
 (= (foo) boo)
 (= (eq $x $x) T)
 (= (find-equal $x $y)
-   (match &self (= (eq $x $y) T) $x))
```
Necr0x0Der, line 61: Personally, I prefer to keep this whitespace

python/tests/scripts/c1_grounded_basic.metta
```diff
@@ -100,21 +127,36 @@
...
+
+; Calculates the sum of element-wise products between two lists
 (= (subsum nil nil) 0)
 (= (subsum (:: $x $xs) (:: $b $bs))
    (+ (* $x $b)
       (subsum $xs $bs)))
 
+; Non-determinism "reasoning":
+; Among all 3-bit binary lists, return the one whose `subsum`
+; with (:: 3 (:: 7 (:: 5 nil))) equals 8, or `nop` if not found
```
Necr0x0Der, line 159: JIC, nop is a grounded function that consumes its arguments and returns nothing.

python/tests/scripts/c2_spaces.metta
```diff
@@ -11,30 +13,41 @@
        ($obj is-a Planet))
     (Color of Planet $obj is $prop))
   ((Color of Planet Mars is Red)))
-; These are not in `&self`
+
+; The knowledge in `&kb` is not `&self`
```
Necr0x0Der, line 17: Not sure I got this sentence. Maybe "in" is missed: "is not in `&self`"?

python/tests/scripts/b5_types_prelim.metta
```diff
@@ -13,97 +16,148 @@
...
   (Add S Z)
   ((Error S BadType)))
-; This will still work, because `Something` is of undefined type,
-; and it might be `Nat`
-!(assertEqual
-  (Add Something Z)
-  Something)
-; This should work normally
+; But these other examples works normally
+!(assertEqualToResult
```
Necr0x0Der, line 33: Why is it changed to `assertEqualToResult`?

python/tests/scripts/b5_types_prelim.metta
```diff
@@ -13,97 +16,148 @@
...
+; The following will be accepted by the interpreter
 !(assertEqualToResult
    (Add Z Ten)
   ((Add Z Ten)))
 
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ; Types can be parameterized
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; List is defined as a dependent function type that
```
Necr0x0Der, line 64: List doesn't have a dependent function type. It has just a function type.

python/tests/scripts/b5_types_prelim.metta
```diff
@@ -13,97 +16,148 @@
...
+; takes a Type and returns a Type
```
Necr0x0Der, line 65: This description sounds vague. List doesn't take `Type` as a symbol. Also, it is misleading to say that it returns a Type. Technically, it doesn't "return" anything.

python/tests/scripts/c3_pln_stv.metta
```diff
@@ -1,23 +1,33 @@
...
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; Some helpful definitions
 (= (min $a $b) (if (< $a $b) $a $b))
 (= (s-tv (stv $s $c)) $s)
 (= (c-tv (stv $s $c)) $c)
 
-; .tv serves as a special type of equality that requires
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ; `TV` "metarule" based on `match` to run
```
Necr0x0Der, line 13: Why description of .tv is dropped? I believe we need to add something about it, e.g., that it is a custom symbol, which is used to connect PLN-like expressions with their truth values. I'm not sure if we should go into detailed comparison with OpenCog Classic (e.g., that in this example .tv-expressions are ordinary expressions unlike OpenCog Classic, in which they are special type of values), or not. But some general comments on semantics of such custom symbols would be useful.

python/tests/scripts/c3_pln_stv.metta
```diff
@@ -27,24 +37,36 @@
    )
 )
 
-
-; Reproducing examples/pln/conjunction (extended with implication)
```
Necr0x0Der, line 31: Why is concrete reference to examples/pln/conjunction is dropped?

python/tests/scripts/d2_higherfunc.metta
```diff
@@ -87,30 +117,44 @@
...
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ; We can implement a generic `fmap`, but it requires
-; concrete patterns in type constructors
+; concrete patterns in the type constructors above
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; Instead, we implement `fmap` as:
 (: fmap (-> (-> $a $b) ($F $a) ($F $b)))
-(= (fmap $f ($C0)) ($C0))
-(= (fmap $f ($C $x)) ($C ($f $x)))
+; Mapping over empty functor, returns empty functor
```
Necr0x0Der, line 127: Not sure if "empty functor" is a standard notion, but let it be...

python/tests/scripts/d3_deptypes.metta
```diff
@@ -1,55 +1,75 @@
...
 ; A classical example of dependent types
-
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; Simple type definitions:
+; Defines the natural numbers `Nat` for counting
 (: Nat Type)
 (: Z Nat)
 (: S (-> Nat Nat))
 
+; `Vec` is defined as a dependent type: it takes a type ``$t` and a `Nat` length.
```
Necr0x0Der, line 11: Excessive `. Also, "it takes a `$t` type and a `Nat` length" sounds more uniform.

python/tests/scripts/d5_auto_types.metta
```diff
@@ -1,49 +1,61 @@
...
 !(assertEqualToResult
    (collapse (+ 5 "S"))
    (((Error "S" BadType))))
 
-; `match` can accept badly typed expressions without type error,
-; which is correct
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; match does not perform type checking on its pattern,
+; so it accepts a badly typed expression, which is expected
```
Necr0x0Der, line 47: We can add that this allows us to write code that analysis badly typed expressions in MeTTa itself (e.g. for self-reflective genetic programming, or whatever). Basically, it might be useful to underline that there are cases when we do want to deal with "wrong" MeTTa programs on a meta-level in MeTTa itself.

python/tests/scripts/e1_kb_write.metta
```diff
@@ -1,5 +1,12 @@
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; MeTTa allows to use a separate knowledge base (AtomSpace) &kb
```
Necr0x0Der, line 3: We don't use "AtomSpace" in MeTTa implementation. Maybe, it is better to call it Space here, although calling it AtomSpace in some other contexts is perfectly ok.

python/tests/scripts/e1_kb_write.metta
```diff
@@ -1,5 +1,12 @@
...
+; to accumulate inferred knowledge
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; Bind &kb to a new empty AtomSpace
```
Necr0x0Der, line 6: Space

python/tests/scripts/e2_states.metta
```diff
@@ -1,25 +1,35 @@
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; State atoms
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; `new-state` creates a new state atom wrapping the initial value (A B).
+; Then it binds this state atom to the variable `&state-token`.
```
Necr0x0Der, line 6: It is not a variable. It is actually a token, which is replaced by the corresponding atom at parse-time.

python/tests/scripts/e2_states.metta
```diff
@@ -1,25 +1,35 @@
...
 !(bind! &state-token (new-state (A B)))
 
-; &state-token is replace by the state atom in the code below its creation
+; &state-token is replaced by the state atom in the code below its creation
 ; we wrap it into a function to show that it's not the token that changes, but
 ; the content of the state atom
 (= (get-token) &state-token)
 
-; Expected content of the state atom
+; Although `new-state $x` creates a `State($x)` structure,
```
Necr0x0Der, line 14: Why "although"? Also, `(State $x)` - not `State($x)`.

Necr0x0Der:
@glicerico , thanks for improving comments. I reviewed them and added minor suggestions. Some of them might be ignored, but I would appreciate if you take all of them into account. Besides minor suggestions, I found descriptions in b5_types_prelim.metta confusing or even incorrect in general. Maybe, I need to try adding descriptions by myself in order to deliver the semantics of type definitions in MeTTa in a right way (which is not too standard). Thus, there is no need to consider this file from your side. We can either merge it as is and I will work on it later, or you can send the current version to me and roll back your changes.

python/tests/scripts/b0_chaining_prelim.metta
glicerico: Oh, I missed that... but actually I don't understand how the result of the first step is used by the second step: if I comment out the "first step", the second one still works, because it is matched with the 3rd knowledge base entry:
```(:= (K $x $y) $x)```

File: python/tests/scripts/b1_equal_chain.metta (hunk 6 above)
Comment (glicerico) [reply python/tests/scripts/b1_equal_chain.metta:62]: I meant the text in parenthesis as an explanation of what is a "total-function". Mostly because I didn't know what it means, but I changed wording to fix your suggestion:

File: python/tests/scripts/b5_types_prelim.metta (hunk 11 above)
Comment (glicerico) [reply python/tests/scripts/b5_types_prelim.metta:33]: I moved the example with "AssertEqual" below, and added a new example here

File: python/tests/scripts/b5_types_prelim.metta (hunk 13 above)
Comment (glicerico) [reply python/tests/scripts/b5_types_prelim.metta:65]: I guess I don't understand this completely... is this more accurate?
```
; List is defined as a function type with
; a typed argument
(: List (-> Type Type))
```

File: python/tests/scripts/c3_pln_stv.metta (hunk 15 above)
Comment (glicerico) [reply python/tests/scripts/c3_pln_stv.metta:31]: my bad, restored this info in the file header

File: python/tests/scripts/c3_pln_stv.metta (hunk 14 above)
Comment (glicerico) [reply python/tests/scripts/c3_pln_stv.metta:13]: I had moved the ".tv" description below, but I added a comment here to avoid confusion

File: python/tests/scripts/d5_auto_types.metta (hunk 18 above)
Comment (glicerico) [reply python/tests/scripts/d5_auto_types.metta:47]: new proposal:
```
; `match` does not perform type checking on its pattern,
; so it accepts a badly typed expression, which is expected.
; It can be useful to deal with "wrong" MeTTa programs on
; a meta-level in MeTTa itself, so this behavior of `match`
; allows us to write code that analyzes badly typed expressions
; within MeTTa (e.g. for self-reflective genetic programming).
```

File: python/tests/scripts/b2_backchain.metta (hunk 7 above)
Comment (Necr0x0Der) [reply python/tests/scripts/b2_backchain.metta:77]: Here "in  the knowledge base" - two spaces between "in" and "the".

File: python/tests/scripts/b5_types_prelim.metta (hunk 11 above)
Comment (Necr0x0Der) [reply python/tests/scripts/b5_types_prelim.metta:33]: `AssertEqualToResult` doesn't reduce the second argument, so it is used, when we want to make sure that the result is not evaluated further. For example, it is senseless to write `(AssertEqual (f x) (f x))`, because both `(f x)` will or will not be reduced. However, if we use `AssertEqualToResult` in this case, we show that `(f x)` is not reduced. Another example is when evaluation produces a error. In your case, the most important part to check is that `Add` works, and `(Add (S Z) Z)` is equivalent to `(S Z)`. Thus, using `AsserEqual` is perfectly ok. It is not a mistake to use `AssertEqualToResult`, but it is just not necessary.

File: python/tests/scripts/e2_states.metta [hunk 23]
```diff
@@ -36,33 +47,49 @@
...
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; scoping doesn't conflict between let bound variables and state atom types.
 !(assertEqual
   (let $x (new-state 1)
      (change-state! $x (+ (get-state $x) 1)))
   (new-state 2))
 
-; Chcking that there is no conflict between variable names in let-expression
-; and type definition of new-state
+; Checking that state atoms don't conflict with surrounding scopes;
```
Necr0x0Der, line 88: The problem is `new-state` has (or had?) internal type definition using `$tnso` variable name. New explanation doesn't say this, so it is not clear what it really means.

python/tests/scripts/b0_chaining_prelim.metta
Necr0x0Der: I mean that the first match produces `(K x (K x))`. This result is "manually" wrapped into the query - `(:= (K x (K x)) $r)`. This query is executed as the second "step" to produce the next result. You can explain explicitly that we take the result of the first query here and construct the next query, which performs the second step of reduction, if it is not clear enough from the code.

Necr0x0Der: And of course, the second step works if you comment out the first step, because the result of the first step is already in the second query.

python/tests/scripts/b1_equal_chain.metta
Necr0x0Der: I'm not sure if this terminology is precise, but it can be improved later if needed.

python/tests/scripts/b5_types_prelim.metta
Necr0x0Der: To be accurate w.r.t. MeTTa semantics, I'd write something like "List has a function type. Its argument should be of Type type, and the type of the whole expression, e.g. (List String), will also have Type type." Does it sound too verbose or unclear?

python/tests/scripts/b0_chaining_prelim.metta
glicerico: Got it, thx

python/tests/scripts/b5_types_prelim.metta
glicerico: Good piece of information about `AssertEqualToResult` and `AssertEqual`, which I didn't internalize before.

The explanation for the former is in the `a1_sumbols.metta` file, but the one for `AssertEqual` comes until file `b1_equal_chain.meeta`. It could be useful to have both definitions together for contrast, perhaps in `a1`, what do you think?

python/tests/scripts/e2_states.metta
glicerico: Okay, restoring your original description.

python/tests/scripts/b5_types_prelim.metta
Necr0x0Der: It could be useful, but it can be done in a separate PR.

Necr0x0Der (approved):

PR #392, opened 2023-08-09
added sealed operator

cowboyphilosopher:
added new operator sealed. 
format: (sealed (Expression with sealed variable list) (Term with variables to seal))
example: (sealed ($x $y) (= ($x $y) ($a $b))
$x and $y should be replaced with new unique variables in the Term value


vsbogd (commented):
Is it correct that `sealed` operator just should make variables unique? Or it is just first step of implementing it? I am asking because in `hyperon-experimental` usually makes variables unique when atom is extracted from the space or for example returned from the interpreter's cache. Thus I am not sure in which situation user should use `sealed`. Do you have some example scenario?

Beside this as far as I see code works and I don't mind merging it. I also can add few advices how to make it more compact and little bit faster, but would like to understand the usage scenario first.

PR #399, opened 2023-08-22
Removing Rc<RefCell<T>> pattern from metta_t in hyperonc

luketpeterson:
In the case of hyperonc's `metta_t`, the `Rc<RefCell<T>>` (aka `Shared<T>`) pattern is unsound and obnoxious to work with.  Also it doesn't add much value.

The usage pattern for `metta_t` lead to a violation of RefCell's contract with `unsafe`.  (using `as_mut` to gain access to the `RefCell`'s inner value while demonstrably not having exclusive access)  Luckily it never caused any problems because the underlying `Metta` object was never changed.  But it was a bug waiting to happen if simple state was ever added to the Metta struct.

Also, The Metta (Rust type) never needs to be mutable because it is 100% internally mutable.  

Finally, the top-level runner has a simple life-cycle, unlike other objects where we rely on reference counting for deallocation and embed handles within other objects (`DynSpace`, `Shared<Tokenizer>`, etc.)

I believe the simplest and best solution is to make the C api use a simple Box for `metta_t`, rather than a `Shared<T>`.  This PR does that.

### DIFF
--- a/c/src/metta.rs
+++ b/c/src/metta.rs
@@ -437,7 +437,7 @@ pub extern "C" fn step_get_result(step: step_result_t,
     }
 }
 
-/// @brief A handle to a top-level MeTTa Interpreter
+/// @brief A top-level MeTTa Interpreter
 /// @ingroup interpreter_group
 /// @note A `metta_t` must be freed with `metta_free()`
 /// @see metta_new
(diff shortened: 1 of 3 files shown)


vsbogd (approved):
Thanks, Luke, agree it doesn't make much sense to have shared MeTTa runner instance.

Necr0x0Der:
Not sure if it happened after this PR, but recently this example didn't crash
https://github.com/trueagi-io/hyperon-experimental/tree/main/python/sandbox/resolve
Now it crashes with
`AssertionError: Method copy should be implemented by grounded object`

```
r.py(32): my_imported_runner_atom
hyperon/ext.py(8): metta_register
hyperon/runner.py(60): load_py_module
hyperon/runner.py(21): <lambda>
hyperon/atoms.py(191): execute
atoms.py(121): _priv_call_execute_on_grounded_atom
hyperon/runner.py(79): run
hyperon/runner.py(72): import_file
metta.py(13): main
metta.py(19): <module>
```

@luketpeterson , could this PR be the reason? Or it happened earlier?

luketpeterson:
I did remove the `copy()` method from the Metta object in my PR.  So yes, that is the cause.

I don't understand what functionality is desired, that this test is intended to check.  Do we want a separate MeTTa runner to be wrapped in an atom, or is there a reason to have the atom contain a reference to the runner that is actively evaluating that atom?  The latter (which is what this test did) sounds like it's asking for bugs.

Necr0x0Der:
This sandbox example implements a way to evaluate a function in the context the space, in which it is declared. Currently, we can only call `match` on another space, but even if we construct an equality query, this query will return the result, which cannot be further evaluated by the current runner, because it can contain symbols defined in the same other space. Unfortunately, there was/is no good API to evaluate an expression without a runner. Thus, in this example, a fresh runner is created for this additional space and it is returned as a grounded atom. We don't want to do a deep copy of the runner itself, though. This sounds extremely inefficient. We wrap spaces into grounded atoms as shared refs I believe. At least, it was the case before, because making deep copy of the whole space each time we encounter `&self` is senseless. Why should we do this with the runner?
This example was intended to start discussing if we want to import libraries as mere spaces or runners, and it showed that representing libraries by runners makes some sense. It is not shown here, but libraries can be remote in this case (the whole interpretation of a library function can be done by an external runner without sending intermediate results back and forth).
While there are other possible solutions to different aspects supposed by this example, e.g. if the interpreter explicitly accepts a space as its argument, then interpreting pure symbols in context of another space can be easily done (although it doesn't solve the problem of referring to grounded symbols defined in the library, which is now done by extending the tokenizer of the main runner, but could be done in a more modular way by keeping these tokens in the runner of the library unless we import this library to `&self`). 
While this example is not absolutely necessary to support, because we may choose a different way of resolving symbols in external libraries, the need for wrapping a runner into a grounded atom can appear in different contexts, so I would argue against throwing this possibility away. 
@vsbogd , what do you think?

luketpeterson:
These are all fair points.  I can re-add the ability to make a shared runner handle.

TBH, my main motive behind this PR was the fact that the old code was abusing RefCell in an unsound way, but we can get back the behavior of a shared handle without using RefCell at all.

I will make this change tomorrow if @vsbogd agrees.  It is late for me tonight.

Necr0x0Der:
Let's keep it as is for now. If there will be a standard solution with the use of wrapping `MeTTa` into grounded objects, we will (re)introduce `copy` or `RefCell`. If someone needs a specific behavior of `copy`, it is always possible to implement it for a child class like in #402 .

Issue #281, opened 2023-04-27
Need better words to conceptually describe states of BindingsSets


luketpeterson:
From @vsbogd  (in code review)
> BindingsSet::single in matcher means that two atoms are equal, and BindingsSet::empty means they cannot be matched. Thus I don't see the use-case for such semantics of _empty() yet. And if this semantics is not used then name can be confusing because one may think it returns true iff no Bindings instances are inside.

I don't feel like this conceptual difference is adequately communicated by the words "empty" and "single".

Maybe "unconstrained" and "fully_constrained"?  Those are clumsy to write and probably not accurate.  Any other ideas?

vsbogd:
> Maybe "unconstrained" and "fully_constrained"? Those are clumsy to write and probably not accurate. Any other ideas?

The instance of `Bindings` is a container for a single set of variable values and equalities. When we match (maybe unify is a better term here, but match is used in the code) two atoms, we can receive a number of such containers (instances of `Bindings`) as a result. Each instance of `Bindings` in this set can be considered as independent from others.

There are three kinds of results:
1. If matching is not successful, then the result is empty set of `Bindings`. 
2. If two atoms being matched are equal, then the result contains a single instance of `Bindings` which doesn't have variables inside.
3. If there are many different options to match two atoms, then the result contains a set of `Bindings`, each instance of `Bindings` contains the unique set of variables and values.

The result of matching can also be represented by a generator instead of a finite set, and it is one of the reasons the `match_atoms` returns the instance of `Iterator`. For now the `BindingsSet` represents the result as a finite set of `Bindings`.

I would like to make the points here:
- The result of the atom matching is not a single set of variable values (i.e. `Bindings`), but a set of such sets;
- User need the ability to construct empty instance of the `BindingsSet` to return it when matching is unsuccessful (situation (1)), it is needed to implement a custom matching interface;
- I am not sure `unconstrained` and `fully_constrained` reflect the situations (1) and (2) respectively.
 
[...]

PR #400, opened 2023-08-22
Improvement of custom query examples

Necr0x0Der:
* "Neurospace" example is extended with a nested query and multiple results returned by custom matching
* A unit test example from test_custom_space is tweaked to support multiple matching results as well to be less misleading

### DIFF
--- a/python/sandbox/neurospace/neurospace.py
+++ b/python/sandbox/neurospace/neurospace.py
@@ -33,7 +33,7 @@ class NeuralSpace(GroundingSpace):
         try:
             val = ValueAtom(int(val))
             bindings = Bindings()
-            bindings.add_var_bindings(var, val)
+            bindings.add_var_binding(var, val)
             new_bindings_set.push(bindings)
         except ValueError:
             ss = re.findall(r'\".*?\"', val)
@@ -42,7 +42,7 @@ class NeuralSpace(GroundingSpace):
             for s in ss:
                 val = S(s[1:-1])
                 bindings = Bindings()
-                bindings.add_var_bindings(var, val)
+                bindings.add_var_binding(var, val)
                 new_bindings_set.push(bindings)
         return new_bindings_set
 
--- a/python/sandbox/neurospace/neurospace.py
+++ b/python/sandbox/neurospace/neurospace.py
@@ -20,21 +20,30 @@ class NeuralSpace(GroundingSpace):
                 temperature=0)
         txt = response['choices'][0]['message']['content']
         res = re.findall(r'\{.*?\}', txt)
-        new_bindings_set = BindingsSet()
+        new_bindings_set = BindingsSet.empty()
         if res == []:
-            return new_bindings_set()
+            return new_bindings_set
         res = res[0][1:-1]
-        var, val = res.split(':')
-        var = re.findall(r'\".*?\"', var)[0][1:-1]
+        _var, val = res.split(':')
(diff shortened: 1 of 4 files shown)


python/sandbox/neurospace/test_nspace.metta
```diff
@@ -13,6 +13,12 @@
 
 ; !(match (my-nspace) (What is the user name $x) $x)
 ; !(match (my-nspace) (How old $x is William) $x)
-!(match (my-nspace) (How old ($x) is Mike) $x)
+; !(match (my-nspace) (How old ($x) is Mike) $x)
 ; !(match (my-nspace) (What $x is father name) $x)
 ; !(match (my-nspace) (What $x is Mike age) $x)
+
+; (match (my-nspace) (What are the names $x of William brothers?) $x)
```
luketpeterson, line 20: Given this is back-ending to an LLM, do you think traditional English grammar is better?  i.e. "What are the names, $x, of William's brothers?" or "What are the names, $x, of the brother's of William?"

Or is a more sparse and regular syntax better?  I am sure the LLM would be able to handle either.

python/sandbox/neurospace/neurospace.py
```diff
@@ -20,21 +20,30 @@ def query(self, query_atom):
                 temperature=0)
         txt = response['choices'][0]['message']['content']
         res = re.findall(r'\{.*?\}', txt)
-        new_bindings_set = BindingsSet()
+        new_bindings_set = BindingsSet.empty()
```
luketpeterson, line 23: Reading this change made stop to think about what the correct behavior for an empty query atom is.  Relates to https://github.com/trueagi-io/hyperon-experimental/issues/281

luketpeterson (dismissed):
Changes look good.  Thanks for fixing the semicolons.  Python is not my most natural language.

Up to you whether you want to make any changes in response to my comments or merge as-is.

python/sandbox/neurospace/test_nspace.metta
Necr0x0Der: Unfortunately `'` doesn't work (in "brother's") because these are not strings. Surrounding `$x` with `,` sometimes work, but gpt seems to be sensitive to the syntax in this case. It can possibly be made slightly more readable in some cases, but the syntax will be less uniform. Since it is just a sandbox example, I'd keep it as is.

Necr0x0Der:
I wonder why ci suddenly failed on the irrelevant change in a different place
```
---- atom::matcher::test::match_variable_with_unique_itself stdout ----
thread 'atom::matcher::test::match_variable_with_unique_itself' panicked at '(left == right some order)
  left: [{ $x = $x#3, $z <- (B $x#3) }]
 right: [{ $x = $x#1, $z <- (B $x#1) }]', src/atom/matcher.rs:1170:9
```

luketpeterson:
The test failure is *very* strange, as it's in a lib (Rust core) test, but neither this PR, nor https://github.com/trueagi-io/hyperon-experimental/pull/399 touches lib at all, and lib is the first thing to be built and tested.

When we got a spurious break in the past it was because of a tool chain upgrade, but that doesn't feel likely in this case.

luketpeterson:
Re-running tests with no changes whatsoever now succeeds.  I don't like intermittent bugs, but I think we just have to wait and see if this pops up again.

luketpeterson (approved):

PR #404, opened 2023-08-24
Fixing Optimized build when nesting a Space within an Atom, Adding new test for C API, etc.

luketpeterson:
There are 3 C API fixes / improvements in this PR:

1. Adding C api test: test_space_nested_in_atom, which corresponds to they Python test `test_match_nested_grounding_space`, but without relying on Python.

2. Fixing oversight in C API in signature of `space_clone_handle()`, which I missed in the C API regularization pass a few weeks ago.

3. Adding `atom_vec_clone()` function, because it's convenient

### DIFF
--- a/c/src/space.rs
+++ b/c/src/space.rs
@@ -39,7 +39,7 @@ impl space_t {
         unsafe{ Box::from_raw(self.space.cast_mut()).0 }
     }
     pub(crate) fn borrow(&self) -> &DynSpace {
-        unsafe{ &(*self.space).0 }
+        unsafe{ &(&*self.space).0 }
     }
 }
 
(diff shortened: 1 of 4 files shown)


c/src/space.rs
```diff
@@ -39,7 +39,7 @@ impl space_t {
         unsafe{ Box::from_raw(self.space.cast_mut()).0 }
     }
     pub(crate) fn borrow(&self) -> &DynSpace {
-        unsafe{ &(*self.space).0 }
+        unsafe{ &(&*self.space).0 }
```
luketpeterson, line 42: FYI, this change should make no difference whatsoever to the behavior, but it more clearly communicates how I like to think about what is happening.

luketpeterson:
Added fix for release build segfault (https://github.com/trueagi-io/hyperon-experimental/issues/405) in 1681a3e commit.

vsbogd (approved):

vsbogd:
It makes sense probably to switch CI build to the release mode. Then we will detect such release build issues earlier. What do you think?

luketpeterson:
> It makes sense probably to switch CI build to the release mode. Then we will detect such release build issues earlier. What do you think?

I think it's a good idea.  The main reason being that it will give the faster performance to users.  The downside is that it'll make crash reports from the wild a little harder to follow, but I think the performance outweighs this concern.

PR #406, opened 2023-08-24
small addition to neurospace example

Necr0x0Der:


Necr0x0Der:
It is just a small extension to a sandbox example, so no review is required, but fill free to leave comments and propose extensions / improvements.

PR #412, opened 2023-08-28
testing metta/torch interoperability

Necr0x0Der:


Necr0x0Der:
Some Python libraries maintain a global state, which should be preserved, when this library is used from MeTTa (being wrapped as its extension). The initial idea of test_torch was to test this functionality on the concrete practical example. But since torch is heavy, and we may want to test its operability on GPU within MeTTa as well, this test was moved to integration tests and replaced with the simple test on a global state. Basically, we want to check that the Python interpreter and its environment are the same in the host Python script and within the MeTTa runner.
FYI, @luketpeterson

luketpeterson:
> we want to check that the Python interpreter and its environment are the same in the host Python script and within the MeTTa runner.

As long as the repl has the directory containing the Python script in the search path*, all of those tests pass with the repl.

*The search path will always include the working directory, and it can be extended with the config and using arguments when the repl is invoked, e.g. `-i ../python/tests`

python/tests/test_extend.py
```diff
@@ -20,3 +23,29 @@ def test_extend(self):
...
+        metta.run('''
+          !(extend-py! extension)
+          !(set-global! 42)
+        ''')
+        # Checking that the object is accessible and its value is correct
+        self.assertEqual(
+          metta.run('!(get-global)'),
+          metta.run('! 42'))
+        # We may have to reimport g_object, but its value should be set
+        from extension import g_object
```
vsbogd, line 46: Minor but we also can use `import` with `extension.g_object` instead of `from` to make it work without re-importing:
```python
        import extension
        # Sanity check
        self.assertEqual(extension.g_object, None)
        metta = MeTTa()
        metta.run('''
          !(extend-py! extension)
          !(set-global! 42)
        ''')
        # Checking that the object is accessible and its value is correct
        self.assertEqual(
          metta.run('!(get-global)'),
          metta.run('! 42'))
        self.assertEqual(extension.g_object, 42)
```

vsbogd (approved):
Looks good to me.

python/tests/test_extend.py
Necr0x0Der: I've added it as a comment because the remark about re-importing can also be helpful

PR #415, opened 2023-08-29
states vs replace atoms benchmark

Necr0x0Der:


vsbogd:
@luketpeterson , which micro-benchmarking framework do you use? Does it support benchmark parameters? As far as I understand `bench` doesn't support it and thus we ended up introducing `xN` functions and comparing their values manually. Probably we should move to another benchmark framework, criterion.rs looks like a good candidate from parameterization perspective.

Necr0x0Der:
Yes, I might need to mark this PR as draft, because I had the same question.

luketpeterson:
> @luketpeterson , which micro-benchmarking framework do you use?

I have used criterion, and it worked pretty well for me.  I can't comment specifically on parameterization however.  It looks like a new maintainer has taken it over and v. 0.5 was released , which is good because I was worried for a while that it may have been on the path to be abandoned.

The one thing I'll say is that it is *heavy* so it definitely needs to be a dev-only dependency.

Necr0x0Der:
@luketpeterson , I'dd prefer to merge it as is for now if there are no strong objections, and refactor later when a better approach for benchmarking is selected.

luketpeterson (approved):

luketpeterson:
Sure.  We can redo the benchmark framework in the future.  Merging.

Issue #390, opened 2023-08-07
Chaining of functions with "empty" results


Necr0x0Der:
There is a recurrent request for chaining functions with side effects with no return value, e.g. `add_atom`, `println!`, etc. We've discussed this for a few times in different places, and this issue is a sort of conclusion to which we come with @vsbogd in the last discussion.
While there are different practical ways of achieving this functionality, the issue boils down to semantics of "empty" result. MeTTa considers all functions as non-deterministic, and an empty result means that the function has produced no evaluation branches, and nothing is to be passed to its super-expression. Basically, the function is not total. In most cases, it is really what is expected, e.g. `+` expects two numbers, and if one its argument is empty, it has nothing to evaluate. `match` can normally return an empty result, which will mean exactly the same. A pure MeTTa function can be defined not for all values of its arguments, and being implemented via `match` will also be evaluated to the empty set of results (once again, not total).
However, grounded functions with side effects may return "nothing". It is natural to return an empty set of results from such functions, although it is technically incorrect. In functional programming, when a function has `Unit` as a return type (which can be represented by `()`), it means that it "returns" the only element of one-element set, thus, its result contains no information, but it is still a non-empty result. In pure functional programming, functions that really return nothing (which can be referred to as `Void`, although in other languages `void` can mean `unit`) are "absurd" and they are not possible (although most such languages have non-strict components to represent the notion of such impossibility).
We also need to distinguish "void" and "unit". Minimal MeTTa ( #380 ) introduces `EMPTY_SYMBOL` to represent an empty set of results, but it doesn't introduce `UNIT_SYMBOL` (or `Void`, `Singleton` depending on naming preferences) and treats an empty set returned by grounded functions as such `unit` (is it so?) - that can also look natural for those who are not hard-core functional programming geeks, because "these functions really return nothing". Having `EMPTY_SYMBOL` may still be convenient, although it may also cause confusion, because right now it is implemented in pure MeTTa as `(superpose ())`, which will be treated differently than `EMPTY_SYMBOL` by Minimal MeTTa. Thus, it might be better to have `EMPTY_SYMBOL` not as a special atom, but just as syntactic sugar for `(superpose ())`, but this may depend on the implementation.
[...]

PR #429, opened 2023-09-12
Minimal MeTTa interpreter fixes and optimizations

vsbogd:
Next batch of minimal MeTTa related fixes and changes:
- `minimal` feature is added to switch core library to the minimal MeTTa interpreter at compile time, one can run Rust unit tests using `cargo test --features minimal`
- minimal MeTTa's `match` is renamed to `unify` because previous `match` definition has different order of arguments and thus incompatible with new one, old `match` is added into minimal MeTTa standard library
- pattern matcher syntax extensions suggestions are added to the minimal MeTTa docs, docs are fixed
- `assert...`, `collapse`, `superpose`, `let` and `let*`, `case`, `match` and `pragma!` are added into minimal MeTTa standard library, some of them are added as grounded atoms because they require the working atomspace to evaluate arguments which is not implemented yet
- `Void` is added as a result of `assert...` operation, this is to handle https://github.com/trueagi-io/hyperon-experimental/issues/390 in minimal MeTTa
- `NotReducible` is returned instead of `Empty` by minimal MeTTa interpreter when function definition is not found. This corresponds to the previous interpreter's behavior. Two major use-cases are type constructors (which are usually have no definition only type declaration) and partially defined functions (which return results on some arguments and stay non-reducible on others).
- benchmark for the nested `chain` operation is added and `chain` implementation is optimized
- few small fixes

Better to review commit by commit.

### DIFF
--- a/c/src/metta.rs
+++ b/c/src/metta.rs
@@ -275,6 +275,14 @@ pub extern "C" fn sexpr_parser_parse(
 ///
 #[no_mangle] pub extern "C" fn ATOM_TYPE_GROUNDED_SPACE() -> atom_t { rust_type_atom::<DynSpace>().into() }
 
+/// @brief Creates a Symbol atom for the special MeTTa symbol used to indicate empty results in
+/// case expressions.
+/// @ingroup metta_language_group
+/// @return  The `atom_t` representing the Void atom
+/// @note The returned `atom_t` must be freed with `atom_free()`
+///
+#[no_mangle] pub extern "C" fn VOID_SYMBOL() -> atom_t { hyperon::metta::VOID_SYMBOL.into() }
+
 /// @brief Checks whether Atom `atom` has Type `typ` in context of `space`
 /// @ingroup metta_language_group
 /// @param[in]  space  A pointer to the `space_t` representing the space context in which to perform the check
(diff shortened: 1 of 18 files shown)


vsbogd:
Added `minimal` feature to the Rust REPL in order to use it with minimal MeTTa.

Unfortunately I see the result of execution from old Python repl and new Rust repl are different. For instance the Nil's example:
```
(: foo (-> Number Number))
(= (foo $x) 123)
(= (Foo) "abc")
!(foo (Foo))
```
returns `[(Error "abc" BadType)]` in Python and `[123]` in Rust. At the same time from the log I see that new minimal interpreter is used in both cases. In Rust REPL's log I see an additional query is made after each `eval`. May be this difference causes the different result.

vsbogd:
> Unfortunately I see the result of execution from old Python repl and new Rust repl are different.

Ok, I have found the root cause: in Python interpreter `"abc"` is considered as a grounded string with specific type `String`, in Rust REPL it is considered as a symbol atom with `%Undefined%` type, thus it can be also casted to `Number`

Necr0x0Der (approved):
Looks convenient to start playing with Minimal MeTTa

vsbogd:
Merging it after confirming with @Necr0x0Der  and @luketpeterson

PR #430, opened 2023-09-12
Add text operators in stdlib.py

ngeiswei:
Add the following operators:
1. `repr`: convert `Atom` to `String`,
2. `unrepr`: convert `String` to `Atom`,
3. `toChars`: convert `String` to tuple of `Char`,
4. `toString`: convert tuple of `Char` to `String`.

See the unit tests for examples.


ngeiswei:
Please treat this PR as a proof-of-concept.  This may not be very good code, there are some places that might be dangerous, like the fact that it takes substrings without any checking, etc.

Also the names are probably bad, please find better names if you can (I'll apply the changes accordingly).

vsbogd:
Risk of `repr/unrepr` is that for some grounded atom `repr()` can generate the text which is not properly parsed by parser. It is true for atomspaces for example. It is not a problem of this PR though, and we can state that user which uses `repr/unrepr` knows what he/she is doing and aware that this doesn't work for specific atoms.

Regarding namings I would propose something like `stringToChars/charsToString` instead of `toChars` and `toString` because latter don't reflect the "from" part of the conversion. `unrepr` probably could be renamed to `parse`, because it is what it does.

Third question is that this functions needs to be duplicated in Rust before we have some marshaling of data between Rust and Python.

Necr0x0Der:
@ngeiswei , @vsbogd , I have a particular use case for working with strings <-> symbols in MeTTa for LLM prompt parameterisation. I want to review this PR in context of this use case, but I will possibly do this not too quickly (from a few days to a week). Is it ok?

vsbogd:
>  I want to review this PR in context of this use case, but I will possibly be able do this not immediately (from a few days to a week). Is it ok?

Its absolutely fine with me.

luketpeterson:
I am happy to merge this functionality for now (once others review).  In my opinion String functionality is central enough that it belongs in Rust (and then Python will get it also)

Probably in 2 months or so I will take the project to migrate this into the Rust stdlib, at the time when I tackle the design for https://github.com/trueagi-io/hyperon-experimental/issues/351

In the meantime, we can converge on what we want the string operations to do using Python.

Necr0x0Der:
> However, In my opinion String functionality is central enough that it belongs in Rust (and then Python will get it also). ... I will take the project to migrate this into the Rust stdlil

JIC, what do you mean by "Python will get it also" and by "migrate"? I would say that Python strings are much more useful (in terms of use cases) than Rust. Having them in Rust looks more optional to me. **Migrating** (not reproducing) them to Rust will make sense only if we have mechanisms to get them in Python (semi-)automatically.

luketpeterson:
Apologies.  Accidentally edited your comment, when I meant to quote-reply

> JIC, what do you mean by "Python will get it also" and by "migrate"? I would say that Python strings are much more useful (in terms of use cases) than Rust. Having them in Rust looks more optional to me. Migrating (not reproducing) them to Rust will make sense only if we have mechanisms to get them in Python (semi-)automatically.

What I mean is that the functionality should be implemented in such that is works seamlessly from Python, Rust, C, any other host languages we want in the future. https://github.com/trueagi-io/hyperon-experimental/issues/351 proposes a mechanism to bridge primitive types in the host language to grounded atoms in a way that's efficient and portable. For example, operations implemented in the Rust stdlib would operate on primitives coming from Python, and the Python user would never need to know or care where the operation was implemented.

ngeiswei:
I have renamed `unrepr` to `parse`, `toString` to `charsToString` and `toChars` to `stringToChars`.  I like `parse`.  For `charsToString`, depending on the conventions (Haskell, Scheme, etc), `chars->string` and `string->chars` could work to as well I suppose (I have no personal preference).

I can use that branch for now, so, @Necr0x0Der, no problem with me to wait a few days before it gets merged.

ngeiswei:
Something else I wanted to ask.  With that operator set, we don't, in principle, need to add functions to process strings like insert a char, etc, cause we can manipulate tuples of chars with `car-atom`, etc.  Yet it could still be beneficial to have them as grounding functions, it'd be more efficient than having them in pure MeTTa.  I did not want to make that decision in that PR, but I thought I'd ask nonetheless.

ngeiswei:
Thinking about it more, another option would be to replace `stringToChars`/`charsToString` by `head`/`tail`/`append`, or by `take`/`drop`/`insert`, etc, to operate directly on `String`, then `stringToChars`/`charsToString` could be implemented from those primitives.  I don't know what is best.

luketpeterson:
I'll let @Necr0x0Der and @vsbogd answer with regard to MeTTa philosophy.

But my opinion more generally depends on what your use case is.  In other words, if taking the last char / first char off a string is a common operation for you, then maybe it makes sense to expose a `split` operator that takes a string index and splits into two substrings, since this would be more direct than decomposing a string into an expression of chars only use car/cdr and then recreate the string again.

On the other hand, if your use case is to create an expression atom with a child for each char of a string, then an operation to do that makes sense, instead of iterating to do it one char at a time.

IMO every mature language will have a rich string library available to it, and MeTTa should too.  Ultimately I think usability is a major guiding goal.  MeTTa will probably never be Perl, as far as text-manipulation power, but we'd ideally get it to the point where it's reasonably easy to do most basic things.
 
Also, MeTTa's matching / unification semantic has some interesting possible applications to string parsing, as an alternative to RegEx.  Just a thought.

vsbogd:
Hm, I would not think a lot about which set of the functions should be a basis. As long as implementation is cheap we can implement all we need as a part of the string library and use it. If in future we have better understanding how to implement it we can implement it again.

To me the more important is to move these functions into separate `string` library and have a separate namespace to prevent conflicts of the names.

python/hyperon/stdlib.py
```diff
@@ -52,6 +53,33 @@ def bool_ops():
...
+
+    see test_stdlib.py for examples.
+
+    """
+    reprAtom = OperationAtom('repr', lambda a: [ValueAtom(repr(a))],
+                             ['Atom', 'String'], unwrap=False)
+    parseAtom = OperationAtom('parse', lambda s: [ValueAtom(SExprParser(str(s)[1:-1]).parse(Tokenizer()))],
+                              ['String', 'Atom'], unwrap=False)
+    stringToCharsAtom = OperationAtom('stringToChars', lambda s: [ValueAtom(E(*[ValueAtom(Char(c)) for c in str(s)[1:-1]]))],
+                                      ['String', 'Atom'], unwrap=False)
```
Necr0x0Der, line 73: We could possibly use `Expression` instead of `Atom` here.

Necr0x0Der (approved):
I'm merging it now. I don't like `toChar` and `toString` too much, because they don't say what is converted (and one may want to have `toString` for other types of the argument). But all the improvements can be done later.

python/hyperon/stdlib.py
ngeiswei: What is the difference between `Atom` and `Expression` (I tried to understand the difference by reviewing code and examples but couldn't)?  And why would `Expression` be better than `Atom` in that context?

ngeiswei:
> I don't like `toChar` and `toString` too much, because they don't say what is converted (and one may want to have `toString` for other types of the argument).

Note that `toChars` (resp. `toString`) was renamed to `stringToChars` (resp. `charsToString`).

Necr0x0Der:
> > I don't like `toChar` and `toString` too much, because they don't say what is converted (and one may want to have `toString` for other types of the argument).
> 
> Note that `toChars` (resp. `toString`) was renamed to `stringToChars` (resp. `charsToString`).

Thanks. I wonder why I saw the older version.

python/hyperon/stdlib.py
Necr0x0Der: For `(: f (-> Expression Atom))`, `(f A)` should not type check, while `(f (A B))` should.
JIC, I put my comment under definition of `charsToStringAtom`, where `a.get_children()` is used (but it was put under `stringToCharsAtom`). Using `Expression` type should prevent `get_children` to be called in case, when non-expression atom is passed as an argument.

Necr0x0Der: Hmm... It doesn't seem to work now, after all. `car-atom` throws a error that it expects `Expression`, but it is not due its type, but it's hardcoded in its Rust implementation. However, `(f A)` gets reduced, if, say, `(= (f $x) $x)`.

ngeiswei: Oh, so `Atom = Symbol + Expression`, correct?

Necr0x0Der: `Symbol`, `Expression`, `GroundedAtom` are subtypes for `Atom`.

ngeiswei: OK, so should I conclude that (as of today)
1. `Atom = Symbol + Expression + GroundedAtom`
2. None of `Symbol`, `Expression` and `GroundedAtom` are subtypes of one another?

vsbogd: yes

PR #436, opened 2023-09-15
Tiny: Display pointer to GroundingSpace to distinguish different spaces

vsbogd:
Tiny PR to address the issue that separate spaces are not distinguishable in a log.


lib/src/space/grounding.rs
```diff
@@ -330,13 +330,13 @@ impl PartialEq for GroundingSpace {
 
 impl Debug for GroundingSpace {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "GroundingSpace")
+        write!(f, "GroundingSpace-{:?}", self as *const GroundingSpace)
```
luketpeterson, line 333: Is the idea that a memory address should be printed for the GroundingSpace?

fmt has a markup for pointers.  https://doc.rust-lang.org/std/fmt/trait.Pointer.html

Necr0x0Der (dismissed):

lib/src/space/grounding.rs
Necr0x0Der: @vsbogd , my approval means I'm agnostic wrt @luketpeterson comment.

vsbogd: Thanks, Luke, applied suggestion in https://github.com/trueagi-io/hyperon-experimental/pull/436/commits/1fb6ee8d69d90894a15e80f1e8ae5b0fdd58bcbc

luketpeterson (approved):

PR #446, opened 2023-09-27
A sketch of DAS integration

Necr0x0Der:
This is a very sketchy and possibly not a canonical (in future) way to integrate DAS into MeTTa, but it might be valuable to have, because there are some potential issues on both sides.

### DIFF
--- a/python/sandbox/das_gate/dasgate.py
+++ b/python/sandbox/das_gate/dasgate.py
@@ -1,0 +1,164 @@
+from hyperon import *
+from hyperon.ext import register_atoms
+
+import os
+import re
+import json
+
+from hyperon_das import DasAPI
+from hyperon_das.das import QueryOutputFormat as Format
+from hyperon_das.pattern_matcher import (
+    Link,
+    Node,
+    Variable,
+    And,
+    Or,
+    Not,
+    PatternMatchingAnswer,
+)
+
+'''
+def to_nested_expr(xs):
+    if isinstance(xs, list):
+        return E(*list(map(to_nested_expr, xs)))
+    return ValueAtom(xs)
+
+def _response2bindings(txt):
+        res = re.findall(r'\{.*?\}', txt)
+        new_bindings_set = BindingsSet.empty()
+        if res == []:
+            return new_bindings_set
+        res = res[0][1:-1]
+        _var, val = res.split(':')
+        var = re.findall(r'\".*?\"', _var)
+        var = var[0][1:-1] if len(var) > 0 else _var.replace(" ", "")
+        if var[0] == '$':
+            var = var[1:]
+        var = V(var)
+        try:
+            val = ValueAtom(int(val))
+            bindings = Bindings()
+            bindings.add_var_binding(var, val)
+            new_bindings_set.push(bindings)
+        except ValueError:
+            ss = re.findall(r'\".*?\"', val)
+            if ss == []:
+                ss = ['"' + val + '"']
+            for s in ss:
+                val = S(s[1:-1])
+                bindings = Bindings()
+                bindings.add_var_binding(var, val)
+                new_bindings_set.push(bindings)
+        return new_bindings_set
+
+class NeuralSpace(GroundingSpace):
+    def query(self, query_atom):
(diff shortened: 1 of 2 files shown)


Necr0x0Der:
Further work on DAS integration requires some steps on the DAS side, so I'd prefer to merge it for now as a sandbox example

python/sandbox/das_gate/dasgate.py
```diff
@@ -0,0 +1,94 @@
...
+    And,
+    Or,
+    Not,
+    PatternMatchingAnswer,
+)
+
+class DASpace(AbstractSpace):
+
+    def __init__(self, unwrap=True):
+        super().__init__()
```
luketpeterson, line 23: Not technically needed since AbstractSpace is pure-virtual, but also harmless.

python/sandbox/das_gate/dasgate.py
```diff
@@ -0,0 +1,94 @@
...
+        return new_bindings_set
+
+    def add(self, atom):
+        dc = self._atom2dict(atom)
+        if "name" in dc:
+            self.das.add_node(dc)
+        else:
+            self.das.add_link(dc)
+
+    #def remove(self, atom):
```
luketpeterson, line 79: Without `remove` and `replace` implementations, an exception will be raised if they're called.

Do we want to modify the Space contract so that a Space can gracefully decline to implement these functions in the same way a Space can decline to implement `atom_count` and `atoms_iter`?  My feeling is "no", but you have a more coherent conceptual vision for MeTTa than I do.

It seems to me that `remove` and `replace` are different from `atom_count` and `atoms_iter` in that `remove` and `replace` affect the Space.  Even in a space where it's impossible to "forget" (e.g. an append-only journal) it's still possible to indicate that an atom is disused or has been superseded.

luketpeterson (approved):
The use of the Space API looks fine with the caveats
1. The comment RE `remove` and `replace`
2. I was unable to actually test the code, so I merely read it

I am not able to comment on the use of the DAS protocol.

python/sandbox/das_gate/dasgate.py
```diff
@@ -0,0 +1,94 @@
+from hyperon import *
+from hyperon.ext import register_atoms
+
+import os
+import re
+import json
+
+from hyperon_das import DasAPI
```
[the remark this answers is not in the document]
Necr0x0Der: Unfortunately, their repo is private now, so this PR was more for information than for the in-depth review. Of course, the link to DAS will be provided, once it is open.

python/sandbox/das_gate/dasgate.py
Necr0x0Der: OK, I see. I'd still keep it here, because if some useful functionality is introduced in the parent class, then it will not be missed in DASpace as well. I'm also ok to remove it if `Abstract` part is a strong hint that no concrete functionality will be ever introduced.

python/sandbox/das_gate/dasgate.py
Necr0x0Der: > Without remove and replace implementations, an exception will be raised if they're called.

This is perfectly ok. 

> Do we want to modify the Space contract so that a Space can gracefully decline to implement these functions

What behavior will be in this case for `remove`, when it is not implemented? I believe it should be exception. A space may not implement some required functionality, but attempts to use the missing functionality should be prevented, and the user should be informed that there is no such functionality and it cannot be used.

> in the same way a Space can decline to implement atom_count and atoms_iter?

Do you mean that if `atoms_iter` is not implemented, and `get-atoms` is executed in MeTTa, a MeTTa `Error`-expression is returned, while if `remove-atom` is executed the exception is raised, which is not turned into `Error`, but the whole MeTTa runner stops? I'm ok with this difference.

vsbogd: >  if remove-atom is executed the exception is raised, which is not turned into Error, but the whole MeTTa runner stops? I'm ok with this difference.

Does it really work by this way? I believe the exception should be wrapped into `Error` as well.

PR #451, opened 2023-10-04
Sql space

besSveta:
add sql wrapper with possibility to select/insert data from / into [PostgreSQL](https://www.postgresql.org/) database


Necr0x0Der (approved):
I think it's ok as a sandbox example, which can be refined if/when we/someone will need it.

PR #452, opened 2023-10-04
introducing unit result in grounded functions

Necr0x0Der:


Necr0x0Der:
This is an attempt to address the issue #390 (or the first step towards it without introducing chaining, which might be already done in Minimal MeTTa).
Some remarks:
* From the category-theoretic point of view, `()` is a unit, and `(->)` is a unit type, so I used them.
* I’m not sure if there is a way to introduce `UNIT_ATOM` and `UNIT_TYPE` as constants in Rust; please, make it prettier if possible;
* It is not precisely clear if `nop` should return an empty result or unit. At first, I encountered the test, in which the empty result is useful. But later I encountered two tests, where unit is more convenient. Thus, I made `nop` into returning unit. Then, I found one more test requiring `nop` to return the empty result.... I used `(superpose ())` instead of `(nop)` in these cases, and kept `nop` with unit return type. But if someone has strong preferences, this can be changed.
* `()` is used in Python and MeTTa scripts, thus `UNIT` is used only as an alias in Rust for now. Not sure if we need special symbols for unit and unit types in MeTTa. Personally, I find `()` and `(->)` convenient and transparent enough, but special names/symbols might also be convenient. Not sure.

Please, @vsbogd take a look in sense if this is a right direction towards bridging the gap with Minimal MeTTa, and will changes in Minimal MeTTa to adopt this approach (explicit unit) be convenient, or something else should be done?

vsbogd (commented):
> I’m not sure if there is a way to introduce UNIT_ATOM and UNIT_TYPE as constants in Rust; please, make it prettier if possible;

Yeah, I believe it is not possible for now, thus keeping them as functions is ok

>    It is not precisely clear if nop should return an empty result or unit.

I feel `nop` has two roles in Rust MeTTa interpreter. In minimal MeTTa to get empty result one can use `Empty` instead. Thus keeping `nop` as an atom which consumes the result and returns unit is ok with me.

>   () is used in Python and MeTTa scripts

Probably makes sense taking into account it is easy to instantiate unit in Python comparing to Rust.

> Please, @vsbogd take a look in sense if this is a right direction towards bridging the gap with Minimal MeTTa, and will changes in Minimal MeTTa to adopt this approach (explicit unit) be convenient, or something else should be done?

Yes, it looks absolutely aligned with minimal MeTTa implementation, thus I am ready to merge. There is a minor thing: I would move `UNIT` definitions into the `metta` module but I can do it after merge in a commit adapting minimal MeTTa branch to this change.

python/tests/test_minecraft.py
```diff
@@ -30,6 +30,7 @@ def newMineOp(inventory):
         lambda obj, tool: mine_op(inventory, obj, tool),
         unwrap=False)
 
+@unittest.skip("no reason")
```
vsbogd, line 33: Is this test ignored by accident? I tried locally and it passes, but `test_minecraft_planning_with_abstractions` has no asserts inside.
```suggestion
```

Necr0x0Der: Oops, sorry. I turned this test off while playing with Minimal MeTTa (because it was taking too much time), and occasionally added this change to this PR.

vsbogd (approved):

PR #462, opened 2023-10-19
Rework rust test functions & add stdlib comparison operators

luketpeterson:
This PR addresses the comment in `lib/metta/mod.rs` about the functions `metta_space` and `metta_atom`.

These functions are both moved into a test-only module called `common::test_utils`, and they are re-implemented in terms of a test-only runner.

NOTE: This PR sits on top of the extensive https://github.com/trueagi-io/hyperon-experimental/pull/450 PR, but actually has only one unique commit.

### DIFF
--- a/lib/examples/sorted_list.rs
+++ b/lib/examples/sorted_list.rs
@@ -18,9 +18,9 @@ fn main() -> Result<(), String> {
     "))?;
 
     assert_eq!(metta.run(SExprParser::new("!(insert 1 Nil)"))?[0],
-        vec![metta.parse_one_atom("(Cons 1 Nil)")?]);
+        vec![SExprParser::new("(Cons 1 Nil)").parse(&*metta.tokenizer().borrow())?.unwrap()]);
     assert_eq!(metta.run(SExprParser::new("(insert 3 (insert 2 (insert 1 Nil)))"))?[0],
-        vec![metta.parse_one_atom("(Cons 1 (Cons 2 (Cons 3 Nil)))")?]);
+        vec![SExprParser::new("(Cons 1 (Cons 2 (Cons 3 Nil)))").parse(&*metta.tokenizer().borrow())?.unwrap()]);
 
     Ok(())
 }
--- a/lib/examples/sorted_list.rs
+++ b/lib/examples/sorted_list.rs
@@ -1,8 +1,8 @@
-use hyperon::metta::*;
-use hyperon::metta::interpreter::*;
+use hyperon::metta::runner::*;
 
-fn main() {
-    let space = metta_space("
+fn main() -> Result<(), String> {
+    let metta = Metta::new(None);
+    metta.run_program_str("
         (: List (-> $a Type))
         (: Nil (List $a))
         (: Cons (-> $a (List $a) (List $a)))
@@ -15,11 +15,12 @@ fn main() {
         (= (insert $x (Cons $head $tail)) (if (< $x $head)
                                               (Cons $x (Cons $head $tail))
                                               (Cons $head (insert $x $tail))))
-    ");
+    ")?;
 
+    assert_eq!(metta.run_program_str("!(insert 1 Nil)")?[0],
+        vec![metta.parse_one_atom("(Cons 1 Nil)")?]);
+    assert_eq!(metta.run_program_str("(insert 3 (insert 2 (insert 1 Nil)))")?[0],
+        vec![metta.parse_one_atom("(Cons 1 (Cons 2 (Cons 3 Nil)))")?]);
 
-    assert_eq!(interpret(&space, &metta_atom("(insert 1 Nil)")),
-        Ok(vec![metta_atom("(Cons 1 Nil)")]));
-    assert_eq!(interpret(&space, &metta_atom("(insert 3 (insert 2 (insert 1 Nil)))")),
-        Ok(vec![metta_atom("(Cons 1 (Cons 2 (Cons 3 Nil)))")]));
+    Ok(())
 }
--- a/lib/examples/sorted_list.rs
+++ b/lib/examples/sorted_list.rs
@@ -1,8 +1,8 @@
-use hyperon::metta::runner::*;
+use hyperon::metta::{runner::*, text::SExprParser};
 
 fn main() -> Result<(), String> {
     let metta = Metta::new(None);
-    metta.run_program_str("
+    metta.run(SExprParser::new("
         (: List (-> $a Type))
         (: Nil (List $a))
         (: Cons (-> $a (List $a) (List $a)))
@@ -15,11 +15,11 @@ fn main() -> Result<(), String> {
         (= (insert $x (Cons $head $tail)) (if (< $x $head)
                                               (Cons $x (Cons $head $tail))
                                               (Cons $head (insert $x $tail))))
-    ")?;
+    "))?;
 
-    assert_eq!(metta.run_program_str("!(insert 1 Nil)")?[0],
+    assert_eq!(metta.run(SExprParser::new("!(insert 1 Nil)"))?[0],
         vec![metta.parse_one_atom("(Cons 1 Nil)")?]);
-    assert_eq!(metta.run_program_str("(insert 3 (insert 2 (insert 1 Nil)))")?[0],
+    assert_eq!(metta.run(SExprParser::new("(insert 3 (insert 2 (insert 1 Nil)))"))?[0],
         vec![metta.parse_one_atom("(Cons 1 (Cons 2 (Cons 3 Nil)))")?]);
 
     Ok(())
--- a/lib/src/metta/interpreter2.rs
+++ b/lib/src/metta/interpreter2.rs
@@ -575,9 +575,8 @@ mod tests {
     fn interpret_atom_evaluate_pure_expression_variable_name_conflict() {
         let space = space("(= (foo ($W)) True)");
         let result = interpret_atom(&space, atom("(eval (foo $W))", bind!{}));
-        assert_eq!(result[0].0, metta_atom("True"));
+        assert_eq!(result[0].0, sym!("True"));
     }
-
 
     #[test]
     fn interpret_atom_evaluate_grounded_expression() {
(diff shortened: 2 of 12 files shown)


lib/src/metta/runner/mod.rs
```diff
@@ -241,11 +241,43 @@ impl Metta {
         self.0.settings.borrow().get(key.into()).map(|a| a.to_string())
     }
 
+    /// Runs a MeTTa program expressed as a string of characters
+    pub fn run_program_str(&self, program: &str) -> Result<Vec<Vec<Atom>>, String> {
```
vsbogd, line 245: I would not add this method, it is trivially replaced by `metta.run(SExprParser::new(program))`

lib/src/metta/runner/mod.rs
```diff
@@ -241,11 +241,43 @@ impl Metta {
...
+        let atom = parser.parse(&*self.tokenizer().borrow())?;
+        if let Some(atom) = atom {
+            Ok(atom)
+        } else {
+            panic!("Expected a single atom");
+        }
+    }
+
+    /// Parses a string into a vector of Atoms
+    pub fn parse_all_atoms(&self, text: &str) -> Result<Vec<Atom>, String> {
```
vsbogd, line 267: This method has the only usage and can be implemented there.

lib/src/metta/runner/mod.rs
```diff
@@ -241,11 +241,43 @@ impl Metta {
...
+        self.run(parser)
+    }
+
     pub fn run<'p, 'a: 'p>(&'a self, parser: SExprParser<'p>) -> Result<Vec<Vec<Atom>>, String> {
         let state = RunnerState::new_with_parser(self, parser);
         state.run_to_completion()
     }
 
+    /// Parses a string into a single Atom
+    pub fn parse_one_atom(&self, text: &str) -> Result<Atom, String> {
```
vsbogd, line 256: I would replace usages in `sorted_list.rs` by macros and move the code of the method to the `test_utils.rs`.

vsbogd (commented):
Looks like reimplementing `metta_atom` and `metta_space` functions using Metta raises a lot of questions. I agree with changes in API usage examples from  `types.rs`. `interpreter2.rs` should be reverted back. All other changes I would not apply. I would instead wait for minimal MeTTa migration and after that rework arithmetics library and may be remove the `Operation` at all.

lib/src/metta/runner/mod.rs
luketpeterson: This is a nuanced point, but here is my thinking:

1. An interpreter acts on just atoms.  Parsing happens before the interpreter.
2. It makes sense to have the ability to run raw atoms, rather than forcing the user of the runner to first render them to text and re-parse them.  This is done with `RunnerState::new_with_atoms()`
3. This function (`Metta::parse_all_atoms()`) is the other half of that flow.  It lets the caller parse a string of MeTTa code as the runner would.

So, why isn't `parse_all_atoms` a `SExprParser` method, instead of the `Metta` method?
1. The parser has an implicit dependency on a certain Tokenizer, in order to create the right atoms from the input text.
2. Tokens get loaded into the Tokenizer as an effect of running MeTTa code (e.g. `bind!`) or loading modules, and therefore the Tokenizer is part of the Runner's state.

lib/src/metta/runner/mod.rs
luketpeterson: I explained my reasoning behind the `parse_all_atoms` public API above.

The main reason I added a second method here.  (`parse_one_atom`) was that I didn't want the overhead of creating a Vec<> for each trip through a loop that consumes atoms one at a time.  This seemed like it might be a common pattern for a caller.

However, I don't feel as strongly about this, and I wouldn't object to making it private or rolling it into its caller(s)

vsbogd:
Luke, I apologize I didn't explain my comments well enough. My main point is that PR introduces a lot of public methods while main usages are in unit tests. And honestly speaking the only reason of using it in unit tests is that there is no convenient method to instantiate a complex atom in Rust. When I am writing the custom Rust code to represent atoms I would prefer instantiate these atoms instead of introducing overhead of a parser execution.

1. `run_program_str`
Let's take one example `let space = metta_space("(: foo (-> A B)) (: a A) (: b B)");` is replaced by
```rs
let metta = Metta::new(None);
metta.run_program_str("(: foo (-> A B)) (: a A) (: b B)").unwrap();
```
but we could replace it by
```rs
let metta = Metta::new(None);
metta.run(SExprParser::new("(: foo (-> A B)) (: a A) (: b B)")).unwrap();
```
Result is the same and no additional public method.

2. `parse_all_atoms` yes, you are right that parsing involves tokenizer and tokenizer at the moment is a part of Metta. But I would make a design for #470 to understand how API should look like before introducing this public method. At the moment `parse_all_atoms` is used only in `metta_space` and it in turn used only in unit tests and in most (if not all) of the unit tests will vanish after type checker is implemented in MeTTa. One reason why we should be careful when using top level tokenizer is that tokenizer may become different for different imported modules. Thus parsing should always be done in some context and this context may be set explicitly when calling parser. Saying this I don't mean it is a only way just mean we didn't yet make this decision thus introducing this method for unit tests seems to me premature.

3. `parse_one_atom` and `metta_atom` I would actually make a convenient macros to instantiate single atom instead. The whole function is a hack to not create such macros. I tried to implement it as a declarative macros (`expr!` and `sym!`) but I don't think they are convenient and I would try to write a procedural macros instead to make a neat syntax instead of `("sym" var {gnd})`.

To summarize there is a little usage for these methods at the moment, most of usages are temporary and I would expect the API will change during implementing #470, thus I don't see a strong reason to introduce these public methods. When I asked you to move these code into a test module I thought about very simple change. May be simple change is not enough here but I would prefer doing nothing then.

Answering your question about `Operation`. I have tried to implement a struct which can be used to easily wrap Rust operators into grounded operations of MeTTa. One goal was to have a static instance of the grounded operation and use it without copying. Definition of `Atom` type has `Box<Grounded>` inside and it cannot by default be kept in static context, thus finally I made a hack with parsing text to represent the operation type. I thought to replace it by normal atom instantiation when I implement `Sync` on `Atom` or something similar to make Rust compiler happy. This work is worth to do but I would make it in separate PR, because I believe it should touch a lot of code.

lib/src/metta/runner/mod.rs
luketpeterson: Ok.  I thought it would be a nice convenience to give users a 1-line method to execute code.  But I see your desire to have syntax-independence at the Runner level is more important.

Removed in d26b225

lib/src/metta/runner/mod.rs
luketpeterson: Ok, after the discussion last night, I removed Metta::parse_all_atoms and Metta::parse_one_atom in 28e464e

Aesthetically I feel like this is not great because parsing atoms requires a tokenizer, which is initialized through a runner, and borrowing a runner's tokenizer is very ugly.

So a conceptually straightforward operation, (parsing MeTTa code into an Atom) the following ugly Rust code:
```Rust
SExprParser::new("(Cons 1 Nil)").parse(&*metta.tokenizer().borrow())?.unwrap()
```
This would be a lot nicer if `Metta::tokenizer` could return an ordinary reference.

lib/src/metta/runner/mod.rs
luketpeterson: API removed in https://github.com/trueagi-io/hyperon-experimental/commit/28e464e3e807741a2309278411dcf49fd8fb84d9  See comments above.

lib/src/metta/runner/mod.rs
vsbogd: Thanks, Luke

lib/examples/sorted_list.rs
vsbogd, line 1: Luke, I have added PR to your branch https://github.com/luketpeterson/hyperon-experimental/pull/1 to fix the test and remove using SExprParser here.

lib/src/metta/interpreter2.rs
```diff
@@ -468,7 +469,7 @@ mod tests {
     fn interpret_atom_evaluate_pure_expression_variable_name_conflict() {
         let space = space("(= (foo ($W)) True)");
         let result = interpret_atom(&space, atom("(eval (foo $W))", bind!{}));
-        assert_eq!(result[0].0, sym!("True"));
+        assert_eq!(result[0].0, metta_atom("True"));
```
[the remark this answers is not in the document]
luketpeterson: The previous test harness Tokenizer was a strange half-initialized Tokenizer, with some of the of the stdlib tokens present but not all of them.

I don’t see the value in that half-initialized tokenizer, and some of the other tests (types.rs in particular) depend on some of the stdlib tokens.  So I changed the test harness to use the tokenizer from a fully-initialized runner.  However, that means `True` and `False` became defined tokens.

If resolving to a symbol (instead of a GroundedAtom) is important for this test, the following code passes with either Tokenizer:

        let space = space("(= (foo ($W)) Yes)");
        let result = interpret_atom(&space, atom("(eval (foo $W))", bind!{}));
        assert_eq!(result[0].0, sym!("Yes"));

Alternatively, the 8d15e25 changes the types.rs tests to use their own version of the convenience functions, and changes the common test setup functions to use an empty Tokenizer.

lib/src/metta/interpreter2.rs
```diff
@@ -402,6 +402,7 @@ fn cons(bindings: Bindings, head: Atom, tail: ExpressionAtom) -> Vec<Interpreted
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::common::test_utils::{metta_atom, metta_space};
```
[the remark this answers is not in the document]
luketpeterson: It is one thing to say the interpreter should not depend on the parser.  As you know, it doesn't.  It's another thing to say its tests should not.

Consider the transformation of rewriting the `metta_turing_machine` test without the parser.  The code would bloat with a lot of Rust boilerplate and become a lot less readable.  And I don't see the benefit aside from potentially isolating the interpreter tests from parser bugs.

Can you explain what you hope to accomplish with this change?

luketpeterson:
What additional changes are required to make this ready for merge?

lib/src/metta/interpreter2.rs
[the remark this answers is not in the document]
vsbogd: Yes, I agree, sorry missed that on a first review.

vsbogd (approved):

vsbogd:
Luke, thank you so much for addressing all of comments, I happy to merge it.

PR #467, opened 2023-10-23
Changing init.metta to run in the runner's top-level Space, rather th…

luketpeterson:
Changing init.metta to run in the runner's top-level Space, rather than as its own MeTTa module.

### DIFF
--- a/lib/src/metta/runner/mod.rs
+++ b/lib/src/metta/runner/mod.rs
@@ -6,7 +6,6 @@ use super::space::*;
 use super::text::{Tokenizer, SExprParser};
 use super::types::validate_atom;
 
-use core::borrow::Borrow;
 use std::rc::Rc;
 use std::path::{Path, PathBuf};
 use std::collections::HashMap;
@@ -243,8 +242,8 @@ impl Metta {
     }
 
     /// Runs the MeTTa code in the specified file
-    pub fn run_from_file<P: Borrow<Path>>(&self, program_path: P) -> Result<Vec<Vec<Atom>>, String> {
-        let path = program_path.borrow();
+    pub fn run_from_file<P: AsRef<Path>>(&self, path: P) -> Result<Vec<Vec<Atom>>, String> {
+        let path = path.as_ref();
         let program = std::fs::read_to_string(path)
             .map_err(|err| format!("Could not read file, path: {}, error: {}", path.display(), err))?;
         self.run(SExprParser::new(program.as_str()))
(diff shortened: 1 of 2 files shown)


Necr0x0Der (approved):
I'm not sure for what purpose init.metta module should be used. If it's ok for now to use it as an initialization script instead of a module, I prefer this approach. We can change this later if needed especially after developing a module management subsystem.

lib/src/metta/runner/mod.rs
```diff
@@ -241,6 +241,14 @@ impl Metta {
         self.0.settings.borrow().get(key.into()).map(|a| a.to_string())
     }
 
+    /// Runs the MeTTa code in the specified file
+    pub fn run_from_file<P: AsRef<Path>>(&self, path: P) -> Result<Vec<Vec<Atom>>, String> {
```
vsbogd, line 245: I would suggest making this method private. Each public method increases the probability of the compatibility issues which happens on the our side. I would introduce an `SExprParser` constructor which is initialized by the instance of the `Path`.
```suggestion
    fn run_from_file<P: AsRef<Path>>(&self, path: P) -> Result<Vec<Vec<Atom>>, String> {
```

vsbogd: One of the ideas behind `SExprParser` is that programmer can implement the parser for another MeTTa syntax and use it with the runner. It requires additional work indeed, for example we need to abstract the parser API and remove places where `SExprParser` instances are created inside the runner.

luketpeterson: That makes sense.

I thought about adding a convenience to SExprParser to ingest a file, but the way the parser borrows its input makes that interface a little tricky.  So ultimately I just decided to remove the `Metta` method altogether.

Possibly in the future we may want to make "Parser" into a trait where we can have a parser that can load code from a file (and start returning results without loading it all into memory at once), and another parser for lambda syntax, etc.  But for now, these are all very far down the priority list.

For now, I've just removed the public API in b2d0314.

vsbogd: > Possibly in the future we may want to make "Parser" into a trait where we can have a parser that can load code from a file (and start returning results without loading it all into memory at once), and another parser for lambda syntax, etc. But for now, these are all very far down the priority list.

Yes, it should be probably not even `Parser` but a trait which represents a stream of atoms. Thus the case when user of the API provides just a list of atoms can also be represented by this trait.

vsbogd (approved):

PR #468, opened 2023-10-23
Adding "module-space" MeTTa stdlib op, to access the space of an impo…

luketpeterson:
Adding "module-space" MeTTa stdlib op, to access the space of an imported module

### DIFF
--- a/lib/src/metta/runner/mod.rs
+++ b/lib/src/metta/runner/mod.rs
@@ -10,6 +10,7 @@ use std::rc::Rc;
 use std::path::{Path, PathBuf};
 use std::collections::HashMap;
 use std::sync::Arc;
+use core::borrow::Borrow;
 
 mod environment;
 pub use environment::{Environment, EnvBuilder};
@@ -221,7 +222,12 @@ impl Metta {
             .chain(self.0.environment.extra_include_paths())
     }
 
-    pub fn modules(&self) -> &Shared<HashMap<PathBuf, DynSpace>> {
+    /// Returns the space of a given loaded module
+    pub fn get_module_space<P: Borrow<Path>>(&self, module: P) -> Option<DynSpace> {
+        self.0.modules.borrow().get(module.borrow()).map(|space| space.clone())
+    }
+
+    pub(crate) fn modules(&self) -> &Shared<HashMap<PathBuf, DynSpace>> {
         &self.0.modules
     }
 
--- a/lib/src/metta/runner/stdlib.rs
+++ b/lib/src/metta/runner/stdlib.rs
@@ -144,6 +144,44 @@ impl Grounded for ImportOp {
 }
 
 #[derive(Clone, PartialEq, Debug)]
+pub struct ModuleSpaceOp {
+    metta: Metta,
+}
+
+impl ModuleSpaceOp {
+    pub fn new(metta: Metta) -> Self {
+        Self{metta}
+    }
+}
+
+impl Display for ModuleSpaceOp {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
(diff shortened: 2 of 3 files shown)


luketpeterson:
> I would not hurry with this solution, since we need to discuss and design a library manager subsystem in much more detail covering various existing issues. However, I'm ok with it as an intermediate fix of this particular piece.

If we merge https://github.com/trueagi-io/hyperon-experimental/pull/467 then this change isn't needed to make the tests pass.

I am just bothered by the way the `f1_imports.metta` test is checking for specific atoms in a certain order in the list returned by `get-atoms`.  This `stdlib` operation was a way to eliminate one of the tests that I found disagreeable.

If there is a greater list of requirements and desiderata for an import / library system, I would like to see it, because I spent a lot of time thinking through library management in the design of the Environment object.

luketpeterson:
> I would not hurry with this solution, since we need to discuss and design a library manager subsystem in much more detail covering various existing issues. However, I'm ok with it as an intermediate fix of this particular piece.

If we merge https://github.com/trueagi-io/hyperon-experimental/pull/467 then this change isn't needed to make the tests pass.

I am just bothered by the way the `f1_imports.metta` test is checking for specific atoms in a certain order in the list returned by `get-atoms`.  This `stdlib` operation was a way to eliminate one of the tests that I found disagreeable.

If there is a greater list of requirements and desiderata for an import / library system, I would like to see it, because I spent a lot of time thinking through library management in the design of the Environment object.

luketpeterson:
Ah!  I just saw that you filed https://github.com/trueagi-io/hyperon-experimental/issues/470.  Thanks!

Necr0x0Der:
> I am just bothered by the way the f1_imports.metta test is checking for specific atoms in a certain order

f1_imports.metta fails on the first line. At this moment, there should be only one atom in accordance with the previous. Thus, this particular test didn't rely on the atom order at all. The next failed test
```
!(assertEqual
  (let* (($x (collapse (get-atoms &m)))
         ($y (car-atom $x)))
        (get-type $y))
  (get-type &self))
```
does rely on the atom order, which definitely doesn't work for `match`, and underspecified for `get-atoms`. The test verifies some concrete behavior that the module loaded not in `&self` contains stdlib. I agree that it is not nice to rely on this assumption here, but the test can be fixed in different way - both more conservative and more fundamental. For example, the test can be moved to the imported script. And it's ok to have an intermediate solution. Loading init.metta to `&self` may make sense depending on what content we want in init.metta. I'm not precisely sure about the latter, and it worths a separate analysis. In any case, changing how init.metta is loaded doesn't improve f1 test by itself. I don't like changing it to the use of dedicated `stdlib` operation, because this test will become much narrower, and will test only for stdlib (and not absence of something else). In any case, the whole situation with imports should be improved, and this should be fixed more fundamentally within the design of the module manager. For example, if we choose the path to insert the module loader atom instead of particular imported spaces / modules, then the tests should cover the supposed behavior of this solution in all detail.

Necr0x0Der:
As per #470 there will hopefully be considerable changes in the module manager. I'm not against `module-space` operation in general (it might be different in the future, but this is also ok), however, I don't like much the change of the test. The main intention of this test was not to test that `stdlib` is working (there are many tests, which implicitly test this), but to explain the metta coders why the seemingly empty space is not actually empty. These scripts are not just tests, but also a sort of the only tutorial we have. This may change in the future, but I really want to keep this explanation of non-emptiness of `&self`. Thus, I'm not in favor of this workaround of `f1` issue, and thus there is not much need for introducing `module-space` in this version of the module manager (although I'm ok with introducing `module-space` by itself).

luketpeterson:
Closing as the module system changes have diverged from this PR

Issue #465, opened 2023-10-20
Errors inside MeTTa runner C bindings shouldn't panic


noskill:
currently repl fails with error "Unexpected end of expression" when  if multi-line expression is pasted

example  
```
metta>         (= (reverse $Xs $Ys)
thread '<unnamed>' panicked at src/metta.rs:936:27:
Returning errors from C API is not implemented yet: "Unexpected end of expression"
```

noskill:
interpreter just need to count parenthesises  and interpret only when the expression is closed

luketpeterson:
Parse errors are now piped through C correctly when interacting directly with the parser.  This issue is that the runner functions (e.g. `metta_run`) just panic on those errors.

I'd like to keep this issue open and close the other one, as this issue actually deals with runner errors, whereas the other issue was for parser errors.

luketpeterson:
NOTE: This fix should be implemented for both the `metta_t` object (`metta_run`, `metta_evaluate_atom`, and `metta_load_module`) as well as the `runner_state_t` object (`runner_state_step`)

Necr0x0Der:
> I'd like to keep this issue open and close the other one

OK, makes sense

luketpeterson:
With the https://github.com/trueagi-io/hyperon-experimental/pull/471/commits changes, the errors are catchable.

luketpeterson:
I believe this should be done.  Assigning to @noskill to verify.

PR #471, opened 2023-10-24
Runner err accessors

luketpeterson:
Plumbing errors from Runner & RunnerState interface through C & Python, to address https://github.com/trueagi-io/hyperon-experimental/issues/465

### DIFF
--- a/lib/src/metta/runner/mod.rs
+++ b/lib/src/metta/runner/mod.rs
@@ -332,16 +332,12 @@ impl<'m, 'i> RunnerState<'m, 'i> {
             } else {
 
                 //This interpreter is finished, process the results
-                match interpreter_state.into_result() {
-                    Err(msg) => return Err(msg),
-                    Ok(result) => {
-                        let error = result.iter().any(|atom| atom_is_error(atom));
-                        self.results.push(result);
-                        if error {
-                            self.mode = MettaRunnerMode::TERMINATE;
-                            return Ok(());
-                        }
-                    }
+                let result = interpreter_state.into_result().unwrap();
+                let error = result.iter().any(|atom| atom_is_error(atom));
+                self.results.push(result);
+                if error {
+                    self.mode = MettaRunnerMode::TERMINATE;
+                    return Ok(());
                 }
             }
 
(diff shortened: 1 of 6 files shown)


lib/src/metta/runner/mod.rs
```diff
@@ -345,7 +345,13 @@ impl<'a> RunnerState<'a> {
 
             // Get the next atom, and start a new intperpreter
             let next_atom = if let Some(parser) = self.parser.as_mut() {
-                parser.parse(&self.metta.0.tokenizer.borrow())?
+                match parser.parse(&self.metta.0.tokenizer.borrow()) {
+                    Ok(atom) => atom,
+                    Err(err) => {
+                        self.mode = MettaRunnerMode::TERMINATE;
```
vsbogd, line 351: Do we really need setting mode to `TERMINATE`? Looks like previous line should work nice.

luketpeterson: Do you see a reason that terminating the RunnerState on a parse error would be bad?

The reason I did it this way was because a parse error essentially results in a crash for that RunnerState (but not for the whole runner).  In other words, I don't see a situation where we'd want to resume execution in the same runner state after some MeTTa code failed to parse.  Potentially that could lead to very hard-to-debug situations where some code was silently skipped causing problems downstream.

In Rust it's a non-issue because Rust passes back `Result` types, which force the caller to consider errors.  But in the C API, it seemed likely code like this would be common:
```C
    while (!runner_state_is_complete(&runner_state))
        runner_state_step(&runner_state);
```
And I didn't want to create an API that would lead to infinite loop if it was slightly misused, and I didn't see a reason not to terminate the RunnerState on a syntax error.

vsbogd: Thanks for explaining this. I asked mainly because there is another line above with error processing which doesn't set `TERMINATE` mode on `RunnerState`, and I was curious what is a difference:
```rs
                match interpreter_state.into_result() {
                    Err(msg) => return Err(msg),
```
Should we change it to terminate runner as well? Probably we will never got error in this snipped though.

vsbogd (approved):

noskill (dismissed):

lib/src/metta/runner/mod.rs
luketpeterson: Good point.  That case should never be hit, but I simplified the code to make that clear in 53fd2ab.

Thanks!

vsbogd (approved):

Issue #470, opened 2023-10-23
Tracking issue for metta libraries / modules / packages


Necr0x0Der:
MeTTa libraries and extensions are to be designed and standartized. There are several aspects to be fleshed out:

=== FEATURE LIST ===

== INSTALL == 
1a. While REPL introduced environments and a standard place to keep MeTTa modules, some features can be added: a way to install custom modules in a standard location (e.g. via repl, or another executable file, or some metta library like `setuptools`), instructions / examples of using `METTA_LIBRARY` or `METTA_PATH` environment variables, installing a library from git links.  --- required for alpha-release

1b. Libraries / modules with submodules as well as imports of modules from other modules should be well supported.  --- necessary for alpha-release

== TOKENS == 
2a. MeTTa extensions (which include some grounded atoms - not only pure .metta files) will be the most typical packages for some time for introducing both AI related stuff (neural networks, etc.) as well as I/O features and such. A minimal requirement here is for `extend-py!` to behave similarly as `import!` in the context of (1), or for .metta library files to be able to use `extend-py!` (e.g. instead of calling `(extend-py! llm-gate)` we can call `(import! llm-gate)`, which will call `extend-py!` by itself).  --- required for alpha-release

2b. Tokens introduced in module being imported should be accessible in the importer module. This may (or may not) be true both for extensions and for pure metta scripts, which can introduce their own tokens. It's a general question what we do with importing tokens.  --- useful for alpha-release

2c. Extensions with dynamically linked libraries (created in c/c++, rust, ...) should be installable and loadable in the same way as pure metta libraries and python extensions. It should be standartized either via using some grounded symbols in .metta files, or, say, if `import!` is called not on a metta file, but on a folder, there should be `__init__.metta` or `config.metta` or whatever, which may refer to foreign language code / binaries as well. --- useful for alpha-release

== EVALUATION == 
3a. It should be possible to evaluate expressions in the context of imported module, so modules could contain callable functions.

3b. Namespaces and selective imports of symbols / tokens can be supported.

MeTTa libraries and extensions are to be designed and standartized. There are several aspects to be fleshed out:

=== FEATURE LIST ===

== INSTALL == 
[...]

Issue #242, opened 2023-02-07
Execution of expressions with variables as operations


Necr0x0Der:
The following code:
```
(= (inc $x) (+ $x 1))
(= (dec $x) (- $x 1))
!($x 2)
```
returns `[1, 3]`. This is fun, and doesn't contradict to evaluation via equality queries. However, I'd prefer to
1) be able to specify expressions like `($x $y)` without automatic reduction;
2) force to explicitly match functional symbols if needed, e.g.
```
!(match &self (= ($func 2) $r) $r)
!(match &self (= ($func $x) $r) ($func 2))
```
both work.
Thus, I propose to consider variables as non-operational if this doesn't require considerable refactoring and breaks no tests.

ngeiswei:
I guess both semantics (operational and non-operational variables) have upsides and downsides.  The downside of the operational variable semantics is that it is more prone to infinite recursion.  Its upside is that it is more elegant as far as defining evaluation as a form of matching.

Maybe a pragma could be introduced to switch between these two semantics?

Or, maybe even better, that behavior could be more generally controlled by overwriting some mini-metta code (though I don't know precisely how, if at all possible)?

ngeiswei:
An inelegance of the operational variable semantics, as currently implemented, is that built-in functions are treated differently than metta functions.  For instance `($x $y $z $w)` unifies with `(if ...)` but not with `(let ...)`.

vsbogd:
> An inelegance of the operational variable semantics, as currently implemented, is that built-in functions are treated differently than metta functions. For instance ($x $y $z $w) unifies with (if ...) but not with (let ...).

It is more inelegance of the grounded functions implementation. As grounded functions are not parts of the atomspace, they cannot be matched by `($x ...)` patterns. It is one of the things we can fix by fixing grounded functions representation.

vsbogd:
> Or, maybe even better, that behavior could be more generally controlled by overwriting some mini-metta code (though I don't know precisely how, if at all possible)?

Yes, one of the ways to make it flexible is allow to redefine `eval` function to the user. One of the ideas is to provide a way to implement `eval` via `match`.

noskill:
why it is blocker? What it blocks? @Necr0x0Der

vsbogd:
> why it is blocker? What it blocks? @Necr0x0Der

[...]

PR #472, opened 2023-10-24
Minimal interpreter: adding function/return and fix most of the tests

vsbogd:
Improving semantics of minimal MeTTa `chain` to make code more predictable. I didn't change the documentation yet because I am going to experiment with `eval/chain/return` further and then finalize the semantics. After adding `function/return` I managed to write MeTTa interpreter in minimal MeTTa which now passes almost all of the tests except:
- Minecraft related tests
- one test in minimal MeTTa Rust test pack
- `f1_import` test

List of changes:
Implement `function` and `return` symbols to allow coding function call flows. Make `chain` one-step operation, i.e. `chain` makes one step and inserts the result into a provided template. One can make `chain` work in a loop by: 
- using nested `chain`, then `chain` waits while nested `chain` returns result
- using nested `function`, then `chain` waits for a `function` to return result
Thus nested `chain` + `function` expressions represent the call stack of the program.

Import `pragma!`, `bind!`, space operations, state operations from stdlib to stdlib of minimal MeTTa.
Implement `case`, `quote/unquote`, `id`, `car-atom`, `cdr-atom`, `or`, `and`, `fold`, `filter` and `map` in minimal MeTTa.
Change `get-type` implementation to be compatible with both Rust and minimal MeTTa interpreters.
Fix type casting issue when type errors were returned for non-deterministic types when one of the types was not compatible with required type.
Fix `test_examples.py` to be compatible with minimal MeTTa standard library.
Adapt "returning unit" from grounded functions fix from `main` branch.

### DIFF
--- a/c/src/metta.rs
+++ b/c/src/metta.rs
@@ -554,13 +554,15 @@ pub extern "C" fn atom_error_message(atom: *const atom_ref_t, buf: *mut c_char, 
 ///
 #[no_mangle] pub extern "C" fn ATOM_TYPE_GROUNDED_SPACE() -> atom_t { rust_type_atom::<DynSpace>().into() }
 
-/// @brief Creates a Symbol atom for the special MeTTa symbol used to indicate empty results in
-/// case expressions.
+/// @brief Creates a Symbol atom for the special MeTTa symbol used to indicate empty results
+/// returned by function.
 /// @ingroup metta_language_group
 /// @return  The `atom_t` representing the Void atom
 /// @note The returned `atom_t` must be freed with `atom_free()`
 ///
-#[no_mangle] pub extern "C" fn VOID_SYMBOL() -> atom_t { hyperon::metta::VOID_SYMBOL.into() }
+#[no_mangle] pub extern "C" fn EMPTY_SYMBOL() -> atom_t {
+    hyperon::metta::EMPTY_SYMBOL.into()
+}
 
 /// @brief Checks whether Atom `atom` has Type `typ` in context of `space`
 /// @ingroup metta_language_group
@@ -705,7 +707,7 @@ pub extern "C" fn step_has_next(step: *const step_result_t) -> bool {
     step.has_next()
 }
 
-/// @brief Consumes a `step_result_t` and provides the ultimate outcome of a MeTTa interpreter session 
+/// @brief Consumes a `step_result_t` and provides the ultimate outcome of a MeTTa interpreter session
 /// @ingroup interpreter_group
 /// @param[in]  step  A pointer to a `step_result_t` to render
 /// @param[in]  callback  A function that will be called to provide a vector of all atoms resulting from the interpreter session
--- a/docs/minimal-metta.md
+++ b/docs/minimal-metta.md
@@ -32,7 +32,7 @@ allowed developing the first stable version with less effort (see `eval` and
 `Return`). If an instruction returns the atom which is not from the minimal set
 it is not interpreted further and returned as a part of the final result.
 
-## Error/Empty/NotReducible/Void
+## Error/Empty/NotReducible/()
 
 There are atoms which can be returned to designate a special situation in a code:
 - `(Error <atom> <message>)` means the interpretation is finished with error;
@@ -46,8 +46,8 @@ There are atoms which can be returned to designate a special situation in a code
   which returns `NotReducible` explicitly; this atom is introduced to separate
   the situations when atom should be returned "as is" from `Empty` when atom
   should be removed from results;
-- `Void` is a unit result which is mainly used by functions with side effects
-  which has no meaningful value to return.
+- Empty expression `()` is a unit result which is mainly used by functions with
+  side effects which has no meaningful value to return.
 
 These atoms are not interpreted further as they are not a part of the minimal
 set of instructions.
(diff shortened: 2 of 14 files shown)


Necr0x0Der:
I'd like to migrate to minimal metta asap, so I'm worried only about tests and performance, but not internals. Thus, I'm happy to approve this PR. A few remarks, which should not be considered within this PR:

* I'm definitely ok with `(Error Z BadType)` changing to `(Error (Cons Z Nil) BadType)`, which is more informative. But I'd either turn the whole assert off with a proper comment, or make the test work in both cases, e.g. checking for `Error` or `BadType` only if it is not cumbersome. Maybe, we can introduce `assertMatchable` and check for `(Error $_ BadType)` if it makes sense.

* I'm ok with moving test_minecraft and test_minology from obligatory tests (to integration tests?). But it might be useful to figure out the problem with test_minecrat and see if it is important (possibly, add a clean unit test if appropriate). The problem with minology is that `BadType` is returned among the correct answers. It's definitely the change in semantics. Previously, BadType was returned when there were no valid alternatives only. Now, it is returned among valid alternatives as well. This makes some sense, but may make the use of non-deterministic types cumbersome. This requires more detailed analysis, which can be postponed, but not ignored. Thus, moving minology to integration tests is ok.

* I'm ok with modifying `test_frog_reasoning` test, but am a little bit sad that the first element of tuples gains even more uniquness.

* I see a lot of code in stdlib.metta, and start wondering if we will introduce local caches in the future (e.g. like `__pycache__`) to load it and other modules faster. Do you think it worths mentioning in #470 or is it a very separate issue?

* stdlib.metta contains a lot of functions with a little bit questionable names, e.g. `filter`, `map`, etc., which are too typical and can cause some name conflicts, e.g. we may want to have these functions on `List` type instead of atoms. But this can be discussed later.

Necr0x0Der (approved):

vsbogd:
> I'm definitely ok with (Error Z BadType) changing to (Error (Cons Z Nil) BadType), which is more informative. But I'd either turn the whole assert off with a proper comment, or make the test work in both cases, e.g. checking for Error or BadType only if it is not cumbersome. Maybe, we can introduce assertMatchable and check for (Error $_ BadType) if it makes sense.

Do you mean turn it off for both minimal and old interpreter? I would prefer doing this instead of adding `assertMatchable` for this only case. 

> The problem with minology is that BadType is returned among the correct answers. It's definitely the change in semantics.

Yes, it is one more place for the type-checking logic to fix in minimal interpreter. I will fix it.

> I'm ok with modifying test_frog_reasoning test, but am a little bit sad that the first element of tuples gains even more uniquness.

Not sure I understand you properly. The problem is the same which is described in #242. As standard library has more functions now, the code of the test matches them. It would happen even with old interpreter if we add new functions into stdlib.

> I see a lot of code in stdlib.metta, and start wondering if we will introduce local caches in the future (e.g. like __pycache__) to load it and other modules faster. 

Do you mean we need to pre-compile MeTTa libraries? It makes sense but I would make it a separate issue from a package management because its complexity is in precompilation not in aligning it with packaging system.

> stdlib.metta contains a lot of functions with a little bit questionable names, e.g. filter, map, etc., which are too typical and can cause some name conflicts, e.g. we may want to have these functions on List type instead of atoms. But this can be discussed later.

Since minimal MeTTa I am writing more code in MeTTa and more frequently facing name conflicts. I can assign unique names for the functions in standard library for sure but namespaces would be really useful here (I believe they are already mentioned in #470)

Necr0x0Der:
> Do you mean turn it off for both minimal and old interpreter? I would prefer doing this instead of adding assertMatchable for this only case.

Yes, to avoid switching back and forth between two versions of the test until we completely migrate to Minimal MeTTa, when we can turn it on back.

> It would happen even with old interpreter if we add new functions into stdlib.

OK, I see.

> Do you mean we need to pre-compile MeTTa libraries?

Pre-compilation is the next step. At first, we can just store parsed spaces in files (for which they should be serializable). But yes, it seems like a separate branch of development. And maybe it (saving on parsing) is really a minor thing, and it will start to be interesting only after pre-compilation, which is really a very separate story.

>  but namespaces would be really useful here

I supposed that stuff from stdlib will not be in a separate namespace to avoid, e.g. `stdlib::if`. But maybe we can import such stuff as `if` directly to the main script namespace, while `map` can remain in stdlib namespace. But I mean that `map` for `List` can also be in stdlib. In any case, I pointed out to the question of naming such widely used but differently implemented functions (e.g. `map-atom` similar to `car-atom`?) just to not forget discuss it later.

Necr0x0Der:
I'm ok with renaming `map` -> `map-atom` right away, but it makes the further review of naming less relevant :)

vsbogd:
> Yes, to avoid switching back and forth between two versions of the test until we completely migrate to Minimal MeTTa, when we can turn it on back.

Fixed in https://github.com/trueagi-io/hyperon-experimental/pull/472/commits/1646df9236bf71e1e65f43a80c05a40d16917990

> I supposed that stuff from stdlib will not be in a separate namespace to avoid

Yes, ok, probably namespaces are not so useful for the functions we are going to export from the stdlib, but we can use them to make functions private, in particular some implementation specific stdlib functions. But yes maybe it is not so important, we always can add prefix stdlib for them.

Necr0x0Der (approved):

Issue #475, opened 2023-10-26
Not create .config/metta directory by default when API is used


vsbogd:
MeTTa runner creates `.config/metta` directory by default and fills it by `init.metta` script when it is started for the first time even when it is called through API. But this behavior is not appropriate by the following reasons:
- it leaves environment dirty and behavior not reproducible - next time user starts MeTTa runner it expects it works the same way as at the first time, but in fact the logic will be different because some initialization files are already in filesystem. If we will change the content of the initialization files in the future the logic for a new version of MeTTa runner will be different and user is not aware of it because he or she don't know about files created
- filesystem can be non-writable - under many conditions writing files into filesystem can fail: lack of permissions or read-only disk
- it is not expected that API usage makes any changes in a filesystem without explicit asking to do it

User can expect that running command line tools creates a configuration folder and keeps some info in it, but even this is not really necessary (for example Python doesn't do it). If user wants to change an environment he can always do it explicitly. But I would suggest disabling creating `.config` directory by default when API is used.

vsbogd:
The quickest fix I can suggest is change `no_cfg_dir` value from `false` to `true` by default.

Necr0x0Der:
I don't have a strong opinion on automatic creation of `.config` in general (although I do agree with some points above), but I have a few remarks:
* MeTTa should be installed somewhere (even its API part including .so files, Python library, etc.);
* we may want to put there some command line tools as well (in particular, for creating environments, etc.);
* currently, `.config` has a very useful `modules` subfolder, which is available via API as well; if `.config` is not created, we should put this folder somewhere else;
* I agree that creating and loading `init.metta` is unexpected by API users, and thus it is kind of undesirable; however, I'd like the behavior to be identical for REPL and API by default, so it's maybe not the question of having `.config` or not, but not creating `metta.init` by default for both REPL and API. However, if REPL prefers to have `init.metta` it might be better just to explicitly separate REPL and API environments;
* besides `modules` (which are or will be needed by default) and `init.metta` (which may not be needed by default for API), other stuff in `.config` is relevant to REPL only (however, other command line tools may utilize this folder in the future as well). Apparently, REPL needs some config files, but maybe we just need to separate what is needed by REPL, and what is needed by core API keeping in mind that we do want to have "one metta" (parameterized by environments, though).

vsbogd:
> currently, .config has a very useful modules subfolder, which is available via API as well; if .config is not created, we should put this folder somewhere else;

[...]

PR #478, opened 2023-10-30
Creating default Environment config files only when explicitly direct…

luketpeterson:
Creating default Environment config files only when explicitly directed.

This addresses https://github.com/trueagi-io/hyperon-experimental/issues/475

### DIFF
--- a/lib/src/metta/runner/environment.rs
+++ b/lib/src/metta/runner/environment.rs
@@ -140,6 +140,9 @@ impl EnvBuilder {
     /// Configures the Environment not to load nor create any config files
     pub fn set_no_config_dir(mut self) -> Self {
         self.no_cfg_dir = true;
+        if self.create_cfg_dir {
+            panic!("Fatal Error: set_no_config_dir is incompatible with create_config_dir");
+        }
         if self.env.config_dir.is_some() {
             panic!("Fatal Error: set_config_dir is incompatible with set_no_config_dir");
         }
--- a/lib/src/metta/runner/environment.rs
+++ b/lib/src/metta/runner/environment.rs
@@ -78,6 +78,7 @@ impl Environment {
 pub struct EnvBuilder {
     env: Environment,
     no_cfg_dir: bool,
+    create_cfg_dir: bool,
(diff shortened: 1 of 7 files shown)


vsbogd:
Luke, do you think there is a case when environment should create directory? In #475 I suggested that `modules` directory should be created by a tool which installs libraries,  `metta.init` script could be created manually (depends on its purpose). And in this case `Environment` could only provide the standard paths for this scripts and paths can be created by the code which uses `Environment` instances.

luketpeterson:
> Luke, do you think there is a case when environment should create directory?

As we discussed on the call, my reasoning for keeping the default config creation inside the library is:

* The Environment needs to be read by the library.  So the logic to read it must be within the library
* The format of the Environment should be centralized in one place, so logic to update and modify the Environment belongs near logic to read it.

Also, it isn't big, so size isn't a factor in bloating the core library.

lib/src/metta/runner/environment.rs
```diff
@@ -124,6 +126,17 @@ impl EnvBuilder {
...
+    pub fn create_config_dir(mut self) -> Self {
+        self.create_cfg_dir = true;
+        if self.no_cfg_dir {
+            panic!("Fatal Error: create_config_dir is incompatible with set_no_config_dir");
+        }
+        self
+    }
+
     /// Configures the Environment not to load nor create any config files
     pub fn set_no_config_dir(mut self) -> Self {
```
vsbogd, line 141: I would also check that `self.no_cfg_dir` value doesn't conflict with `self.create_cfg_dir` value.

luketpeterson: Added in [168231b](https://github.com/trueagi-io/hyperon-experimental/pull/478/commits/168231b61bdec8f1c80f368eceb86bd6669ea794)

luketpeterson:
What additional changes are required to make this ready for merge?

vsbogd (approved):

vsbogd:
Thanks Luke for fixing this!

PR #480, opened 2023-11-01
change arg type to atom in print and trace

noskill:
with this change one can print what actually is during expression evaluation 

this program will print now *(woman $Y)* vs printing *False* without these changes
```
(= (woman $X) False)
(= (sisters (woman $X) (woman $Y)) (trace! (woman $Y) True))
!(let $W (sisters $X $Y) $W)
 ```

### DIFF
--- a/lib/src/metta/runner/stdlib.rs
+++ b/lib/src/metta/runner/stdlib.rs
@@ -758,7 +758,7 @@ impl Display for PrintlnOp {
 
 impl Grounded for PrintlnOp {
     fn type_(&self) -> Atom {
-        Atom::expr([ARROW_SYMBOL, ATOM_TYPE_UNDEFINED, UNIT_TYPE()])
+        Atom::expr([ARROW_SYMBOL, ATOM_TYPE_ATOM, UNIT_TYPE()])
     }
 
     fn execute(&self, args: &[Atom]) -> Result<Vec<Atom>, ExecError> {
@@ -816,7 +816,7 @@ impl Display for TraceOp {
 
 impl Grounded for TraceOp {
     fn type_(&self) -> Atom {
-        Atom::expr([ARROW_SYMBOL, ATOM_TYPE_UNDEFINED, Atom::var("a"), Atom::var("a")])
+        Atom::expr([ARROW_SYMBOL, ATOM_TYPE_ATOM, Atom::var("a"), Atom::var("a")])
     }
 
     fn execute(&self, args: &[Atom]) -> Result<Vec<Atom>, ExecError> {

### CODE REVIEW

**Comment by Necr0x0Der** [comment]
I considered this before, but decided not to change the argument type of `println!` to `Atom`, because it is not too usable:
* if you want to print the result of some calculations, they will be reduced in any case.
* if you want to avoid wrapping `println!` into `let` each time you want to print the results of calculations, the type of `println!` argument **should not** be `Atom`. And this is quite a common use cae.
For example, if the argument type is `Atom`, and we want to print `get-atoms` result, `! (println! (get-atoms &self))` will print `(get-atoms &self)`, which is useless, while `! (let $atoms (get-atoms &self) (println! $atoms))` will still output the reduced atoms. Thus, `Atom`-type argument makes only sense to print the exact provided expression. This is a less frequent case than printing the result of evaluation. If one wants to print an unreduced expression, it can be quoted, e.g.
```
(: TRACE (-> Atom Atom))
(= (woman $X) False)
(= (sisters (woman $X) (woman $Y)) (trace! (TRACE (woman $Y)) True))
!(let $W (sisters $X $Y) $W)
```
I'm not sure, maybe printing atoms as is in case of `trace!` can be more demanded. I don't have a strong opinion here. But it looks questionable for `println!`. @vsbogd , @luketpeterson , do you have any opinion?

vsbogd:
I agree with @Necr0x0Der arguments about `println!` it is probably not convenient to use `let` each time and usually you need to reduce atom before printing.

Regarding `trace!` I would ask @ngeiswei opinion because he introduced `trace!` and probably most active user of it. I almost always use `trace!` to print atom without reducing it, and I use quoting to reach it. If it is more frequent use-case then I think making first argument of `trace!` atom may help.

noskill:
Quotation solves problem for me, i tried quote from standard library before doing PR but it didn't work

Necr0x0Der:
But I'd also like to hear @ngeiswei opinion. Maybe, `Atom` in `trace!` can indeed be convenient.

ngeiswei:
It looks like I'm always using `trace!` with fully reduced first argument.  And one can always wrap a `let` around `trace!` to reduce its first argument if needed.  So I guess it wouldn't hurt to have its type be `Atom`.  But Just keep in mind that it is not how Idris or Haskell `trace` works.

Maybe we should consider introducing quotes instead of having to decide at the type level how a function should behave.

Necr0x0Der:
OK, if @ngeiswei has no preferences for `Atom` type, then let's keep it as is. Some official quoting operation in stdlib would be useful for sure.

PR #485, opened 2023-11-02
Implement boolean and, or, and not operations in Rust stdlib

vsbogd:
It was triggered by inability to write unit test for the #483 
I added this unit test as well.


luketpeterson (approved):
Looks good

Issue #489, opened 2023-11-07
Evaluating `<=` or `>=` raise error


ngeiswei:
## What is the problem?

Evaluating `<=` or `>=` raise error.

## How to reproduce the problem?

Evaluate

```
!(<= 0 1)
```

## What is output?

```
[(Error (<= 0.1 1.1) <= expects two number arguments)]
```

## What should normally be output?

```
[True]
```

## Remarks

Operator `>=` has the same problem, but `<` or `>` work normally.

Necr0x0Der:
Most likely, the issue is that `<=` and `>=` got implemented in Rust stdlib, but not in Python.

Necr0x0Der:
It might be a good first issue, but since it is marked as a blocker I created #490 . I'm not sure if it solves the issue, because I observe a little bit different behavior than described. Namely, `(<= 0 1)` is just not reduced (instead of raising the error). Thus, it seems there is some difference in environments. So, @ngeiswei , please check if it works.

ngeiswei:
Thanks for being so reactive, @Necr0x0Der.  Yes the problem is on the rust end, but I could not figure what.  Your quick fix works however, so I'm going to use that for now and I'll let it to you and @vsbogd to decide what is a proper longer-term fix.

Necr0x0Der:
Could you please describe the environment, in which you got it. Was it a purely Rust version? It looks like that in your case `>=` was from Rust, while numbers where from Python. If so, it is not a Rust problem per se (and not a bug at all), but the question remains what to do with different versions of grounded atoms in different host languages.
But I still wonder how you got the error. It might be an issue for a specific environment meaning that some Rust tokens are loaded, which are language-specific and should not be loaded in presence of Python support.

ngeiswei:
* hyperon-experimental 21dbd82
* Python 3.9.2
* rustup 1.26.0
* cargo 1.73.0
* rustc 1.73.0

Anything else you need?

ngeiswei:
It was running from the `metta` command line interpreter, if that is what you were asking.

Necr0x0Der:
> It was running from the metta command line interpreter, if that is what you were asking.

Yes, I guessed so. Thus, this reveals that there is some difference in the behavior of running scripts from REPL and from Python runner. I don't think this difference is critical, but I'd like to hear some thoughts from @luketpeterson .

vsbogd:
[...]

PR #490, opened 2023-11-07
add <= and >= to Python stdlib

Necr0x0Der:
Relevant to #489


ngeiswei (approved):
I works.  I can use that branch for now so it's entirely up to you whether you want to merge it now or wait for a longer-term fix.

Necr0x0Der:
It's more an underimplemention issue. While there is still a longer-term issue #351 , this PR is definitely not a hacky hot-fix, since we have `<` and `>` in Python.

ngeiswei:
Oh I see, it was just missing on the python side!  Cool then, thanks again.

PR #494, opened 2023-11-08
Stop caching result of the grounded operations

vsbogd:
The simple example is calling grounded function to get time twice which returns same result.


vsbogd:
@DaddyWesker FYI

PR #495, opened 2023-11-09
Disable variable_operation feature by default

### LINKED ISSUE #242: Execution of expressions with variables as operations

vsbogd:
Add comments to the features. Disable `variable_operation` feature by default. This is to change behaviour according to comments https://github.com/trueagi-io/hyperon-experimental/issues/242#issuecomment-1801136493 and https://github.com/trueagi-io/hyperon-experimental/issues/242#issuecomment-1803609289

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -19,6 +19,8 @@ crate-type = ["lib"]
 
 [features]
 default = []
+# Add one of the features below into default list to enable.
+# See https://doc.rust-lang.org/cargo/reference/features.html#the-features-section
 minimal = [] # enables minimal MeTTa interpreter
 variable_operation = [] # enables evaluation of the expressions which have
                         # a variable on the first position, doesn't affect
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -18,7 +18,8 @@ path = "src/lib.rs"
 crate-type = ["lib"]
 
 [features]
-default = ["variable_operation"]
-#default = ["variable_operation", "minimal"]
-minimal = []
-variable_operation = []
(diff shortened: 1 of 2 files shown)


lib/Cargo.toml
```diff
@@ -18,7 +18,8 @@ path = "src/lib.rs"
 crate-type = ["lib"]
 
 [features]
-default = ["variable_operation"]
-#default = ["variable_operation", "minimal"]
-minimal = []
-variable_operation = []
+default = []
```
Necr0x0Der, line 21: Since not everyone might be familiar with exact way of enabling features in Cargo.toml, maybe, it is worthy to write a comment to `[features]` section how to enable features or don't remove ` #default = ["variable_operation", "minimal"]`?

vsbogd: done in https://github.com/trueagi-io/hyperon-experimental/pull/495/commits/e4ede3382f812a3d9f19b950eb2511093cce33b1

Necr0x0Der (approved):

PR #499, opened 2023-11-13
turn module path to file path in extend-py

Necr0x0Der:
The reason for the change.
`MeTTa.load_py_module(self, mod_name)` in `load_py_module_from_mod_or_file` may fail even if `mod_name` is a relative path w.r.t. cwd. It happens if `mod_name` contains submodules, e.g. `(extend-py! A.B)`. This doesn't happen with `export PYTHONPATH="${PYTHONPATH}:."`. Thus, this change might not be required. However, since we have
```py
# If that failed, try and load the module from a file
file_name = mod_name if ".py"
```
this means that we try to turn the module name into the file path. Turning `A.B` into `A.B.py` will not work in any case. If we turn it into `A/B.py` it will make more sense. It makes adding `.` to `PYTHONPATH` unnecessary.

### DIFF
--- a/python/hyperon/runner.py
+++ b/python/hyperon/runner.py
@@ -133,9 +133,10 @@ class MeTTa:
         if not isinstance(mod_name, str):
             mod_name = repr(mod_name)
         mod = MeTTa.load_py_module(self, mod_name)
-        if (mod is None):
+        if mod is None:
             # If that failed, try and load the module from a file
-            file_name = mod_name + ".py"
+            file_name = mod_name if ".py" in mod_name else \
+                        mod_name.replace('.', '/') + ".py"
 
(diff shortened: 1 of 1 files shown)


Necr0x0Der:
Well, it appears to be not a complete solution, because `spec.loader.exec_module(module)` fails. However, the file is at least found.

Necr0x0Der:
@luketpeterson , I'm not sure how this PR interferes with your work on modules. If there are already some conflicts, they you can close this PR. If it doesn't conflict, but you envision a revision of this code in the nearest future, you can merge it (or close for some reasons). If you don't envision this code to be reworked, I can ask someone from our team to flesh this out.

Necr0x0Der:
Yes, it arose from a local practical need.

PR #501, opened 2023-11-15
Remove Bindings::resolve_and_remove method

vsbogd:
Remove resolve_and_remove because it makes logic of Bindings more complex and doesn't add anything valuable. It can be replaced by applying `Bindings` to the variable.


luketpeterson (approved):
I'm happy with this clean-up

vsbogd:
Thanks Luke!

PR #502, opened 2023-11-15
Fix unit test to take into account global counter state

vsbogd:
Fix CI failure https://github.com/trueagi-io/hyperon-experimental/actions/runs/6875166920/job/18698225594


luketpeterson (approved):
To be honest, I don't love the Variable differentiation mechanism generally (unique ids that are part of the Atom).  But that aspect of the design is not the subject of the PR.

This PR is just to change the test, and I don't see any issues with the change.

PR #503, opened 2023-11-15
Relext

Necr0x0Der:
@luketpeterson , sorry, it's a dirty-hacky hotfix, but I need it now for extensions from relative paths, e.g. `(extend-py! ../myext/ext1.py)`, since `sys.modules[mod_name] = module` doesn't work for `mod = import_module(name)` if `mod_name` contains `..` or `/`. 
Hopefully, it doesn't interfere much with your work.


luketpeterson (approved):
Given it's a "dirty hacky hotfix", I'm fine with it.

But I need you to help me understand the conceptual need, so I can make sure the module resolution behavior supports this elegantly.

My thinking is that a module can have a single (optional) resource path, which ultimately maps to a file system directory (can be a directory with contents cloned from git, or whatever)  The point is that the module should only care about relative locations inside the directory, and not the position of the directory in the rest of the file system.

I am guessing you want a module to import a peer module as its dependency.  (peer in the file system)  If that's the case, the module resolution logic is designed to handle that elegantly.  So long as the imports aren't circular.

Is that all you need, or is there something else more subtle?

Necr0x0Der:
My current usecase is quite simple. I want to run scripts like this
https://github.com/trueagi-io/metta-motto/blob/main/tests/
without additional environment configuration.
While the structure of the package/extension can become different and more strict, there will still be the need for loading this extension using the relative path. With the present PR, these scripts work well. I'm ok if the syntax changes from `!(extend-py! ../motto/llm_gate.py)` to, say, `!(extend-py ../motto)` or `!(from ../motto use llm_gate)` or whatever else, but the very possibility to load extensions using relative paths should remain.

Necr0x0Der:
Well, in Python, relative imports are allowed only within packages. If one tries a relative import from a script, it is blocked with "No parent package" error. A typical solution is to create a symlink to the local package folder, which is also supported by github. Thus, I will possibly change `!(extend-py! ../motto/llm_gate.py)` to just `!(extend-py! motto)` and add the symlink in the motto library repo. Thus, I'm not insisting on supporting relative imports for this case. However, Python does support relative imports within packages, and we may find this useful for large packages as well. We can sort this out later...

PR #508, opened 2023-11-17
changed call of query to pattern_matcher_query in dasgete.py

CICS-Oleg:


CICS-Oleg:
The last commint doesn't work for` !(match &das (Test $v2) (This $v2 works))`, but this one `!(match &das ($v1 (Test 2)) (This $v1 works))` matches...

CICS-Oleg:
The last commit works with hyperon-das 0.3.10 version

PR #512, opened 2023-11-21
added sealed operator

cowboyphilosopher:
added new operator sealed.
format: (sealed (Expression with sealed variable list) (Term with variables to seal))
example: (sealed ($x $y) (= ($x $y) ($a $b))
$x and $y should be replaced with new unique variables in the Term value
this new version uses UUID values to make them more explicitly unique rather than the ReplacingMapper

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -12,6 +12,15 @@ env_logger = "0.8.4"
 directories = "5.0.1" # For Environment to find platform-specific config location
 smallvec = "1.10.0"
 
+[dependencies.uuid]
+version = "1.5.0"
+features = [
+    "v4",                # Lets you generate random UUIDs
+    "fast-rng",          # Use a faster (but still sufficiently random) RNG
+    "macro-diagnostics", # Enable better diagnostics for compile-time UUIDs
+]
+
+
 [lib]
 name = "hyperon"
 path = "src/lib.rs"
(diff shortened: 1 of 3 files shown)


lib/Cargo.toml
```diff
@@ -12,6 +12,15 @@ env_logger = "0.8.4"
 directories = "5.0.1" # For Environment to find platform-specific config location
 smallvec = "1.10.0"
 
+[dependencies.uuid]
+version = "1.5.0"
+features = [
+    "v4",                # Lets you generate random UUIDs
+    "fast-rng",          # Use a faster (but still sufficiently random) RNG
+    "macro-diagnostics", # Enable better diagnostics for compile-time UUIDs
+]
```
luketpeterson, line 21: A very minor point, but style-wise personally prefer to use the more compact notation unless there was a particular reason you want a comment explaining your choice of features.

e.g. `uuid = {version="1.5.0", features=["v4", "fast-rng", "macro-diagnostics"]}`

Also, uuid v 1.6.x has been released.  https://crates.io/crates/uuid

vsbogd:
@cowboyphilosopher, question to the previous PR https://github.com/trueagi-io/hyperon-experimental/pull/392#pullrequestreview-1573961189 is still valid. 
In practice every query to the atomspace replaces variables in returned atoms, thus effectively `sealed` doesn't add any value.

vsbogd:
Replaced by #609

PR #517, opened 2023-12-01
Add get-metatype op into stdlib

vsbogd:


vsbogd:
@luketpeterson just realized this intersects with your changes

PR #518, opened 2023-12-01
Fix (unify ($x a) (b $x) ok nok) case

vsbogd:
(let (quoted $a) (quoted $b) ...) replaces variables in $a by the unique variables thus making matching in error case possible. Case operation also has this property. Thus new function is required to check the case when variables cannot be matched to fix the issue. 

@tanksha please check it fixes the case you have.

### DIFF
--- a/lib/src/metta/runner/stdlib.rs
+++ b/lib/src/metta/runner/stdlib.rs
@@ -1274,7 +1274,11 @@ pub static METTA_CODE: &'static str = "
     (: unify (-> Atom Atom Atom Atom %Undefined%))
     (= (unify $a $a $then $else) $then)
     (= (unify $a $b $then $else)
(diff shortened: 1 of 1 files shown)


vsbogd:
I compared performance with version below and implementation using `case` is faster, so I included it in  this PR.
```
    (: unify (-> Atom Atom Atom Atom %Undefined%))
    (= (unify $a $b $then $else)
      (if (are-not-unified (are-unified $a $b)) $else $then))
    (: are-unified (-> Atom Atom Bool))
    (= (are-unified $a $a) True)
    (: are-not-unified (-> Bool Bool))
    (= (are-not-unified True) False)
    (= (are-not-unified (are-unified $x $y)) True)
```

Necr0x0Der (approved):

noskill (approved):

PR #522, opened 2023-12-18
Fix minimal interpreter to pass all tests except import!

vsbogd:
Filter out type errors and interpretation errors from the final result to make almost all tests green. One test still fails (this one https://github.com/trueagi-io/hyperon-experimental/blob/main/python/tests/scripts/f1_imports.metta) but it seems the reason is one of the module system improvements made recently. I would like to wait for a module system refactoring before fixing it. Most changes in PR are attempts to make code faster and simpler after making tests green. 

Further plans are to make code faster and think about making minimal MeTTa semantics more clear and explaining it in the documentation.

In order to be able simpler debug the code printing a stack trace of minimal MeTTa calls is added. Stack has a structure which allows searching for the function call point and function return point in the log.

Status of the current operation is explicitly represented in the interpreter by two values `Final` and `InProgress`. It allows using `chain` with any other minimal MeTTa operation and `chain` uses this flag to understand whether the operation is finished and its result should be applied to the last argument of `chain`.

Collapse and superpose operations are added into minimal instruction set but with unique names `collapse-bind` and `superpose-bind`. They operate not by a expression of atoms but by expression of pairs `(atom, bindings)` in order to allow `superpose` keep tracking bindings after `collapse`. These two instructions are used to replace the workarounds introduced before. This change made interpreter slower because of processing results of `collapse` using `foldl-atom` and `map-atom` implemented in MeTTa.

One cosmetic change which affects some old code: I have to introduce `VariableSet` in order to allow passing both `HashSet<VariableAtom>` and `HasSet<&VariableAtom>` to the `Bindings`'s methods.


Necr0x0Der (approved):

luketpeterson (approved):

luketpeterson (dismissed):
These all look fine.

One minor thing: Is there a reason not to change the `&dyn VariableSet` argument to a generic?  This would then mean you could change the Iterator type in the `VariableSet` trait into an associated type, and eliminate the box for the return type on `iter()`.

It's an easy change to make if it ends up mattering.

vsbogd:
> One minor thing: Is there a reason not to change the `&dyn VariableSet` argument to a generic? This would then mean you could change the Iterator type in the `VariableSet` trait into an associated type, and eliminate the box for the return type on `iter()`.

Do you mean this https://github.com/trueagi-io/hyperon-experimental/pull/522/commits/ee447d168d52b39ca27d2a294573823403b075e6 ?

luketpeterson:
Exactly!  :-)

luketpeterson (approved):

vsbogd:
Thanks Luke!

vsbogd:
There is a code duplication in `interpreter2.rs` where operation arguments are parsed twice to print stack and interpret them. I think it makes sense to eliminate this duplication but I will try doing it in a separate PR.

PR #523, opened 2023-12-18
Remove _e2 test as outdated and never used

Necr0x0Der:


Necr0x0Der:
@vsbogd , @luketpeterson , FYI

PR #525, opened 2023-12-21
Fix log format strings in stdlib

vsbogd:


vsbogd:
Minor change to have logs in stdlib formatted using same pattern.

PR #526, opened 2023-12-27
add regex support

besSveta:


Necr0x0Der:
A few minor comments.

- `regex_objects.py` in the `hyperon` root doesn't look nice. I'd just put its code into `stdlib.py` for now, and wouldn't introduce a separate `regex_atom` just for one token. There are `text_ops`, where `regex` can be put for now. If regex-related functionality becomes richer (or united with other string-processing tools), then we can either have a subfolder (`stdlib` or whatever) with this and possibly other larger parts of `stdlib` in separate files, or move it to its own module.
- Regarding `[[`, etc., you can take a look how `String` is implemented in `stdlib` (with `"\"[^\"]*\""`). Something like `"r\"[^\"]*\""` (like in Python, or `"regex:...` if we want to be more verbose) may work
- `regex_matcher_test.metta` is not added as a unit test, and `scripts` folder is organized in a certain way. So, this test/example should either be put at some place in the other scripts, or should be just a Python test. But I can sort it out later, so maybe no actions should be taken regarding this point.

python/hyperon/stdlib.py
```diff
@@ -94,7 +131,8 @@ def type_tokens():
         r"[-+]?\d+(\.\d+)?e[-+]?\d+": lambda token: ValueAtom(float(token), 'Number'),
         "\"[^\"]*\"": lambda token: ValueAtom(str(token[1:-1]), 'String'),
         "\'[^\']\'": lambda token: ValueAtom(Char(token[1]), 'Char'),
-        r"True|False": lambda token: ValueAtom(token == 'True', 'Bool')
+        r"True|False": lambda token: ValueAtom(token == 'True', 'Bool'),
+        r'regex:"[^\"]*"': lambda token: RegexMatchableAtom(token)
```
Necr0x0Der, line 135: `lambda token: RegexMatchableAtom(token)` could be just `RegexMatchableAtom` I believe. Or you can just have `lambda token: G(RegexMatchableObject(token), 'Regex')` without `RegexMatchableAtom`. I don't see any immediate use of `RegexMatchableAtom` with additional `type_name` and `atom_id`.

Necr0x0Der (approved):
I merge it for now

PR #535, opened 2024-01-11
Using im::HashSet for the `interpreter2::Variables` type

luketpeterson:
The "Immutable Set" pattern is essentially already already being used through a combination of `Rc` and cloning.
~20% speedup for chain-heavy tests.

Unfortunately not a silver bullet, but a worthwhile change given its low impact.


vsbogd (approved):
Thanks Luke, it drastically improves performance on a deep recursive cases! It reached point where minimal interpreter performs better then pure Rust version when recursion is deep enough.

PR #542, opened 2024-01-17
Add remote das support to dasgate.py

CICS-Oleg:
Needs to cleanup some bugs


python/sandbox/das_gate/test_das.metta
```diff
@@ -1,6 +1,7 @@
 !(extend-py! dasgate)
 
-!(bind! &das (new-das))
+;!(bind! &das (new-das))
+!(bind! &das (new-remote-das 104.238.183.115 8081))
```
Necr0x0Der, line 4: I would put such an IP in the open repo

Necr0x0Der: It can be either a function for getting IP from env (by wrapping Python `os.environ`), or just
`!(bind! &das (new-remote-das (Error "Put your IP here" ()) 8081))`

Necr0x0Der (approved):

PR #550, opened 2024-01-27
Minimal integration of Hypervectors to Metta (via PyBHV python library)

astroseger:
Minimal integration for Hypervectors to Metta  (via PyBHV python library). 

It should be noted that integration of permutation probably could be done even better. See `02_example_perm.metta`. What we need is only perm1, but in the current implementation we need to write two lines to "create" perm1.

```
!(bind! &perm1 (bhv-new-perm))
(= (perm1 $x) (bhv-apply-perm &perm1 $x))
```
If we somehow could do it in one line, it would probably be better.


Necr0x0Der (approved):
I'm ok with it as a sandbox example of a very straightforward way of gating bhv to MeTTa, but I'd do this in a different way. I'd create a bhv space and let it keep bindings of symbols to hypervectors and to try making re-association of hypervectors back to symbols, when it works.

astroseger:
@Necr0x0Der I've fixed examples with simple !bind . So feel free to merge it, if it can be an example of a simple binding...

Necr0x0Der (approved):

PR #555, opened 2024-01-30
support for loading extensions from exts

Necr0x0Der:


luketpeterson:
Related to https://github.com/trueagi-io/hyperon-experimental/pull/554

Your proposal is fine... My concern is what it implies for the MeTTa module system more generally. Specifically with what I implemented, the module namespace is flat. I don't think it makes sense to delegate the module name resolution to Python. So that means is that MeTTa would need a hierarchical module namespace, and that all Python modules would need to be treated like empty MeTTa modules for the purpose of resolving sub-modules.

I *think* that would all work fine. But I want to try it out to be sure.

luketpeterson:
Ok.  I think it is fine to merge this.  I will make corresponding changes to keep the behavior consistent in the `modules` branch.

luketpeterson (approved):

Necr0x0Der:
> Ok. I think it is fine to merge this. I will make corresponding changes to keep the behavior consistent in the modules branch.

If your worry is about putting `hyperon.exts` directly in `extend-py`, then it can be moved into env_builder, but `__init__.py` fix will still be needed. If your concern is about both changes, then the question is whether the desirable behavior is already achieved or will be achieved by the modules branch. But this can be figured out after migration to the new modules system.

luketpeterson:
> If your worry is about putting hyperon.exts directly in extend-py, then it can be moved into env_builder, but __init__.py fix will still be needed. If your concern is about both changes, then the question is whether the desirable behavior is already achieved or will be achieved by the modules branch. But this can be figured out after migration to the new modules system.

I'm not worried about either.  It's just that the structure of the code is very different in the modules branch.  There is no `extend-py!` operation anymore because `import!` loads both pure MeTTa and python-based modules.  The handling of the `__init__.py` file would be done in the python module format, and looking in the hyperon.exts directory would be done in MeTTa.init through the environment initialization.

But it is fine. Those changes are straightforward and I will make them in my branch when I merge in this PR.

I got a little worried because I thought maybe you were saying it should be possible to load a nested sub-module without loading its parent module, but I realized you didn't actually say that - I was extrapolating incorrectly.  So I will avoid changing the behavior of the module namespace unless / until we have a good reason.

PR #557, opened 2024-01-31
Fix crash on retain all variables from single binding

### LINKED ISSUE #556: Run-time error on MeTTa script doing lambda abstraction
## What is the problem?

A MeTTa script involving backward chaining on lambda abstraction crashes the metta interpreter.

## How to reproduce the problem?

Run the following metta file:

```clojure
;; Knowledge base
!(bind! &kb (new-space))

;; Successor
!(add-atom &kb (: S (-> (: $x Nat) Nat)))

;; Anything can serve as a variable.  Even a closed term, it simply
;; will not unify with any other closed term than itself, which makes
;; for a very specific, yet valid, abstraction.
!(add-atom &kb (: $x Var))

;; Lambda abstraction
!(add-atom &kb (: λ (-> (: $x Var) (-> (: $y $b) (-> (: $x $a) $b)))))

;; Backward chainer
;; Base case
(= (bc (: $prf $ccln) $_)
   ;; Query the knowledge base for rule or fact
   (match &kb (: $prf $ccln) (: $prf $ccln)))
;; Recursive step
(= (bc (: ($prfabs $prfarg) $ccln) (S $k))
   (let* (;; Recurse on proof abstraction
          ((: $prfabs (-> (: $prfarg $prms) $ccln))
           (bc (: $prfabs (-> (: $prfarg $prms) $ccln)) $k))
          ;; Recurse on proof argument
          ((: $prfarg $prms)
           (bc (: $prfarg $prms) $k)))
     (: ($prfabs $prfarg) $ccln)))

;; Tests

;; Synthesize unary functions over natural numbers, that is
;; prove (-> (: $k Nat) Nat).
!(bc (: $prf (-> (: $k Nat) Nat)) (S (S (S (S (S Z))))))
```

## What happens?

It crashes with the following error

```
thread '<unnamed>' panicked at /home/nilg/Work/TrueAGI/hyperon-experimental/lib/src/atom/matcher.rs:199:14:
Unexpected state
```

## What should normally happen?

It should not crash.

## What else do have to say?

Sometimes it does not crash.  In fact one can control the probably of crashing by augmenting or reducing the depth of the backward chainer call.  A depth of 5 (as used in the metta script above) reliably crashes the interpreter.

A depth of 2, i.e.

```clojure
!(bc (: $prf (-> (: $k Nat) Nat)) (S (S Z)))
```

reliably does not crash the interpreter.

A depth of 3, crashes the interpreter sometimes.
A depth of 4, crashes the interpreter most of the time.

## Anything else?

I tried to provide a minimal metta script but anything I strip away from that script result in no longer crashing.

vsbogd:
Fixes #556

### DIFF
--- a/lib/src/atom/matcher.rs
+++ b/lib/src/atom/matcher.rs
@@ -181,7 +181,7 @@ impl Bindings {
                     self.bindings.remove(binding_id).atom
                 } else {
                     if binding.var == *var {
-                        self.rename_binding(binding_id, var);
+                        let _ = self.rename_binding(binding_id);
                     }
                     None
                 }
@@ -190,14 +190,15 @@ impl Bindings {
         }
     }
 
-    fn rename_binding(&mut self, binding_id: usize, old_var: &VariableAtom) {
+    fn rename_binding(&mut self, binding_id: usize) -> Result<(), ()> {
         let binding = &mut self.bindings[binding_id];
         let var = self.binding_by_var.iter()
-            .filter(|(v, i)| **i == binding.id && *v != old_var)
+            .filter(|(v, i)| **i == binding.id && **v != binding.var)
             .map(|(v, _i)| v)
             .next()
-            .expect("Unexpected state");
+            .ok_or(())?;
         binding.var = var.clone();
+        Ok(())
     }
 
     fn move_binding_to_binding(&mut self, from_binding_id: usize, to_binding_id: usize) {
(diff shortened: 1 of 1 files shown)


ngeiswei (approved):
Thanks for fixing it that quickly! (I don't understand the fix but I approve nonetheless).

luketpeterson (approved):
Looks good.  More unit tests are a good thing.

PR #558, opened 2024-01-31
Turn off evaluation cache in interpreter

### LINKED ISSUE #553: Unexpected behavior which seems to be caused by caching.
Let's define function which generate random number. For example via python binding.

```python
from hyperon.atoms import OperationAtom
from hyperon.ext import register_atoms
import random

@register_atoms
def my_atoms():
    return {
        'random': OperationAtom('random', random.random),
    }
```
Save this in `random_binding.py`
Now let's run the following metta script

```lisp
!(extend-py! random_binding)

; will print False as expected
!(println! (== (random) (random)))

; will print False as expected
!(let $a (random) (let $b (random) (println! (== $a $b))))


(= (aa) (random)) 

; will print True... Seems to be a bug
!(println! (== (aa) (aa)))

; will print True... Seems to be a bug
!(let $a (aa) (let $b (aa) (println! (== $a $b))))


!(bind! a1 (aa))
!(bind! a2 (aa))

; will print False as expected 
!(println! (== a1 a2))
```

vsbogd:
Fixes #553, the price is some decrease of the performance. About 20% on SICP example. Taking into account we anyway plan to migrate to minimal MeTTa which doesn't use caching it should not be an issue.

### DIFF
--- a/lib/src/metta/interpreter.rs
+++ b/lib/src/metta/interpreter.rs
@@ -219,11 +219,32 @@ pub fn interpret<T: Space>(space: T, expr: &Atom) -> Result<Vec<Atom>, String> {
     }
 }
 
+#[derive(Debug)]
+struct NoInterpreterCache();
+impl NoInterpreterCache {
+    fn new() -> Self {
+        Self()
+    }
+
+    fn get(&self, _key: &Atom) -> Option<Results> {
+        None
+    }
+
+    fn insert(&mut self, _key: Atom, mut _value: Results) {
+    }
+}
+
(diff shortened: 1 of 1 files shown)


Necr0x0Der:
> About 20% on SICP example. 

I wonder if it is an issue for other projects like PLN or Pattern Miner... I agree that migration to minimal metta makes this change safe, but do we really need it? bhv integration will be done in a different way in any case.

Necr0x0Der (approved):

vsbogd:
Ok, I agree that it may have some consequences for PLN and Pattern Miner, so lets not merge it.

PR #567, opened 2024-02-01
Fix unstable minimal test

vsbogd:
Main root cause is an inaccurate work with variable bindings. Standard library for minimal interpreter includes a lot of MeTTa code, and thus much more variable bindings are generated during interpretation. Most of these bindings are temporal and should be cleaned up to maximize performance. Inaccurate cleaning up causes different issues in interpreted code. This also causes 50% performance improvement for minimal MeTTa.

This PR fixes:
- work with variables for `superpose` and `collapse`: take into account variable bindings which were found between `collapse` and `superpose` of the `collapse` results. Keep variables which were collected before `collapse` is called and put by `collapse` inside resulting atoms.
- work with variables after querying atomspace, on the one hand more agressive cleanup is performed on the other hand new variables which were introduced in atoms are kept.


vsbogd:
I also investigated why `f1_imports.metta` fails and found that in `trueagi/main` it fails because MeTTa part of a standard library is not included into a loading runner instance. In old interpreter it also causes troubles when pure MeTTa functions are called from imported space. These calls are finished incorrectly. In minimal MeTTa the intrepreter itself is defined in MeTTa part of standard library and it leads to the situation when no code in imported space is executed.

vsbogd:
@luketpeterson, I am merging it as mainly this change touches minimal interpreter and it is more a hotfix. If you have any comments please leave here, I will address them.

PR #569, opened 2024-02-02
Add Cargo workspace configuration

vsbogd:
Introducing virtual workspace Cargo.toml file allows reusing workspace configuration (version, dependencies) in all Rust subprojects.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,0 +1,17 @@
+[workspace]
+members = [
+    "lib",
+    "c",
+    "repl",
+]
+resolver = "2"
+
+[workspace.package]
+version = "0.1.6"
+edition = "2021"
+
+[workspace.dependencies]
+hyperon = { path = "./lib", version = "0.1.6" }
+regex = "1.5.4"
+log = "0.4.0"
+env_logger = "0.8.4"
--- a/README.md
+++ b/README.md
@@ -70,12 +70,10 @@ python3 -m pip install pip==23.1.2
 
 # Build and run
 
-## Hyperon library
+## Rust library and REPL
 
-Build and test the library:
+Build and test the Rust binaries:
 ```
-cd ./lib
-cargo build
 cargo test
 ```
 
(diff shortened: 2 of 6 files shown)


vsbogd:
I changed version of the REPL to the workspace version because I don't think we will release them independently. @luketpeterson what do you think?

luketpeterson:
> I changed version of the REPL to the workspace version because I don't think we will release them independently. @luketpeterson what do you think?

I this this is generally the right thing to do.

As things currently stand there are basically two ways to build the REPL, with or without Python.  If you build it with Python, it doesn't actually call hyperon lib directly, only through Python.  If you build with `--features no_python` then it calls hyperon lib directly.  The default is Python.

So, by default, the Rust hyperon lib gets statically linked into hyperonc which is then linked by hyperonpy, which is finally called by the `hyperon` python module.

In the future we may want to make the hyperon lib a stand-alone component, but I think that is further down the road.

luketpeterson:
I figured out the crash.  It's actually a really subtle bug and I'm not quite sure why this didn't bite us sooner.

It's in the `tokenizer_register_token` function of the C API.  The issue is this code that says:
```
    let c_token = CToken{ context, api };
    tokenizer.register_token(regex, move |token| {
        let constr = unsafe{ (&*c_token.api).construct_atom };
        let atom = constr(str_as_cstr(token).as_ptr(), c_token.context);
        atom.into_inner()
    });
```
The naive expectation is that the closure would capture the c_token, and then the c_token object would be dropped when the closure object was dropped.  But because the closure is not `FnOnce` (because it needs to be able to run multiple times) c_token can't capture by value.  So instead it captures the pointers inside `c_token`.

Then, when c_token goes out of scope, its drop implementation is called, freeing the object holding the pyfunction needed to make the token.

The fact the compiler didn't catch it is a perfect storm of `unsafe` and https://doc.rust-lang.org/edition-guide/rust-2021/disjoint-capture-in-closures.html

luketpeterson:
I'm not totally sure why it broke now, but my best guess is that the version of the Rust language changed from 2018 to 2021, which subtilely changed the rules about variable captures.

Anyway, I added a fix to make the capture explicit

luketpeterson (approved):

PR #570, opened 2024-02-02
Fix REPL compiler error and warning

vsbogd:
Small fix for the error when building REPL with --features minimal. Fix lifetime eliding warning.


vsbogd:
I see that compilation error raises only when project is built with `--features minimal`, probably because `no_python` is enabled. But I am not sure why Python is disabled when minimal is enabled?

luketpeterson (approved):
I had exactly the same changes character-for-character in my branch.  I should have opened this PR to save you the work.  Sorry.

PR #573, opened 2024-02-06
Set name for the built-in spaces

vsbogd:
Add GroundingSpace Builder to be able setting name in one line. It can be extended to add atoms as well.

@luketpeterson , this is a PR to discuss briefly if we need builder here. I added it because I wanted be able to construct `GroundingSpace` in one line instead of calling new. But if you don't like the style or have other objections please feel free to reject :-)

### DIFF
--- a/lib/src/metta/runner/mod.rs
+++ b/lib/src/metta/runner/mod.rs
@@ -93,7 +93,7 @@ impl Metta {
     {
         let space = match space {
             Some(space) => space,
-            None => DynSpace::new(GroundingSpace::new())
+            None => DynSpace::new(GroundingSpace::builder().set_name("runner".into()).build())
         };
 
         //Create the raw MeTTa runner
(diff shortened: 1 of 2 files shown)


luketpeterson (approved):
I'm ok with the code as it's written.  But can I ask about the motivation behind these changes?

The reason I ask is because some code in the Modules branch lets you pass in a Space to use for a newly initialized module, but then that module init code also sets the space name to the module name, unless the space already has a name.

If there is a good reason to keep spaces immutable after they are created, hence the "Builder" pattern, then I will work out another way to deal with module init (maybe take a SpaceBuilder trait object when creating a module, or more likely just eliminating the name setting code except for spaces that are created by default within the module init function).  Thanks.

vsbogd:
To me binding an instance of space to its name in a moment of its creation is just natural. I didn't know about `set_name` usages in your branch, so I introduced `Builder` to be able initializing space in a single statement instead of using two statements: create space, set space name.

> The reason I ask is because some code in the Modules branch lets you pass in a Space to use for a newly initialized module, but then that module init code also sets the space name to the module name, unless the space already has a name.

If a space is created before a module is created and we want to see the name of the module in logs then we probably should put a module into a space explicitly instead of having space representing the module inside other space.

I think the best way to deal with this PR is to close it and continue discussion in your module system PR.

luketpeterson:
I don't have a strong opinion.  But my weak opinion is that I don't like a "Builder" pattern for Spaces.

A new Builder object is a lot of new API surface area, and if we want a 1-line creation function with a name we can add a `new_with_name` method.  My opinion would be different if there were a lot of parameters, and especially if some of them were incompatible with others, or if the parameters needed to be built up over time.

Again this is just my slight opinion so please feel free to disagree.

vsbogd:
One more use-case for this, which I had in mind is initialization of the space content in unit-tests. But anyway I agree that introducing Builder just for setting name is overkill.

Issue #572, opened 2024-02-02
MeTTa definitions of stdlib don't work in modules imported by import! function


vsbogd:
Code which uses MeTTa definitions of stdlib inside a module imported using `import!` doesn't work:

**Steps to reproduce**
`imported.metta`:
```
!(if True imported-ok imported-nok)
```

`import.metta`:
```
!(import! &m imported.metta)

!(if True self-ok self-nok)
```

Run `import.metta`

**Expected result**
```
metta import.metta
[()]
[self-ok]
```

**Actual result**
```
metta import.metta
[()]
[self-ok]
```

**Root cause**
`Metta::new_loading_runner` creates an empty space but doesn't import stdlib into it while it gets all standard tokens. Thus all MeTTa definitions of the stdlib are not available for the interpreted code. It affects minimal MeTTa even more serious because interpreter itself is also part of the MeTTa stdlib code.

https://github.com/trueagi-io/hyperon-experimental/blob/0c17c629dce868a3d954c2ff725e7586ac4c8a8b/lib/src/metta/runner/mod.rs#L151-L164

This should be fixed during work on https://github.com/trueagi-io/hyperon-experimental/issues/470

vsbogd:
It is a blocker for minimal MeTTa work.

PR #574, opened 2024-02-07
Import hotfix

vsbogd:
Hotfix for the #572 which can be used before modules system refactoring is merged and unblocks minimal MeTTa.

The idea of the fix is to add a temporal instance of atomspace which contains stdlib into an imported space and remove it after running imported code. Also needed to fix a test to not rely on an order of atoms returned. Last thing I have found that `get-atoms` returns non-unique instances of the variables and thus unification of `$x` and `(get-atoms $m)` is not possible when `&m` contains instances of `$x`.

### DIFF
--- a/lib/src/metta/runner/mod.rs
+++ b/lib/src/metta/runner/mod.rs
@@ -147,6 +147,11 @@ impl Metta {
         metta
     }
 
+    // Loads stdlib into a separate space and returns it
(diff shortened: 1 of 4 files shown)


luketpeterson (approved):
This all looks fine.  In some ways (conceptually) this does for stdlib what the module system does for all modules - although I don't love the behavior in the long term.

PR #581, opened 2024-02-12
Fixing build warning, on account of Rust 1.76 being more strict about pointer comparisons

luketpeterson:
Fixing build warning, on account of Rust 1.76 being more strict about pointer comparisons


vsbogd (approved):
Thanks Luke!

PR #583, opened 2024-02-12
Handle Python 3.12

vsbogd:
Conan version is upgraded to 1.62, Python 3.12 is excluded from supported versions because `subunit 1.4.0` dependency of `hyperonc` is incompatible with Python 3.12. This is hotfix to unblock Python packages release.


vsbogd:
See https://github.com/trueagi-io/hyperon-experimental/issues/584 for the issue description

PR #588, opened 2024-02-13
Make type of the EqualOp compatible with one from Python stdlib

vsbogd:
In Python stdlib `==` has type `(-> $t $t Bool)` which makes interpreter to evaluate each argument before passing them to `==`. In Rust stdlib it was `(-> Atom Atom Bool)` which causes different behaviour.

This PR is to discuss and unify the behaviour.

### DIFF
--- a/lib/src/metta/runner/stdlib.rs
+++ b/lib/src/metta/runner/stdlib.rs
@@ -1123,7 +1123,7 @@ impl Display for EqualOp {
 
 impl Grounded for EqualOp {
     fn type_(&self) -> Atom {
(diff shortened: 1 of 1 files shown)


Necr0x0Der (approved):

vsbogd:
JIC PR where `EqualOp` was introduced https://github.com/trueagi-io/hyperon-experimental/pull/462

luketpeterson (approved):
Your reasoning makes sense.  I don't see any problems with the change.  I'm not sure I have much more to add regarding the subtlety as you two have a better understanding of the MeTTa type system.

PR #593, opened 2024-02-19
fast simplified load

Necr0x0Der:
The reason for this PR: it is necessary to load large (n*100Mb) files with simple content (mostly symbolic, but possibly with numbers). Customized tokenizer works slowly and takes additional memory, which is critical for experimenting with such files. Python loader for such files works several times slower and takes several times more memory than C++ version. C++ version is preferable over Rust since Python objects are needed. While this might not be a fundamental solution, it is needed atm.


Necr0x0Der:
@vsbogd , I wonder why it doesn't work in Minimal MeTTa? It seems related not to the added functionality, but to `!(match &space $x $x)`

vsbogd:
Raised fix as https://github.com/trueagi-io/hyperon-experimental/pull/594

Necr0x0Der:
> I can replace match by grounded version in a minimal MeTTa as a quick fix.

I would be ok with fixing the test in this PR, but if this difference touches other use cases, then I'm ok with your hotfix as well.

> should we implement fast load in Rust to make it available for a Rust REPL implementation as well, or you would like to start in Python to experiment with it?

Let's keep it as a Python version for experimentation. If/when we have automatic conversion both ways and if it works fast, we can put fast load into Rust and load Rust grounded atoms.

vsbogd:
> if this difference touches other use cases, then I'm ok with your hotfix as well.

Potentially it can touch other code which uses single variables as a patterns in `match`. Thus I would merge it anyway.

vsbogd (approved):

PR #594, opened 2024-02-20
Replace match implemented via unify by the grounded version

vsbogd:
`unify` doesn't behave as `match` in a case when space is matched with variable. `match` returns atoms from the space. `unify` assigns the space as a value for the variable.


Necr0x0Der (approved):
I'm ok with it if you consider this is as a suitable approach for minimal metta as a temporal hotfix at least.

PR #596, opened 2024-02-22
add escape seq to fast_load

Necr0x0Der:


Necr0x0Der:
@besSveta , please check that new bio-ai files can be correctly loaded with this change.

besSveta (approved):

PR #603, opened 2024-02-27
An update to module loading, mainly for relative paths and recursive loading

luketpeterson:
This addresses a number of the outstanding issues (unfortunately not all yet) with the module loading behavior.

- It's now possible to reliably use the `self` token to refer to the currently running module in a module name path
- explicit loading of sub-modules will now trigger parent modules to load
- All directories can be loaded as modules, making their contents available as sub-modules


Necr0x0Der:
Seems to be working for me. Thanks, @luketpeterson . 
@besSveta , could you please check that this works for you to import Python extensions in the SNet Assistant from prototyping subfolders.

besSveta:
> Seems to be working for me. Thanks, @luketpeterson . @besSveta , could you please check that this works for you to import Python extensions in the SNet Assistant from prototyping subfolders.

yes it works for SNet Assistant

Necr0x0Der (approved):

Issue #432, opened 2023-09-13
REPL cannot load `libpython` when built and run under Python environment


vsbogd:
I use [Conda environment](https://docs.conda.io/projects/miniconda/en/latest/) to manage Python versions and dependencies. When I build MeTTa REPL under Conda and try to run it after the build I see the following error:
```
$ ./target/debug/metta 
./target/debug/metta: error while loading shared libraries: libpython3.7m.so.1.0: cannot open shared object file: No such file or directory
```

According to https://github.com/PyO3/pyo3/issues/1576 when using PyO3 app under Python env one need to add the path to the Python library into `LD_LIBRARY_PATH`. This kind of works but breaks other executables because `lib` directory in Conda env contains other libraries which overrides system library loading paths.

luketpeterson:
When the repl is built, it looks like we can point pyo3 at the python version we want to use. https://pyo3.rs/v0.19.2/building_and_distribution.html#configuring-the-python-version

So I'll see what's involved in querying Conda, and potentially overriding pyo3's defaults if Conda is installed.  The other option (which we might want to do regardless) is to document the `PYO3_PYTHON` environment variable in our docs.

vsbogd:
I didn't mention it in an issue description, but in fact PyO3 compile against and uses absolutely correct version of the library. So I don't see any inconsistency. The issue is that this shared Python library from Conda is not found by OS when program is executed under the same environment. It sounds more like a Conda issue btw.

vsbogd:
This is [how Conda manages it for its packages](https://docs.conda.io/projects/conda-build/en/stable/resources/use-shared-libraries.html#shared-libraries-in-macos-and-linux):
> On Linux, conda-build modifies any shared libraries or generated executables to use a relative dynamic link by calling the patchelf tool. On macOS, the install_name_tool tool is used.

vsbogd:
I would say the proper way to fix is to have two different dynamic links for two different cases:
- when executable is being built under Python virtual env (library is not visible to the system loader) we should use absolute link to the Python library inside executable
- when executable is being built against system Python library (library is visible to the system loader) we use just the name of the library without path as a dynamic loading link.

luketpeterson:
> This is [how Conda manages it for its packages](https://docs.conda.io/projects/conda-build/en/stable/resources/use-shared-libraries.html#shared-libraries-in-macos-and-linux):
> 
[...]

PR #605, opened 2024-02-27
Enable REPL as a default MeTTa script inside Docker

vsbogd:
Enable REPL as a default MeTTa script inside Docker. Make a release build inside Docker by default.
NB: this Docker image contains temporary build files has big size and thus cannot be used as a MeTTa distribution package. If we want to distribute MeTTa as a Docker image it is better to improve this script or add another Dockerfile.


Necr0x0Der:
I don't have a strong opinion whether REPL should be the default MeTTa runner in Docker. I'm OK with this if this is a request from the Docker users.

vsbogd:
> I don't have a strong opinion whether REPL should be the default MeTTa runner in Docker. I'm OK with this if this is a request from the Docker users.

It was not requested by the Docker users but one of the goals of adding REPL was to add additional flexibility to the MeTTa command line tool: syntax highlighting, step by step execution. Should we start using REPL by default?

luketpeterson:
I don't have a strong opinion here either.  The REPL app has more capabilities than the Python runner (namely interactive mode)

Will https://github.com/trueagi-io/hyperon-experimental/issues/432 ever be issue or does the Docker environment mean this won't come up?  I think ie makes sense to merge it if the conda / pyenv issue is moot.

luketpeterson (approved):

vsbogd:
> Will https://github.com/trueagi-io/hyperon-experimental/issues/432 ever be issue or does the Docker environment mean this won't come up? I think ie makes sense to merge it if the conda / pyenv problem is a non-issue.

It will not be a problem inside a Docker container because global Python environment is used.

vsbogd:
Replaced by #732

PR #608, opened 2024-02-28
Make float tokens in Python and Rust stdlib equal

vsbogd:


luketpeterson (approved):
This change looks fine to me.

Issue #592, opened 2024-02-19
Change in import behavior for spaces


Necr0x0Der:
With the new module system, I got the following error:
```
(Error (import! GroundingSpace-0x12a7846d8 space.metta) import! destination argument must be a symbol atom naming a new space, or &self.  Found: GroundingSpace-0x12a7846d8)
```
Why can't we import into the already existing space? It is inconvenient.

luketpeterson:
The short answer is because the behavior of the `import!` atom is a bad conceptual fit for the abstractions in the new MeTTa (or any language's) module system.  For example, it wouldn't make any sense in Python to say `from numpy import * into torch`.  If torch needed something from numpy it would do the import itself.

In MeTTa there is also a question about what to do about tokens from the modules, although that question also exists for just importing into a new space.  https://github.com/trueagi-io/hyperon-experimental/blob/8540b90d8669484e945fb78d0d11135b68de9233/docs/modules_internal_discussion.md?plain=1#L41C1-L42C1

The arguments to `import!` need to be redone to better reflect the underlying semantics of the operation as discussed here: https://github.com/trueagi-io/hyperon-experimental/issues/509 and also https://github.com/trueagi-io/hyperon-experimental/pull/580#discussion_r1491332304

But this issue seems to be about providing a convenience.  So I have two ideas:
1.) The `inline!` operation.  I could add a space argument to it.  So any atoms created by the inlined code would go into the specified space. `inline!` seems to be what a lot of people want for a lot of reasons.
2.) An automatic space-merge if the specified space already exists.  I don't like this because it further muddies up what `import!` is conceptually supposed to do.  But it would bring back the old behavior.

Necr0x0Der:
> The current behavior of the import! operation is a bad conceptual fit for the abstractions in the new MeTTa (or any language's) module system. For example, it wouldn't make any sense in Python to say from numpy import * into torch

[...]

Issue #510, opened 2023-11-20
Should Tokenizer Entries have names or be associated with Type atoms?


luketpeterson:
Question that came up in the implementation of MettaMod::import_item_from_dependency_as
(Think "from module import foo as bar" in Python)

Behavior as implemented:
* If the `from_name` argument exactly matches a [Tokenizer] entry in the source module,
    then that entry will be imported, and the `name` argument will be ignored.  In this case no
    atom is imported.
* If an exact [Tokenizer] entry was not found, this method will attempt to resolve `from_name`
    into an atom, using the [Tokenizer] and [Space] associated with the dependent (source) module,
    and the resolved atom will be imported into the `&self` [Space]
* If `name` is provided, then if the resolved atom not a Symbol or if the resolved atom is a
    symbol that doesn't perfectly match `name`, a new [Tokenizer] entry will be created to
    access the atom in the &self module

However this strikes me as very clumsy.

Exactly matching a regex makes importing a Tokenizer pattern pretty unfriendly.
Does it make sense to require Tokenizers entries to be associated with atoms, for
example "Type Atoms"?  For example, we could have an "Number" type that is tied to all the
Tokenizer regex patters used to parse different types of numbers?  Then a user could
"!(import! Number from Arithmetic)" or whatever, and get all the Tokenizer patters that parse
numbers?

vsbogd:
Copied from https://github.com/trueagi-io/hyperon-experimental/pull/580#discussion_r1491280940

Yes, I think it make sense to add some identifiers to the tokens for the importing purposes. It is not a blocker for this PR and can be added as a separate change. I don't think it must be a type. For example if we are importing a token which is create by bind! it is better to import it by a token name rather than by type.

PR #615, opened 2024-03-10
Adding `mod-space!` Op Atom, to access the space of a module, loading it if necessary

luketpeterson:
Adding `mod-space!` Op Atom, to access the space of a module, loading it if necessary
See https://github.com/trueagi-io/hyperon-experimental/issues/592

It's now possible to do this: `!(add-atom &new_space (mod-space! mod_name))`

Therefore, `!(add-atom &self (mod-space! stdlib))` adds the same atom(s) as `!(import! &self stdlib)`, but does not add any tokens.

For me the remaining question is whether we want to enshrine the "nested-sub-space" behavior into MeTTa semantics?  We could add a `merge-atoms` or similar operation, to make it clear we're bringing in all atoms from the source space, but the implementation in the present code would be identical to `add-atom`

### DIFF
--- a/lib/src/metta/runner/stdlib.rs
+++ b/lib/src/metta/runner/stdlib.rs
@@ -146,6 +146,64 @@ fn strip_quotes(src: &str) -> &str {
     src
 }
 
+/// mod-space! returns the space of a specified module, loading the module if it's not loaded already
+//NOTE: The "impure" '!' denoted in the op atom name is due to the side effect of loading the module.  If
+// we want a side-effect-free version, it could be implemented by calling `RunContext::get_module_by_name`
+// instead of `RunContext::load_module`, but then the user would need to use `register-module!`, `import!`,
+// or some other mechanism to make sure the module is loaded in advance.
+#[derive(Clone, Debug)]
+pub struct ModSpaceOp {
+    //TODO-HACK: This is a terrible horrible ugly hack that should be fixed ASAP
+    context: std::sync::Arc<std::sync::Mutex<Vec<std::sync::Arc<std::sync::Mutex<&'static mut RunContext<'static, 'static, 'static>>>>>>,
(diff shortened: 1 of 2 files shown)


Necr0x0Der (commented):
Taking into account that modules drift away from pure spaces as was supposed, `mod-space` operation makes sense. I wonder if it has to have side effects, though. Do we need to load the module if it is not loaded, or would it be better to return `Empty`?

luketpeterson:
> Taking into account that modules drift away from pure spaces as was supposed...

The biggest conceptual issue here is the Tokenizer.  There are other minor things, but if we can come up with a way the Tokenizer leverages the space then I am sure we can solve the rest of the issues.  https://github.com/trueagi-io/hyperon-experimental/issues/510  Of course the Alpha release timeline might constrain us a bit.

> Do we need to load the module if it is not loaded, or would it be better to return Empty?

Not sure.  My assumption was that load-on-demand would be more convenient.  Otherwise people would need to use `register-module!` or `import!` prior.  Which might be inconvenient.  Especially because `import!` does additional stuff (e.g. registering tokens) and `register-module!` works with paths and therefore is limited to modules in the file system.

Necr0x0Der:
> My assumption was that load-on-demand would be more convenient.

Maybe, this functionality should be covered by a unit test.

luketpeterson:
> Maybe, this functionality should be covered by a unit test.

The Op Atom generally has this unit test:  https://github.com/luketpeterson/hyperon-experimental/blob/b686aa950964c089621a6ad5a60ea46e2d2cc799/lib/src/metta/runner/stdlib.rs#L1527

as far as a unit test specifically for load-on-demand, it calls through exactly the same pathway as `import!`, so I don't think there is much untested surface area in that particular interaction.

PR #617, opened 2024-03-12
Adding `add-reduct` operation, to merge atoms from a src space into a dst

luketpeterson:
This provides a friendlier way to accomplish the requirements of https://github.com/trueagi-io/hyperon-experimental/issues/592, getting the space access op for the src space to reduce before adding it to the dst space

### DIFF
--- a/lib/src/metta/runner/stdlib.metta
+++ b/lib/src/metta/runner/stdlib.metta
@@ -268,7 +268,7 @@
     (let $pattern $atom (let* $tail $template))
     $template )))
 
-(: add-reduct (-> Grounded %Undefined% ()))
+(: add-reduct (-> Grounded %Undefined% (->)))
 (= (add-reduct $dst $atom)  (add-atom $dst $atom))
 
 (: car-atom (-> Expression Atom))
--- a/lib/src/metta/runner/stdlib.metta
+++ b/lib/src/metta/runner/stdlib.metta
@@ -268,9 +268,8 @@
     (let $pattern $atom (let* $tail $template))
     $template )))
 
-(: merge-space (-> Atom %Undefined% Atom))
-(= (merge-space $dst $space)
-  (let $x $space (add-atom $dst $x)))
+(: add-reduct (-> Grounded %Undefined% ()))
+(= (add-reduct $dst $atom)  (add-atom $dst $atom))
 
 (: car-atom (-> Expression Atom))
 (= (car-atom $atom)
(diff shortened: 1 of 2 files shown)


luketpeterson:
Upon more reflection, this isn't limited to src spaces.  So we could name it something more general like `into-space`

What do you think?

lib/src/metta/runner/stdlib.rs
```diff
@@ -1466,6 +1466,7 @@ mod non_minimal_only_stdlib {
         (= (if False $then $else) $else)
         (: Error (-> Atom Atom ErrorType))
 
+        (= (merge-space $dst $space)  (let $x $space (add-atom $dst $x)))
```
vsbogd, line 1469: Should work as well because argument is reduced before call (same for minimal MeTTa):
```suggestion
        (= (merge-space $dst $space)  (add-atom $dst $space))
```

vsbogd:
Yes, it is more general than "merge", may be `add-reduced-atom` or `reduce-and-add-atom`?

lib/src/metta/runner/stdlib.metta
```diff
@@ -268,6 +268,10 @@
     (let $pattern $atom (let* $tail $template))
     $template )))
 
+(: merge-space (-> Atom %Undefined% Atom))
```
Necr0x0Der, line 271: `add-atom` returns `unit`, so this operation should also have `Unit` as its return type.

Necr0x0Der: And shouldn't the first argument be `Space`?

Necr0x0Der:
Regarding naming, I definitely agree that this operation should not be called `merge` even if it is about spaces (we don't merge spaces, but insert one into another). And it is definitely not about spaces only. As for naming, I don't have strong preferences, but I propose `add-reduct` (which is similar to `add-atom` and also indicates the difference).

lib/src/metta/runner/stdlib.rs
```diff
@@ -1466,6 +1466,8 @@ mod non_minimal_only_stdlib {
         (= (if False $then $else) $else)
         (: Error (-> Atom Atom ErrorType))
 
+        (: add-reduct (-> Grounded %Undefined% ()))
```
vsbogd, line 1469:
```suggestion
        (: add-reduct (-> Grounded %Undefined% (->)))
```

PR #625, opened 2024-03-16
Add pragma to turn off library interpreter written in minimal MeTTa

vsbogd:
Executing `!(pragma! interpreter bare-minimal)` turns off interpreter written in minimal MeTTa and only minimal MeTTa interpreter is used.

I used "interpreter" keyword because I would like to try implementing switchable interpreter.

### DIFF
--- a/lib/src/metta/runner/mod.rs
+++ b/lib/src/metta/runner/mod.rs
@@ -432,7 +432,11 @@ impl Metta {
     // to the minimal MeTTa
     pub fn evaluate_atom(&self, atom: Atom) -> Result<Vec<Atom>, String> {
         #[cfg(feature = "minimal")]
-        let atom = wrap_atom_by_metta_interpreter(self.0.top_mod_space.clone(), atom);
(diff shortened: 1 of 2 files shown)


luketpeterson:
Right now, pragma settings are global across a whole runner.  But is this something we want to change on a single-module level?

vsbogd:
My understanding right now it is required more for experimenting with code. Generally it would be great to have ability switching interpreter on the module level.

Necr0x0Der (approved):

luketpeterson:
If you want to merge this to make progress elsewhere or unblock other work, that's fine.

At some future point, I think we should decide out how settings work.  i.e. should we try an unify the settings-like mechanisms like PkgInfo (module-level), Environment (runner-level), pragmas (runner-level), REPL settings (A space created by repl app) etc.

Fundamentally some settings have different scopes and different mutability (and different mutability at different scopes!).  But maybe we can define a mechanism to capture that.

PR #627, opened 2024-03-18
Update requirements.txt

CICS-Oleg:


Necr0x0Der:
I'm not sure in general that these requirements should be in `python` folder of hyperon in general since they are needed only for das-gate

vsbogd (changes requested):
Current list of requirements is kept in `pyproject.toml` file (see `project.dependencies` setting). If we keep das-gate as a part of the `hyperon` Python package then we should keep requirements there. Otherwise we need to distribute it as a separate Python package.

Necr0x0Der:
DAS is invoked as MeTTa extension/module. While it is placed in hyperon package for convenience, putting its requirements into hyperon requirements is not good I think

vsbogd:
> DAS is invoked as MeTTa extension/module. While it is placed in hyperon package for convenience, putting its requirements into hyperon requirements is not good I think

Let's put requirements into a separate file in `python/hyperon/exts/das_gate` directory as discussed.

python/requirements.txt
vsbogd, line 1: Please remove this file it doesn't make sense. All dependencies are kept in `pyproject.toml` file.

vsbogd (approved):

PR #629, opened 2024-03-19
hot temporary fix for dasgate.py

CICS-Oleg:


DaddyWesker (approved):
Both test_das and test_bio_das works for me.

Necr0x0Der (approved):

Issue #321, opened 2023-06-06
MeTTa libraries documentation


Necr0x0Der:
This is a subissue of #319 .

The goal is to introduce MeTTa code documentation machinery for MeTTa code in general, libraries in particular, and `stdlib` especially.

The proposal is to document MeTTa code in MeTTa itself. There are a few choices to be made:

- In what format to place such pieces of documentation: doc-strings, doc-expressions with certain type structure, or something else?
- Where exactly to place them: the library space itself, a global `&doc` space, a separate doc-space for each library?
- Where MeTTa code for documentation is formed? Should `register_token` accept it as an additional argument? Should the space being imported just contain doc-expressions as stand-alone expressions similar to type definitions, or should these expressions be somehow places inside equalities like Python docstrings?
- How should documentation (+type declarations) be automatically gathered and can missing documentation be automatically detected?

There are a few issues that may prevent us from answering all these questions atm:

- we don’t have a notion of “known tokens” both for pure symbols and tokens for grounded atoms. Pure symbols don’t require obligatory definition to be used. While typing expressions can be used for “detecting” known symbols, even typing expressions can be tricky, e.g. `(: ($t A) (B $t))` is a valid expression. All mentioned symbols can be extracted from an entire space, though. Tokens for grounded atoms are regular expressions. Apparently, we cannot document each concrete instance, e.g., each number. The situation may partly change if we introduce bindings as a part of spaces.
- We have libraries, imports and py-extensions. We still don’t have a strict way to define new libraries. As for `stdlib`, we can either put doc-expressions in its space or do anything else we want. But will “anything else” work for custom libraries? For the case of importing pure MeTTa scripts, we have no choice but to put doc-expressions into scripts themselves, although it doesn’t necessarily mean that these expressions will remain in the same space. E.g., documenting can be done via import-time execution (that is, with something like `! (doc …)`). Also, documentation for libraries and for pure MeTTa code may not necessarily be done identically.

Let’s consider some examples, of how it could look like.

1) Doc-expressions are mere expressions over strings put arbitrarily into spaces, e.g.
```
(: doc (-> Atom String Atom))
(doc doc “Document any atom or expression”)
(doc String “The type for grounded strings”)
(doc + “Grounded operation for summing two numbers”)
```
[...]

Issue #354, opened 2023-06-30
case cannot be used with function calls defined out of the case metta module scope


vsbogd:
When `case` calls the passed function in the imported module, for example (`synthesize.metta`):
```
;; Define synthesize
(: synthesize (-> Atom (-> Atom) Atom))
(= (synthesize $query $axiom) (case ($axiom) (($query $query))))
```
and it is called from importing module, where function is passed as an argument (`bug.metta`):
```
;; Import synthesize
!(import! &self synthesize.metta)

;; Define axiom
(: axiom (-> Atom))
(= (axiom) (: f (-> Number String)))

;; Test synthesize
!(synthesize (: $term $type) axiom)
```
`case` cannot call the passed function because this function is defined in the outer module while `case` runs interpreter with space of the module where the call of the `case` is defined.

Expected result: `[(: f (-> Number String))]`
Actual result: `[]`

vsbogd:
As the space into which pure function is passed as an atom doesn't import the space where the function is defined we cannot execute pure function in it properly. In order to execute the passed pure function we need to have a reference to the space where the function is defined.

Usually this issue is not visible because top-level interpreter works on top-level space which imports all other spaces and thus search of the atom in top-level space brings all needed results. In case from the issue description the search starts from the top level space from which `case` was imported i.e. from `synthesize.metta`'s space. This space doesn't contain `axiom`'s definition and thus search fails.

luketpeterson:
The current behavior is essentially correct for `import` because modules are expected to be self-contained.  It would be impossible to validate a module if its environment could be affected implcitly by items from the client.

This situation is where an `include` behavior is desirable, to introduce additional code into the same running context.

vsbogd:
> This situation is where an include behavior is desirable, to introduce additional code into the same running context.

I am not sure if you mean this but `include` doesn't help in most of the cases. Consider the situation when module `a` passes its function into a module `b` and module `b` forwards it to the module `c`. Then module `a` cannot know in advance it should include something to use module `b`. Thus we rather need some referencing from atom to module where it is defined.

luketpeterson:
I don't follow what you are saying...  The term "pass into" is confusing me because modules don't explicitly declare any export directives.  So any item used must be imported or included.

If we have something like this:
A.metta
```
(= (synthesize $query $axiom) (case ($axiom) (($query $query))))
```
B.metta
```
!(include A)
```
C.metta
```
!(include B)
[...]

PR #640, opened 2024-03-26
Add doc strings related functions

vsbogd:
Work on #321. 

This PR is to review:
- doc strings format
- format of the `doc-formal` atom returned by `get-docs` function
- types of the doc strings and `doc-formal` constructors

If items above are approved next steps are:
- add unit tests
- add docs for a rest of the standard library
- add docs into a minimal MeTTa library

What is done

Add doc strings to `if`, `add-atom`, `ErrorType` standard library functions as example. Add `get-doc` which compiles documentation as a `doc-formal` atom from doc strings and type information. Add `help!` to print `doc-formal` in a human friendly way. Add `format-args` function to format strings inside `help!`. Add quick implementation for a `String` type in order to remove double quotes around printed strings.


vsbogd:
`format-args` implementation is also topic for a discussion. I used `dyn-fmt` library because `strfmt` requires named argument positions inside format string. We can do it using something like:
`(format-args "{first} {second}" ((first <first-atom>) (second <second-atom>)))`
for now I used arguments without names:
`(format-args "{} {}" (<first-atom> <second-atom>))`

Necr0x0Der (dismissed):
I'm OK with this approach. Not sure if descriptions themselves should be lengthier, but this can be changed later.

luketpeterson (dismissed):
This looks nice.

I see you went with dyn-fmt instead of strfmt.  I mainly chose strfmt in the past because of its momentum.  I didn't like how strfmt used a HashMap in its "simple" API, but it did allow you to implement a custom accessor function.

Looking over dyn-fmt, however, I don't see support for named arguments (which might be hard to support in MeTTa anyway given the need to query the bindings)  Slightly more disturbing is that it doesn't seem to support format specifiers.

For example: `let test = "Hex Value = {:x}".format(&[0xFFFF]);` renders as "Hex Value = :x"

So merging this is fine.  But I may replace the dyn-fmt crate in the near future unless you object.

vsbogd:
> So merging this is fine. But I may replace the dyn-fmt crate in the near future unless you object.

Yeah, I used `dyn-fmt` because I didn't want to construct a hash map and assign names to the arguments on the first place. But I don't think `dyn-fmt` is a best approach. I absolutely open for any good replacements, including `strfmt` if it supports all necessary modes.

vsbogd:
One relatively minor question: Should we replace long names: "description", "parameters", "parameter" by their shorter versions: "descr", "params", "param"? Or should we hope that auto-completion will make typing documentation easier with long names?

luketpeterson:
> 
One relatively minor question: Should we replace long names: "description", "parameters", "parameter" by their shorter versions: "descr", "params", "param"? Or should we hope that auto-completion will make typing documentation easier with long names?

Personally I feel like 4-7 characters is the sweet spot for these kind of names.  Long enough to be visually distinct, short enough not to waste space on the screen & elsewhere.  So "descr", "params", etc. is where I vote.

Also, unless you feel differently, I think it would be good to adopt the keywords from another system where our keywords do the same thing.  We obviously don't need to copy every feature, but where we implement the same feature, we could use the same keyword.

For example, doxygen's are here. https://www.doxygen.nl/manual/commands.html  But I'd be just as happy with any doc system.  I don't especially love doxygen.

vsbogd:
> Also, unless you feel differently, I think it would be good to adopt the keywords from another system where our keywords do the same thing.

Yeah, I agree with this, but I tried to copy names from metta-motto examples provided by @Necr0x0Der. @Necr0x0Der , do you mind some renaming? Which names do you prefer?

luketpeterson:
I actually don't feel strongly.  I just thought we could leverage the documentation and expectation of others if we copied their names.  But it's not a big deal at all.

Necr0x0Der:
> do you mind some renaming? 

I don't have strong opinion here

vsbogd:
Not merging it because after writing unit tests I have found the issue explained in https://github.com/trueagi-io/hyperon-experimental/issues/354#issuecomment-2024809298 I am trying to figure out how to resolve it properly.

luketpeterson:
From #354
> I am thinking about adding `&top` to the stdlib...

With the fix in https://github.com/trueagi-io/hyperon-experimental/pull/643 you can access `!(mod-space top)` without changing the stdlib. I know that's not a fix for #354, but slightly more convenient than adding to stdlib.

I'm thinking about the best solution for #354 in parallel.

vsbogd:
I added `get-type-space` operation see commit https://github.com/trueagi-io/hyperon-experimental/pull/640/commits/06d0901ddbd6441b2e04c89f7441f37562d958b2 and used it to get type using top level space.

Always using top level space in `get-type` doesn't work for the scripts like `g1_gadt.metta` because they are executed using `metta.load_module_at_path` (see `test_run_metta.py`) and during this execution top space doesn't include current working space. Thus `get-type` cannot find the type of the atom if type is defined in a space being loaded.

Commits https://github.com/trueagi-io/hyperon-experimental/pull/640/commits/72faef58f2f081b3cf7609481b2cb4697975b466 and https://github.com/trueagi-io/hyperon-experimental/pull/640/commits/06d0901ddbd6441b2e04c89f7441f37562d958b2 are used to implement unit tests in https://github.com/trueagi-io/hyperon-experimental/pull/640/commits/aee5eea650be9ee2042f3dea5dc645ec00d83b6a

vsbogd:
Commit https://github.com/trueagi-io/hyperon-experimental/pull/640/commits/7f0a4aeb7841c6b16b8c390eef083d157d3e560c adds doc related functions into a minimal MeTTa stdlib.

vsbogd:
CI failed because one of the workflow modules was unavailable.

Necr0x0Der (dismissed):

vsbogd:
Last problem with this PR is that `return` is actually internal symbol in minimal MeTTa. This means if it type-checks than it is evaluated. Thus having `(doc ... (return <something>))` forces the evaluation of the `(return <something>)`.

On the one hand this is a problem of aliasing. We have two different returns `docs:return` and `minimal:return` and we could use aliases to distinguish them (for instance using "import as" from `docs` modules). But as we don't have aliasing right now we need another solution.

One possible solution is to rename "minimal return" or "docs return". For example use short form "ret" for documentation. Or add a prefix like `@return` to the documentation tags. 

Another possible solution still use `return` but use `Atom` instead of `DocReturn` type inside `doc` and `doc-formal` types. 

I like the renaming option more because I feel these two `return` are different symbols effectively. @Necr0x0Der @luketpeterson what do you think?

Necr0x0Der:
`ret` or `@return` sounds good to me

luketpeterson:
> ret or @return sounds good to me

No strong opinion on the symbol.  Although `@return` is distinct in a way that we should probably maintain across the other doc-associated keywords/ symbols.  So my vote is for consistency.

> On the one hand this is a problem of aliasing. We have two different returns `docs:return` and `minimal:return` and we could use aliases to distinguish them (for instance using "import as" from `docs` modules). But as we don't have aliasing right now we need another solution.

It's an interesting idea to extend the symbol space into the module space.  I certainly thought this is a direction we might want to go.  And we absolutely could.  But currently we don't have anything like this.  It would involve making ':' a reserved character in symbols to avoid ambiguity.

luketpeterson (approved):

vsbogd:
> It would involve making ':' a reserved character in symbols to avoid ambiguity.

JIC, I used `:` only as an example, we can use anything appropriate. 

> Although @return is distinct in a way that we should probably maintain across the other doc-associated keywords/ symbols.

Yeah, I think it is worse to use `@` prefix for all doc symbols then.

luketpeterson:
> > It would involve making ':' a reserved character in symbols to avoid ambiguity.
> 
> JIC, I used `:` only as an example, we can use anything appropriate.
> 

':' would make sense since it's the separator module names use.  So it would be consistent.  But adding "full-path-symbols" a pretty big change to introduce before Alpha because it would likely need interpreter coordination to implement.

vsbogd:
Implemented showing docs for the functions without explicit types https://github.com/trueagi-io/hyperon-experimental/pull/640/commits/21faf744df4459c336c3f76eaf621b93dda89206. Renamed all doc tags to the versions with `@` prefix https://github.com/trueagi-io/hyperon-experimental/pull/640/commits/3fdd12c0385823dc9ab5bd0121f5635f9089020a. Doc tags has very common wording thus I think it is better to make them unique in order to eliminate possible conflicts.

vsbogd:
I don't see any other functional issues with this PR. One non-functional issue is that unit test for `get-doc` takes a lot of time for the minimal interpreter version.

vsbogd:
Ignored tests in https://github.com/trueagi-io/hyperon-experimental/pull/640/commits/1d41ab1190d16f3b52359abf14f3c0b170ff8e08 raised https://github.com/trueagi-io/hyperon-experimental/issues/660 to fix it.

Issue #466, opened 2023-10-20
Need accessor for loaded module's space, within MeTTa


vsbogd:
When I run `f1_imports.metta` manually I get the following error:
```
$ metta tests/scripts/f1_imports.metta 
[(Error (assertEqual ((let $x (get-atoms GroundingSpace-0x55f3df6e5278) (get-type $x))) ((get-type GroundingSpace-0x55f3df6e5278))) 
Expected: [(hyperon::space::DynSpace)]
Got: [(hyperon::space::DynSpace), (hyperon::space::DynSpace)]
Excessive result: (hyperon::space::DynSpace))]
```
`!get-atoms(&self)` returns `[GroundingSpace-0x55abbf778a08, GroundingSpace-0x55abbf8c9f58]` while `&self` is `GroundingSpace-0x55abbf7ab278` thus `&self` contains imported modules atomspaces which is expected but doesn't contain `&self`.

luketpeterson:
The issue is that running the script manually initializes the environment using the default `init.metta` file while the test environment does not contain any `init.metta` file at all.

An updated version of the test is here: https://github.com/luketpeterson/hyperon-experimental/blob/b504aeb1de476c13d0f9fbf4e9d7112088cca939/python/tests/scripts/f1_imports.metta

But this modified test fails in the test environment.

Harking back to the discussion here: https://github.com/trueagi-io/hyperon-experimental/pull/450#discussion_r1342803339 , the deeper issue is that some aspects of this test are testing an incidental behavior rather than a fundamental behavior.

In my opinion, a better test would bind a name to the loaded module, and then make sure that module was successfully loaded into the space by retrieving it by name, rather than taking for granted that the space would contain exactly N atoms and that the ith atom in a `get-atoms` list would be a specific atom.  (The current test is also sensitive to atom ordering within the space, which is also a problem)

vsbogd:
Thanks Luke, my description is incorrect. I was confused by the fact the test fails in minimal MeTTa on a different line than it fails when I run it using REPL. Thus you are right that it is an issue of the test itself.

Necr0x0Der:
[...]

PR #641, opened 2024-03-28
Fixing misleading obsolete comment in f1_imports.metta test

luketpeterson:
Tiny PR.  So we can close https://github.com/trueagi-io/hyperon-experimental/issues/466


vsbogd (approved):
Thanks for fixing it Luke!

PR #649, opened 2024-04-03
Adding module resource API and `include` operation

luketpeterson:
This PR adds the `include` operation and associated changes so module loaders can make additional named resources available through a resource API.

I also included a fix for an occasional panic caused by the C API being misused, passing a NULL ptr to a function taking an array ptr, when a zero-sized array is desired.

### DIFF
--- a/c/src/atom.rs
+++ b/c/src/atom.rs
@@ -882,7 +882,7 @@ pub unsafe extern "C" fn atom_vec_from_array(atoms: *mut atom_t, size: usize) ->
         if size == 0 {
             return vec![].into()
         } else {
-            panic!();
+            panic!("Null pointer is passed to the constructor of the vector with non-zero size");
         }
     }
     let c_arr = std::slice::from_raw_parts_mut(atoms, size);
--- a/c/src/atom.rs
+++ b/c/src/atom.rs
@@ -213,7 +213,7 @@ pub unsafe extern "C" fn atom_expr(children: *mut atom_t, size: usize) -> atom_t
         if size == 0 {
             return Atom::expr(vec![]).into()
         } else {
-            panic!();
+            panic!("Null pointer is passed to the constructor of the expression with non-zero size");
         }
     }
     let c_arr = std::slice::from_raw_parts_mut(children, size);
--- a/c/src/atom.rs
+++ b/c/src/atom.rs
@@ -209,6 +209,13 @@ pub unsafe extern "C" fn atom_sym(name: *const c_char) -> atom_t {
 ///
 #[no_mangle]
 pub unsafe extern "C" fn atom_expr(children: *mut atom_t, size: usize) -> atom_t {
+    if children == core::ptr::null_mut() {
+        if size == 0 {
+            return Atom::expr(vec![]).into()
+        } else {
+            panic!();
+        }
+    }
     let c_arr = std::slice::from_raw_parts_mut(children, size);
     let children: Vec<Atom> = c_arr.into_iter().map(|atom| {
         core::mem::replace(atom, atom_t::null()).into_inner()
@@ -871,6 +878,13 @@ pub extern "C" fn atom_vec_clone(vec: *const atom_vec_t) -> atom_vec_t {
 ///
 #[no_mangle]
 pub unsafe extern "C" fn atom_vec_from_array(atoms: *mut atom_t, size: usize) -> atom_vec_t {
+    if atoms == core::ptr::null_mut() {
+        if size == 0 {
+            return vec![].into()
+        } else {
+            panic!();
+        }
+    }
     let c_arr = std::slice::from_raw_parts_mut(atoms, size);
     let atoms: Vec<Atom> = c_arr.into_iter().map(|atom| {
         core::mem::replace(atom, atom_t::null()).into_inner()
--- a/lib/src/metta/runner/mod.rs
+++ b/lib/src/metta/runner/mod.rs
@@ -873,15 +873,16 @@ impl<'input> RunContext<'_, '_, '_, 'input> {
 
         // Resolve the module name into a loader object using the resolution logic in the pkg_info
         #[cfg(feature = "pkg_mgmt")]
-        self.load_module_recursive(mod_name, true)
+        {
+            let parent_mod_id = self.load_module_parents(mod_name)?;
+            self.load_module_internal(mod_name, parent_mod_id)
+        }
     }
 
     /// Internal function used for recursive loading of parent modules by [Self::load_module]
-    /// Calling with `load_target == false` will ensure a module's parents are loaded without
-    /// loading the module itself, and the module's parent's [ModId] will be returned.  Passing
-    /// `load_target == true` will load the module and return its [ModId]
+    /// Returns the ModId of the loaded parent module
     #[cfg(feature = "pkg_mgmt")]
-    fn load_module_recursive(&mut self, mod_name: &str, load_target: bool) -> Result<ModId, String> {
+    fn load_module_parents(&mut self, mod_name: &str) -> Result<ModId, String> {
 
         //Normalize the path in the context of this running module
         let normalized_mod_path = self.normalize_module_name(mod_name)?;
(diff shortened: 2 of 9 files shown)


lib/src/metta/runner/modules/mod.rs
```diff
@@ -373,10 +389,26 @@ impl MettaMod {
...
-    /// A function to load the module my making MeTTa API calls.  This function will be called by
+    /// A function to load the module my making MeTTa API calls.  This function will be called
     /// as a downstream consequence of [Metta::load_module_at_path], [Metta::load_module_direct],
     /// [RunContext::load_module], or any other method that leads to the loading of modules
     fn load(&self, context: &mut RunContext) -> Result<(), String>;
+
+    /// Returns a data blob containing a given named resource belonging to a module
+    ///
+    /// The following values for `res_key` are commonly accepted:
+    /// * `module.metta`: The MeTTa code for the module in S-Expression format, if the module is
```
vsbogd, line 400: 1. I would suggest use name which doesn't look like a filename for the main code of the module. For instance we can call it `text`. The reason is that there can be different modules: ones which load source code from other files, ones which load source code from standard input. And I don't see why this main resource should have a name which is equal to the name of the file in one particular module format.

2. I see in code `"module.metta"` is frequently used as a string. I would suggest adding a constant for it. May be even some enum to keep all standard keys and allow non-standard ones:
```rs
enum ResKey<'a> {
  Text,
  Authors,
  Description,
  Other(&'a str),
}
```

lib/src/metta/runner/mod.rs
```diff
@@ -396,6 +397,12 @@ impl Metta {
         modules.get(mod_id.0).unwrap().space().clone()
     }
 
+    /// Returns a buffer containing the specified resource, if it is available from a loaded module
+    pub fn get_module_resource(&self, mod_id: ModId, res_key: &str) -> Result<Vec<u8>, String> {
+        let modules = self.0.modules.lock().unwrap();
+        modules.get(mod_id.0).unwrap().get_resource(res_key).clone()
```
vsbogd, line 403: Cloning is redundant here:
```suggestion
        modules.get(mod_id.0).unwrap().get_resource(res_key)
```

lib/src/metta/runner/mod.rs
```diff
@@ -819,12 +825,15 @@ impl<'input> RunContext<'_, '_, 'input> {
...
+        self.load_module_recursive(mod_name, true)
     }
 
     /// Internal function used for recursive loading of parent modules by [Self::load_module]
+    /// Calling with `load_target == false` will ensure a module's parents are loaded without
+    /// loading the module itself, and the module's parent's [ModId] will be returned.  Passing
+    /// `load_target == true` will load the module and return its [ModId]
     #[cfg(feature = "pkg_mgmt")]
-    fn load_module_recursive(&mut self, mod_name: &str) -> Result<ModId, String> {
+    fn load_module_recursive(&mut self, mod_name: &str, load_target: bool) -> Result<ModId, String> {
```
vsbogd, line 836: Adding `load_target` flag is probably ok while method is private, but I would replace call with `false` flag by pair of calls: `load_module(get_parent_mod_name(mod_name))`. Thus the code clearly states that parent module is loaded and explanation of the `load_target` semantics is not needed.

vsbogd: At least I would rename flat into `load_parent` as it is exactly its semantics as far as I understand.

lib/src/metta/runner/mod.rs
```diff
@@ -856,6 +870,39 @@ impl<'input> RunContext<'_, '_, 'input> {
...
+    /// NOTE: Although this method won't load the module itself, it will load parent modules if necessary
+    pub fn load_resource_from_module(&mut self, mod_name: &str, res_key: &str) -> Result<Vec<u8>, String> {
+
+        // LP-TODO-NOW!, We should assume mod_name is relative by default, not absolute
+
+        // Resolve the module name and see if the module is already loaded into the runner
+        if let Ok(mod_id) = self.get_module_by_name(mod_name) {
+            self.metta().get_module_resource(mod_id, res_key)
+        } else {
+            #[cfg(not(feature = "pkg_mgmt"))]
```
vsbogd, line 885: While we have this feature, the code doesn't compile if feature is removed from list of features. Thus I am not sure we need it.

lib/src/metta/runner/stdlib.rs
```diff
@@ -147,6 +147,63 @@ fn strip_quotes(src: &str) -> &str {
...
+    fn execute(&self, args: &[Atom]) -> Result<Vec<Atom>, ExecError> {
+        let arg_error = || ExecError::from("include expects a module name argument");
+        let mod_name_atom = args.get(0).ok_or_else(arg_error)?;
+
+        // TODO: replace Symbol by grounded String?
+        let mod_name = match mod_name_atom {
+            Atom::Symbol(mod_name) => mod_name.name(),
+            _ => return Err(arg_error())
+        };
+        let mod_name = strip_quotes(mod_name);
```
vsbogd, line 186: I am not sure why do we need strip quotes here. Symbol name should not contain quotes AFAIK. Is it needed for some specific usecase?

lib/src/metta/runner/stdlib.rs
```diff
@@ -147,6 +147,63 @@ fn strip_quotes(src: &str) -> &str {
...
+        //TODO: Remove this hack to access the RunContext, when it's part of the arguments to `execute`
+        let ctx_ref = self.context.lock().unwrap().last().unwrap().clone();
+        let mut context = ctx_ref.lock().unwrap();
+        let program_buf = context.load_resource_from_module(mod_name, "module.metta")?;
+
+        // Interpret the loaded MeTTa S-Expression text
+        let program_text = String::from_utf8(program_buf)
+            .map_err(|e| e.to_string())?;
+        let parser = crate::metta::text::OwnedSExprParser::new(program_text);
+        context.push_parser(Box::new(parser));
```
vsbogd, line 197: Does it mean included module is added to the end of the including script?

c/src/atom.rs
```diff
@@ -871,6 +878,13 @@ pub extern "C" fn atom_vec_clone(vec: *const atom_vec_t) -> atom_vec_t {
 ///
 #[no_mangle]
 pub unsafe extern "C" fn atom_vec_from_array(atoms: *mut atom_t, size: usize) -> atom_vec_t {
+    if atoms == core::ptr::null_mut() {
+        if size == 0 {
+            return vec![].into()
+        } else {
+            panic!();
```
vsbogd, line 885:
```suggestion
            panic!("Null pointer is passed to the constructor of the vector with non-zero size");
```

lib/src/metta/runner/modules/mod.rs
luketpeterson: > I would suggest use name which doesn't look like a filename...

I think you're right.  When I conceived of the idea, I thought of them as files in a virtualized file-system, so it's not an accident they feel like file names.  In fact, my implementation started as a simple shim over files in the resources directory, but I felt that might be insecure so I locked it down more.

I now see that making the keys feel like files is going to lead to confusion when people think they can use any filename and it doesn't work.

> I would suggest adding a constant for it. May be even some enum to keep all standard keys and allow non-standard ones

That's a good idea.

lib/src/metta/runner/mod.rs
luketpeterson: Good catch!

I had an API that initially returned a borrowed value, then I changed it but didn't change the clone.  I think the compiler would have optimized it away, but good catch.

lib/src/metta/runner/stdlib.rs
luketpeterson: > I am not sure why do we need strip quotes here.

I was following the logic of `import!` and `mod-space!` and didn't give it much thought this time.  But a couple weeks ago there was a bug caused by people (Adam ;-) quoting module names.

It's not necessary to quote module names / mod name paths because spaces are illegal there, but people have quotes in their code as a leftover from when they were file paths, and also when we transition to grounded strings, quotes will be ok again.

So my feeling is to leave this as it is.

lib/src/metta/runner/stdlib.rs
luketpeterson: That's a good point.  This enqueues new work into the RunContext.  But people probably expect the include to be inlined immediately at the site of the `include` operation...

lib/src/metta/runner/mod.rs
luketpeterson: Done in 79a24c6

lib/src/metta/runner/modules/mod.rs
luketpeterson: `ResourceKey` enum added in 3deba55

lib/src/metta/runner/stdlib.rs
```diff
@@ -194,7 +194,11 @@ impl Grounded for IncludeOp {
         let program_text = String::from_utf8(program_buf)
             .map_err(|e| e.to_string())?;
         let parser = crate::metta::text::OwnedSExprParser::new(program_text);
-        context.push_parser(Box::new(parser));
+        //QUESTION: do we want to do anything with the result from the `include` operation?
```
vsbogd, line 197: I would say it should behave like included file is a part of the modules itself. Thus user which includes module B from module A I think will expect the results are returned as they are results of the A. To make it behave absolutely identical we need to implement it differently. Included it on a level of parser I for example. Thus may be we can leave the code as is while we don't have a real usecase.

PR #652, opened 2024-04-06
Simple python import

astroseger:
Example of the implementation of simple but rather universal python import, which also adds support for keyword arguments for the call of python objects.  

Few notes:
* It is a simple implementation to demonstrate the idea.
* It should be implemented on the lower level to remove necessity of __unwrap. So it should be implemented on the basic level of integration of python objects. 
* Support for the call of methods for objects is done with the dirty “call_dot” operator. It should be done differently, for example fixing “call:” operator, or maybe better by simply parsing “.”


Necr0x0Der (approved):
Works well as a sandbox implementation, thanks

PR #657, opened 2024-04-10
Addition to simple python import

astroseger:
Small Addition to the sandbox simple_import example
It adds "import" without as or from, see `example_04_numpy_simple_import.metta`

- With such import it is more common that we could call import twice, but since we work on the level of tokenizer it does not cause any problem (see `example_04_numpy_simple_import.metta`).
- For calls like "numpy.random.rand" it adds a "dirty" call_dot2 operator

### DIFF
--- a/python/sandbox/simple_import/example_04_numpy_simple_import.metta
+++ b/python/sandbox/simple_import/example_04_numpy_simple_import.metta
@@ -1,0 +1,11 @@
+!(import! &self simple_import)
+
+!(import numpy)
+!(import numpy)
+!(import numpy.linalg)
+
+
+!(bind! m1 (call_dot2 numpy random rand 3 3 ))
+!(bind! m1_inv (call_dot2 numpy linalg inv m1))
+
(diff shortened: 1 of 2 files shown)


python/sandbox/simple_import/example_04_numpy_simple_import.metta
```diff
@@ -0,0 +1,11 @@
+!(import! &self simple_import)
+
+!(import numpy)
+!(import numpy)
```
Necr0x0Der, line 4: just wonder why are there two `!(import numpy)`?

astroseger: Only to demonstrate that it does not cause any problem.

astroseger: Everything, including numpy.linalg.inv works after import numpy. My point was that with simple import it is more common to import something twice, because of submodules in python. So I wanted to make sure it would not cause problems in this implementation. It doesn't.

Necr0x0Der (approved):

PR #663, opened 2024-04-12
add kwargs for python functions

Necr0x0Der:


astroseger (commented):
In my example Kwargs works. As well as import from the local directory.

astroseger (dismissed):

Issue #662, opened 2024-04-12
Python package version is determined incorrectly in repository forks as "0.1.devX+Y"


vsbogd:
**Describe the bug**

`cargo run metta` fails to start claiming Python hyperon package version is incorrect when being started from a local repo created from a fork of the main repository.

**To Reproduce**
1. Create a fork of `trueagi-io/hyperon-experimental` on GitHub
2. Clone fork locally `git clone <your-fork-uri>`
3. `cd <you-fork-local-dir>`
4. `git tag` should show empty list of tags (if the fork has no tags)
5.  Build app using "Manual installation" instruction in [README.md](https://github.com/trueagi-io/hyperon-experimental#manual-installation)
6. Install Python package from the root directory of the local repo using `python3 -m pip install ./python[dev]`
7. Run `python -c "import hyperon; print(hyperon.__version__)"`
8. Run `cargo run`

**Expected behavior**
`import hyperon; print(hyperon.__version__)` prints last version of the Python package for example `0.1.8.dev50+g1d41ab11.d20240412`. `cargo run` starts REPL.

**Actual behavior**
`import hyperon; print(hyperon.__version__)` prints v0.1 version like `0.1.dev2411+g65e26e9`.
`cargo run` fails with a message "Fatal Error: MeTTa repl requires HyperonPy version matching '>=0.1.0, <0.2.0'.  Found version: '0.1.dev2411+g65e26e9'"

vsbogd:
Root cause is that dynamic version logic defined in `pyproject.toml`: https://github.com/trueagi-io/hyperon-experimental/blob/07021cb73d1fbaba19a41c98599a981a644177ab/python/pyproject.toml#L22 uses tags to find out what was the latest release of the package. If in a fork repo no tags are present then current version is determined as `0.1` which causes incorrect behavior of the REPL.

vsbogd:
We already have the fallback version in `hyperon/__init__.py`:
https://github.com/trueagi-io/hyperon-experimental/blob/07021cb73d1fbaba19a41c98599a981a644177ab/python/hyperon/__init__.py#L15
Because some users doesn't use `pip install` at all and we still need to return a proper version. See discussions in #434 and #440.

Thus I would suggest adding the single place where current "base" version of the Python package would be defined. And then derive any "specific" version using local git commit or anything else.

vsbogd:
Three use-cases need to be supported:
[...]

PR #667, opened 2024-04-16
Issue 662 solution, hyperon version refactor

DaddyWesker:
Check issue https://github.com/trueagi-io/hyperon-experimental/issues/662

Code should be simple, but I wasn't been able to test it properly. In my environment this variant works:
`user installs Python package pip install -e hyperon[dev] and gets the version`

Also in a new environment I was able to make this variant work:
`user clones repo, builds code and adds ./python dir into PYTHONPATH, after that gets version using import hyperon; print(hyperon.__version__)`

Variant with cibuildwheel worked for me only once and then I wasn't been able to reproduce it. Maybe it is lack of experience of using this cibuildwheel. Anyway, if someone could check cibuildwheel path and get version of hyperon properly - it will be really great.


vsbogd:
1. I am sorry I didn't mentioned another important use-case when package is installed using PyPi repo or from distribution package file. In this case user doesn't even have `git` installed and code in PR doesn't work in this case. Something similar in fact should happen when using cibuildwheel because it copies source code (excluding git root) and releases package from the copy.

2. Calling `git` each time when package is loaded has two other issues:
- Package can be imported from any location in the filesystem thus at least git should be called from the git repo root dir.
- It is a bit overhead even for dev environment. I would not say it is absolutely not a way to go but I would try a solution when git is called only on the stage of the package building first. For instance current code calls git only at the build stage. Yes, we loose ability to assign a git tag to the package which is loaded via `PYTHONPATH` modification but: (a) we can call git as a last resort measure when version is not found (localbuild branch in current code) (b) it may not that important as it is a rare case.
- Git can be absent on a user machine (see item (1)) thus we need to check the call for errors.

3. Importing `__version__` from `hyperon` in `pyproject.toml` also has problems. Importing `hyperon` on a build stage may require dependency libraries which can be unavailable at the moment of the package building (see item 6 at the link https://packaging.python.org/en/latest/guides/single-sourcing-package-version/#single-sourcing-the-version). It is more theoretical issue for our project though.

vsbogd:
I would propose closing this PR or converting it into a "Draft PR" before new change is ready for review.

DaddyWesker:
I've converted PR to draft.

python/setup.py
```diff
@@ -1,11 +1,41 @@
...
+        return fp.read()
+
+def get_version(rel_path):
+    new_ver = None
+    with open(rel_path) as f:  lines = f.read().splitlines()
+    with open(rel_path, 'w') as f:
+        for line in lines:
+            if line.startswith('__version__'):
+                delim = '"' if '"' in line else "'"
+                new_ver = line.split(delim)[1].split("+")[0] + get_git_revision_short_hash()
```
vsbogd, line 32: You don't need to get `git` revision manually. It is already returned from `scmgit.parse` in `ScmVersion` instance as a `ScmVersion.node` field. You also has `ScmVersion.distance`.

python/setup.py
```diff
@@ -71,7 +100,17 @@ def _build_with_cmake(self, ext: CMakeExtension) -> None:
             ["cmake", "--build", ".", *build_args], cwd=build_temp, check=True
         )
 
+
+def parse(root, config):
```
vsbogd, line 104: Original parse also has `**kvargs` parameters: https://github.com/pypa/setuptools_scm/blob/d75201a03ea01bf2d99a2521b87cdea30320f887/src/setuptools_scm/git.py#L200-L205 which should be passed through.

python/hyperon/__version__.py
vsbogd, line 1: I am not sure why, but it seems canonical version of the name for this file is `_version.py`

python/setup.py
```diff
@@ -1,11 +1,41 @@
...
+
+def get_version(rel_path):
+    new_ver = None
+    with open(rel_path) as f:  lines = f.read().splitlines()
+    with open(rel_path, 'w') as f:
+        for line in lines:
+            if line.startswith('__version__'):
+                delim = '"' if '"' in line else "'"
+                new_ver = line.split(delim)[1].split("+")[0] + get_git_revision_short_hash()
+                f.write(line.replace(line.split(delim)[1], new_ver))
```
vsbogd, line 33: You can ask `setuptools_scm` to write version file using `version_file` and `version_file_template`.

python/setup.py
vsbogd: Actually function should implement the following protocol: https://github.com/pypa/setuptools_scm/blob/d75201a03ea01bf2d99a2521b87cdea30320f887/src/setuptools_scm/_config.py#L54-L57

python/setup.py
DaddyWesker: If scmgit.parse(root, config) returns None then this ScmVersion.node is also None. So If I left something like that:
```
ver = scmgit.parse(root, config)
    if (ver == '0.0') or (ver is None):
        version = get_version("hyperon/_version.py") + scmgit.ScmVersion.node
        return scm.version.meta(tag=version, config=config, preformatted=True, dirty=True)
    else:
        return ver
```
I'll get an error that  scmgit.ScmVersion.node is None since if is True if ver from scmgit.parse is None. Though probably you meant something else?

python/setup.py
```diff
@@ -1,11 +1,41 @@
...
+    try:
+        sha = subprocess.check_output(['git', 'rev-parse', '--short', 'HEAD']).decode('ascii').strip()
+        return f"+{sha}"
+    except Exception:
+        return ''
+
+def read(rel_path):
+    here = os.path.abspath(os.path.dirname(__file__))
+    with codecs.open(os.path.join(here, rel_path), 'r') as fp:
+        return fp.read()
```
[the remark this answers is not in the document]
DaddyWesker: It was used. I forgot to delete it.

python/pyproject.toml
```diff
@@ -19,7 +19,9 @@ classifiers = [
 dependencies = [
     'importlib-metadata==6.6.0; python_version<"3.8"',
```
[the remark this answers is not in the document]
DaddyWesker: Done.

python/hyperon/__version__.py
DaddyWesker: Ok. Done.

python/setup.py
DaddyWesker: So `def parse(root, *, config):` ?

vsbogd: Could you please check what this weird syntax does mean in Python and write a proper signature which passes all the parameters if you are still going to override `parse`.

python/setup.py
vsbogd: If version is `None` it means `setuptools_scm` didn't find git repo or git command, so you will not find the distance either. Thus I would suggest return bare version if `setuptools_scm` returned None. Even simpler way is to redefined `version_scheme`

python/setup.py
```diff
@@ -71,7 +84,14 @@ def _build_with_cmake(self, ext: CMakeExtension) -> None:
             ["cmake", "--build", ".", *build_args], cwd=build_temp, check=True
         )
 
+
+def version_scheme(version):
+    return get_version("hyperon/_version.py")
```
vsbogd, line 89: I provided this code as an example. For the real code one need to check whether `version` was detected correctly by checking if it is `None` or `0.0`. If it isn't detected correctly then use hard-coded version.

DaddyWesker: You mean?
```
def version_scheme(version):
    if (version is None) or (version == "0.0"):
        return get_version("hyperon/_version.py")
    else:
        return version
```

vsbogd: Taking my words back, in fact we don't want two sources for version: tags and file, it is better to have a single source and I would prefer it would be the file.

DaddyWesker:
Currently I'm testing latest changes. So probably something can be changed still.

python/pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools==65.6.3", "conan==1.62", "cmake==3.26.4", "setuptools_scm[toml]>=6.2"]
+requires = ["setuptools==65.6.3", "conan==1.62", "cmake==3.26.4", "setuptools_scm[toml]>=8.0.0"]
```
vsbogd, line 2: We should either remove support for the Python 3.7 or stick to the `setuptools_scm` version `7.1.0`.

DaddyWesker: Yes, I've already changed this line to 7.1.0. My bad.

DaddyWesker:
@vsbogd everything should be fine now. Please check if you agree with those changes.

vsbogd (approved):

Issue #642, opened 2024-03-28
Unifying a structure of variables with a structured function name is not disabled


ngeiswei:
## What is your problem?

Unifying a structure of variables with a structured function name is not disabled by the absence of `variable_operation`.

## How to reproduce your problem?

Run the following code

```
(= ((W $x) $y) (($x $y) $y))
!(($x $y) $y)
```

## What would you normally expect?

It should end with the output

```
[(($x $y) $y)]
```

## What do you get instead?

It gets trapped in an infinite recursion.

## What else do you have to say?

My assumption is that the absence of `variable_operation` should cover that case.  If not then I need another solution, maybe involving Minimal MeTTa or inference control.  To give more context I'm trying to emulate combinatory logic, `W` is an actual combinator.

The problems occurs in both old and Minimal MeTTa.

ngeiswei:
Any progress on that?  It is somewhat of a blocker.

vsbogd:
I raised #668 with the fix which prevents from evaluation any expression which has a variable on the first position of the nested expression on the first position. Despite the PR fixes this specific scenario I would say the overall semantics of the evaluation is not clear.

@ngeiswei could you please point to the specific code in https://github.com/trueagi-io/hyperon-pln/pull/44 which is blocked by this issue? I would like to understand which semantics do you need.

vsbogd:
@ngeiswei does the following type definition suits your needs and fixes the example?
```
(: λ (-> Variable Atom Atom))
```

ngeiswei:
I've tried the following

1. No type definition of λ
```
(= ((λ $x $f) $y) (let ($νx $νf) (sealed ($x) ($x $f)) (let $νx $y $νf)))
!((λ $n (λ $f (λ $x ($f (($n $f) $x))))) (λ $f (λ $x $x)))
```
outputs
```
[(λ $f (λ $x ($f $x))), (λ $f (λ $x ($f $x))), (λ $f (λ $x ($f $x))), (λ $f (λ $x ($f $x))), (λ $f (λ $x ($f $x))), (λ $f (λ $x ($f $x))), (λ $f (λ $x ($f $x))), (λ $f (λ $x ($f $x)))]
```
which is correct because `((λ $n (λ $f (λ $x ($f (($n $f) $x))))) (λ $f (λ $x $x)))` represents the successor of 0, which is 1, encoded by `(λ $f (λ $x ($f $x)))`.  It's just that it is duplicated 8 times.

[...]

Issue #235, opened 2023-01-13
Multiple type definitions cause multiple results


Necr0x0Der:
Related to #188 , but a cleaner problem. The code
```
(: inc (-> Number Number))
(= (inc $x) (+ $x 1))
!(inc 1)
```
works as expected, but
```
(: inc (-> Number Number))
(: inc (-> Number Number))
(= (inc $x) (+ $x 1))
!(inc 1)
```
produces 4 (!) identical results. With equality non-determinism, the number of results multiplies. The interpreter works correctly, if the argument type doesn't match the (non-deterministic) type signature, e.g.
```
(: inc (-> Number Number))
(: inc (-> String Number))
(= (inc $x) (+ $x 1))
!(inc 1)
```
produces only one result. We don't have a settled semantics for non-deterministic types that match one equality. For example,
```
(: inc (-> Number Number))
(: inc (-> Atom Number))
(= (inc $x) (+ $x 1))
!(inc 1)
```
produces 2 results. There can be tricky cases, when we do want to use same equality interpreting its arguments as being of different types, and one may claim that getting 2 results in the latter case is OK (to be discussed and decided, I'd say; maybe, it would still be preferable to use the equality once, but to keep track of different possible types of its argument). But why there are 4 results for the duplicated type definition case?

vsbogd:
One theoretical way to improve the situation is to not generate alternatives when required type is `%Undefined%` which means that any type works i.e. no type specifics is involved. In such case specific type doesn't add an information and alternative is just a duplicate. @Necr0x0Der what do you think?

ngeiswei:
I see, I can live with that for now, so if the duplicate issue is too difficult to fix now I'm happy to have that merged.

Necr0x0Der:
Can the use of type information be viewed not as generating alternatives, but filtering out them - not just for `%Undefined%`, but for all type signatures?

PR #668, opened 2024-04-16
Fix nested variable operations

vsbogd:
Quick fix for #642

### DIFF
--- a/lib/src/metta/interpreter_minimal.rs
+++ b/lib/src/metta/interpreter_minimal.rs
@@ -517,12 +517,16 @@ fn call_to_stack(call: Atom, vars: Variables, prev: Option<Rc<RefCell<Stack>>>) 
 #[cfg(not(feature = "variable_operation"))]
 fn is_variable_op(atom: &Atom) -> bool {
     match atom {
-        Atom::Expression(expr) => {
-            match expr.children().get(0) {
-                Some(Atom::Variable(_)) => true,
-                _ => false,
-            }
-        },
+        Atom::Expression(expr) => is_variable_op_expr(expr),
+        _ => false,
+    }
+}
+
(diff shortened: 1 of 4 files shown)


ngeiswei (changes requested):
Running

```
(= ((W $x) $y) (($x $y) $y))
!(($x $y) $y)
```

outputs

```
[(($x $y) $y), (($x $y) $y)]
```

instead of

```
[(($x $y) $y)]
```

Is it normal?

vsbogd:
Thanks for checking Nil, I am sorry for that. It is not normal and this happens because tuples of variables are matched with random types. In this case `($x $y)` tuple matches with `(: (@kind atom) ...)` and (: (@kind function) ...)` type, which causes two different interpretation to be returned.

vsbogd:
The effect above is raised as issue #235

ngeiswei (approved):

vsbogd:
I don't want to merge it now, I would like to finish discussion in #642 first. Please see possible solution https://github.com/trueagi-io/hyperon-experimental/issues/642#issuecomment-2061567989

vsbogd:
@ngeiswei , should be fixed now, please try it.

ngeiswei:
With the latest revision 23a22d28

```
(= ((W $x) $y) (($x $y) $y))
!(($x $y) $y)
```

still spawns MeTTa in a infinite recursion (with `variable_operation` disabled).

ngeiswei:
Oh, these are new commits, let me check then!

ngeiswei:
It behaves as expected on my end.

ngeiswei (approved):

ngeiswei:
It's awesome to see these old issues being knocked down one by one!  Thanks!

PR #680, opened 2024-04-26
Enable MacOSX Apple Silicon release

vsbogd:
Fix deprecation warnings from release job, opt out `musllinux` by default in `pyproject.toml`, fix issue with insufficient permissions when building using `macos-14` GitHub runner.

To install Python package for MacOSX Apple Silicon from test PyPi:
```
python3 -m pip install --index-url https://test.pypi.org/simple/ hyperon
```
Version `0.1.8.1` should be installed.

### DIFF
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -31,7 +31,7 @@ jobs:
 
     steps:
       - name: Check out repository code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Install Rust stable
         uses: actions-rs/toolchain@v1.0.6
@@ -57,7 +57,7 @@ jobs:
           args: cbindgen
 
       - name: Set up Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: ${{ inputs.python-version }}
 
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -9,8 +9,8 @@ jobs:
   deploy-docs:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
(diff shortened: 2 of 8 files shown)


vsbogd:
@Necr0x0Der already tried to install it on M1 so I am merging it.

luketpeterson:
One thing that's worth mentioning.  My changes moved the C lib tests from testing the static lib to testing the dynamic hyperonc lib.  Meaning I switched the `BUILD_SHARED_LIBS` argument to `ON`.  But that has no bearing on this change.

vsbogd:
> Meaning I switched the BUILD_SHARED_LIBS argument to ON in the install-hyperonc script. But that has no bearing on this change.

Actually it means we completely don't support `musllinux` platform, so the comment in https://github.com/trueagi-io/hyperon-experimental/blob/14bf9ee36467ee391f5b96d4bd61b3bda1e47af7/python/pyproject.toml#L39-L40 should be changed. @luketpeterson do you mind doing it in your branch then?

luketpeterson:
Do you think I should go back to testing the static lib?  My reason for switching the tests to the dynamic lib was a thought that the C users are likely to use the dynamic lib so it provides a cleaner test (closer to the user's config) and I figured it didn't harm anything because I thought we didn't support musl linux anyway.

But if there is a reason we might want to keep musl linux in the future, I can put the tests back to testing the static lib.

Otherwise I will do as you say and remove the comment entirely.

vsbogd:
> But if there is a reason we might want to keep musl linux in the future, I can put the tests back to testing the static lib.

I don't think supporting musl is critical, thus I am ok switching to the using of dynamic libraries in tests. Anyway if something will change we can return to this question. The only thing which bothers me is to have nice clue in comments to understand what exactly is not support on musl platform.

PR #682, opened 2024-05-02
Cleanup minimal MeTTa interpreter code especially code related to variables handing

vsbogd:
The code cleanup, mainly I localized code which works on variables handing in order to make it more optimal. This is work in progress but I would like to merge these changes because they have value on their own. It should be simpler to review each commit separately.


lib/src/metta/interpreter_minimal.rs
```diff
@@ -163,7 +165,7 @@ impl<'a, T: Space + 'a> SpaceRef<'a> for T {}
 #[derive(Debug)]
 struct InterpreterContext<'a, T: SpaceRef<'a>> {
     space: T,
-    phantom: PhantomData<&'a GroundingSpace>,
+    phantom: PhantomData<dyn SpaceRef<'a>>,
```
luketpeterson, line 168: This is outside the scope of this PR, but the lifetime on the InterpreterContext is unnecessary in minimal MeTTa.

luketpeterson (approved):
These changes all look good.  I don't feel like I have a deep enough familiarity with the code to predict all future implications, but I can see that the changes simplify the design by limiting the places that need to care about the Variables

lib/src/metta/interpreter_minimal.rs
vsbogd: Thanks Luke, done in https://github.com/trueagi-io/hyperon-experimental/pull/682/commits/cb63a1257c8d16de06594d5d659ad63c96981d4a

luketpeterson (approved):
I didn't mean I expected you to remove the extraneous lifetime in this PR.  I see it's there for compatibility with the old interpreter.  Sorry I communicated badly.

Your changes to raise it up one level are good because it gets us one step closer to being able to remove it.  So now that you have made the changes I think it makes sense to keep them.

PR #684, opened 2024-05-06
`include` op to return results from evaluated sub-script

luketpeterson:
This is a change requested by @TeamSPoon to match behavior of the old `import` op.  With this change, the `include` op will return the last result from evaluating the sub-script.


lib/src/metta/runner/stdlib.rs
```diff
@@ -195,13 +195,16 @@ impl Grounded for IncludeOp {
...
             context.push_parser(Box::new(parser));
             Ok(())
         })?;
 
-        unit_result()
+        //NOTE: Current behavior returns the result of the last sub-eval to match the old
+        // `import!` before before module isolation.  However that means the results prior to
+        // the last are dropped.  I don't know how to fix this or if it's even wrong, but it's
+        // different from the way "eval-type" APIs work when called from host code, e.g. Rust
+        Ok(eval_result.into_iter().last().unwrap_or_else(|| vec![]))
```
vsbogd, line 207: Well, yeah, as only last result is returned and other are dropped I am not sure if this really required change. On the other hand I don't object so we can merge it. @TeamSPoon does this change help?

vsbogd (approved):

vsbogd:
@luketpeterson , please change `type()` method as well.

luketpeterson:
Ah.  Oops.  Thanks for catching that.

PR #687, opened 2024-05-06
%void% -> Empty

Necr0x0Der:
I propose to get rid of `%void%` in `case` and replace it with `Empty`, since it is more consistent with minimal metta

### DIFF
--- a/lib/src/metta/runner/stdlib_minimal.rs
+++ b/lib/src/metta/runner/stdlib_minimal.rs
@@ -408,8 +408,6 @@ impl Grounded for CaseOp {
         log::debug!("CaseOp::execute: atom results: {:?}", results);
         let results = match results {
             Ok(results) if results.is_empty() =>
-                // TODO: MINIMAL in minimal MeTTa we should use Empty in both
-                // places here and in (case ...) calls in code
                 vec![switch(EMPTY_SYMBOL)],
             Ok(results) =>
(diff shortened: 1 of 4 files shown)


Necr0x0Der:
@vsbogd , please check that it is correct in terms of minimal metta stdlib

lib/src/metta/runner/stdlib_minimal.rs
```diff
@@ -412,7 +410,7 @@ impl Grounded for CaseOp {
             Ok(results) if results.is_empty() =>
                 // TODO: MINIMAL in minimal MeTTa we should use Empty in both
                 // places here and in (case ...) calls in code
```
vsbogd, line 412: Could be removed as you done it in this PR:
```suggestion
                // TODO: MINIMAL in minimal MeTTa we should use Empty in both
                // places here and in (case ...) calls in code
```

vsbogd (approved):

lib/src/metta/runner/stdlib_minimal.rs
vsbogd: @Necr0x0Der please commit it before merge, I cannot for some reason.

Necr0x0Der: I cannot either, because the suggestion consists in replacing the comment with the same comment (not removing the fragment). But I can do this locally

PR #693, opened 2024-05-17
Fix for dasgate: another way to parse link targets

CICS-Oleg:


Necr0x0Der (approved):

python/integration/test_bio_das.metta
```diff
@@ -17,6 +17,13 @@
 !(assertEqualToResult
     (match &das (translation_of (protein O43264) (transcript ENST00000200135)) OK)
     (OK))
+
+!(match &das (translation_of (protein $x) (transcript ENST00000200135)) $x)
```
Necr0x0Der, line 21: I'd not do this twice since this query is surprisingly slow

Necr0x0Der (approved):

PR #694, opened 2024-05-20
current status of documentation

DaddyWesker:
Subj

`get-doc/help!` for `empty` doesn't work just like for `=`. 
For empty help! gives
```
Function empty (type (-> %Undefined%)) No documentation
[()]
```

For `=`:
```
Function = (type (-> $t#48 $t#48 Atom)) No documentation
[()]

```

Another problem is duplicating functions. There are several of them. I've tried to create some documentation which aggregate both variants but I wasn't been able to do it for every function. For example,` @param`. It uses two type notations. In the second one `@param` have two inputs but I don't get which ones. Same goes to `@return`.

Since currently documentation is duplicating for such functions even if there is only one documentation I'm currently commenting documentation for those funcs.

For some unknown to me reason, same problem doesn't occur for `@doc-formal` though it has two type notations as well.

`help!` also doesn't show anything for `(@kind function)` and `(@kind atom)`.


lib/src/metta/runner/stdlib.metta
```diff
@@ -45,8 +46,14 @@
...
-; unify matches two atoms and returns $then if they are matched
-; and $else otherwise.
+(@doc unify
+  (@desc "Matches two atoms and returns $then if they are matched and $else otherwise")
+  (@params (
+    (@param "First atom to unify with")
+    (@param "Second atom to unify with")
+    (@param "Result if two atoms unified successfully")
+    (@param "Result otherwise")))
+  (@return "Unit atom"))
```
vsbogd, line 56: 1. It returns not "unit atom" but atom of any type.
```suggestion
  (@return "Third argument when first two atoms are matched of forth one otherwise"))
```
2. Below you sometimes used "Unit atom" sometimes used "Unit value" to describe `(->)` type returned. It is better to use same description in all cases.

lib/src/metta/runner/stdlib.metta
```diff
@@ -147,6 +256,11 @@
           (if (== $first ->) True False) ))
       ($_ False) ))))
 
+(@doc help!
+  (@desc "Function returns documentation for input value. It uses println! to output documentation")
```
vsbogd, line 260:
```suggestion
  (@desc "Function prints documentation for the input atom.")
```

lib/src/metta/runner/stdlib.metta
```diff
@@ -138,6 +242,11 @@
 ; constructor for instance, thus in practice it matches because -> has
 ; %Undefined% type. We need to assign proper type to -> and other type
 ; constructors but it is not possible until we support vararg types.
+(@doc is-function-type
+  (@desc "Function checks if input type is a function type")
+  (@params (
+    (@param "Type notation")))
+  (@return "True if input type notation is a function type, else - otherwise"))
```
vsbogd, line 249:
```suggestion
  (@return "True if input type notation is a function type, False - otherwise"))
```

lib/src/metta/runner/stdlib.metta
```diff
@@ -55,46 +62,126 @@
...
 
+(@doc @params
+  (@desc "Used for function documentation purposes. Contains several @param entities with description of each @param")
+  (@params (
+    (@param "Several (@param ...) entities")))
+  (@return "DocParameters containing description of all parameters of function in form of (@params ((@param ...) (@param ...) ...))"))
+(: @params (-> Expression DocParameters))
 
+(@doc get-doc
+  (@desc "Returns documentation for the given Atom/Function. It checks if given variable is an expression and evaluates get-doc-atom on the input value or get-doc-single-atom otherwise")
```
vsbogd, line 155: One can see the second part from the code of the function. Here the goal is to explain what function should be used for.
```suggestion
  (@desc "Returns documentation for the given Atom/Function")
```

lib/src/metta/runner/stdlib.metta
```diff
@@ -55,46 +62,126 @@
...
 
+(@doc @doc-formal
+  (@desc "Used for documentation purposes. get-doc returns documentation starting with @doc-formal symbol. @doc-formal contains 6 or 4 parameters depending on the entity being described (functions being described using 6 parameters, atoms - 4 parameters)")
+  (@params (
+    (@param "DocItem. Function/Atom name for which documentation is to be displayed. Format (@item name)")
+    (@param "DocKindAtom/DocKindFunction. Contains (@kind function) or (@kind atom) depends on entity which documentation is displayed")
+    (@param "DocType. Contains type notation of function/atom")
+    (@param "DocDescription. Function/atom description")
+    (@param "DocParameters (functions only). Description of function parameters")
+    (@param "DocReturn (functions only). Description of function's return value")))
```
vsbogd, line 117:
```suggestion
    (@param "Function/Atom name for which documentation is to be displayed. Format (@item name)")
    (@param "Contains (@kind function) or (@kind atom) depends on entity which documentation is displayed")
    (@param "Contains type notation of function/atom")
    (@param "Function/atom description")
    (@param "(Functions only). Description of function parameters")
    (@param "(Functions only). Description of function's return value")))
```

lib/src/metta/runner/stdlib.metta
DaddyWesker: 1. Of course. I've copy-pasted this from previous function and forgot to edit result...
2. So I'll replace unit value with unit atom then.

lib/src/metta/runner/stdlib.metta
```diff
@@ -55,46 +62,126 @@
 (= (unify-or-empty $a $a) unified)
 (= (unify-or-empty $a $b) (empty))
 
-; empty removes current result from a non-deterministic result
+(@doc empty
+  (@desc "Removes current result from a non-deterministic result")
+  (@return "Empty [] brackets"))
```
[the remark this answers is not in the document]
DaddyWesker: Done

lib/src/metta/runner/stdlib.metta
DaddyWesker: Done

lib/src/metta/runner/stdlib.metta
DaddyWesker: Done

lib/src/metta/runner/stdlib.metta
DaddyWesker: Done

lib/src/metta/runner/stdlib.metta
DaddyWesker: Done

DaddyWesker:
There are two types of problems. First, for some reason get-doc/help for some reason returns "No documentation". Second, for &self metta hangs for some time and then segmentation fault. I've tried to understand what is a reason behind this. For the second problem, segmentation fault could be triggered using such code:
!(let $top-space (mod-space! top)
    (unify $top-space (@doc &self $desc (@params $params) $ret)
        $ret
        Err))

So segmentation fault happening on this step. Unfortunately, I don't know why.

As for the first problem, it is somewhat tricky. For example, two functions: xor and or. Both have two input parameters and output. Here is docs:
```
(@doc or
  (@desc "Logical disjunction of two arguments")
  (@params (
    (@param "First argument")
    (@param "Second argument")))
  (@return "True if any of input arguments is True, False - otherwise"))
 
(@doc xor
  (@desc "Exclusive disjunction of two arguments")
  (@params (
    (@param "First argument")
    (@param "Second argument")))
  (@return "Return values are the same as logical disjunction, but when both arguments are True xor will return False"))
```

So, those docs are kinda similar, but `!(help! xor)` gives:
```
Function xor: (-> Bool Bool Bool) Exclusive disjunction of two arguments
Parameters:
  (type Bool) First argument
  (type Bool) Second argument
Return: (type Bool) Return values are the same as logical disjunction, but when both arguments are True xor will return False
[()]
```

and `!(help! or)` gives:
```
Function or (type (-> Bool Bool Bool)) No documentation
[()]
```
I don't get it. If I'll use this code once again
```
!(let $top-space (mod-space! top)
    (unify $top-space (@doc or $desc (@params $params) $ret)
        $ret
        Err))
```
I'll get `Err`. So for some reason unofy doesn't work for `or`. 

Moreover, if I'll add doc for `or` inside script I'm launching, everything will be fine. Unfortunately, I don't know inner structure of metta so I can't understand what is happening here.

lib/src/metta/runner/stdlib.metta
```diff
@@ -193,3 +325,346 @@
...
+  (@return "Unit atom"))
+
+(@doc sealed
+  (@desc "Replaces all occurrences of any var from var list (first argument) inside atom (second argument) by unique variable. Can be used to create a locally scoped variables")
+  (@params (
+    (@param "Variable list e.g. ($x $y)")
+    (@param "Atom which uses those variables")))
+  (@return "Second argument but with variables being replaced with unique variables"))
+
+; TODO: get-doc/help! not working for capture
```
vsbogd, line 472: @DaddyWesker  could you please recheck and remove TODO from docs which work after merging https://github.com/trueagi-io/hyperon-experimental/pull/708

DaddyWesker: @vsbogd 
That is exactly what I'm currently doing.

lib/src/metta/runner/stdlib.metta
```diff
@@ -1,4 +1,5 @@
-(@doc = 
+; TODO: for some reason get-doc doesn't work for '='
```
vsbogd, line 1: One parenthesis after return is not correct.

DaddyWesker: Yes, you right. I've fixed this and now `help!` for `=` works.

lib/src/metta/runner/stdlib.metta
```diff
@@ -193,3 +324,338 @@
...
+
+; TODO: get-doc/help! not working for and
+(@doc and
+  (@desc "Logical conjunction of two arguments")
+  (@params (
+    (@param "First argument")
+    (@param "Second argument")))
+  (@return "Returns True if both arguments are True, False - otherwise"))
+
+; TODO: get-doc/help! not working for or
```
vsbogd, line 636: @luketpeterson I would like to check it with you. Analysis shows that it doesn't work because when this module is loaded `or` token is replaced by Rust implementation. But when user executes `!(help or)` `or` is replaced by the Python implementation it is the reason why documentation is not returned (Python and Rust implementations do not match). 

In `MettaMod` code I see the following snippet: https://github.com/trueagi-io/hyperon-experimental/blob/5ab9f113dd65ecae4222298af986345430879c9f/lib/src/metta/runner/modules/mod.rs#L96-L100

To me it sounds like one one implementation of the `stdlib` tokens should be loaded: either Rust or Python but looks like in fact we have two of them at the same time. Is it made by intention or it is a bug?

vsbogd: If it is a bug and we can fix it then it could fix the documentation behaviour. On the other hand this operation duality can mean that we need separate documentation for Rust grounded atoms and Python grounded atoms.

vsbogd:
@DaddyWesker could you please also add docs for operations added by #698 and here https://github.com/trueagi-io/hyperon-experimental/pull/691/files#diff-0cc50ef2951a3babcc599e443e489112ae31c6f8e76da2d22b97160ad85d9a32

lib/src/metta/runner/stdlib.metta
luketpeterson: > To me it sounds like one one implementation of the stdlib tokens should be loaded

The python `stdlib` *could* be changed to not load `corelib`, but that's not what it currently does.  Right now it loads `corelib` (which is the Rust stdlib), and overwrites / supersedes certain definitions from corelib with its own definitions.

But I feel like this docs issue is only tangentially related to that question.

My first thought is: If the Python `hyperon/stdlib.py` is providing alternative versions of these operation atoms, it should provide documentation for them.

But my second thought is: Since the changes made a few months ago make the type conversion for grounded atoms much more flexible and add powerful arithmetic & logic ops to `corelib`, can we get rid of these op atoms in `hyperon/stdlib.py` altogether?  Although certainly in a different PR from this one?

lib/src/metta/runner/stdlib.metta
```diff
@@ -55,46 +62,128 @@
...
+  (@desc "Used for documentation purposes. Converts atom/function's name to DocItem")
+  (@params (
+    (@param "Atom/Function name to be documented")))
+  (@return "(@item Atom) entity"))
 (: @item (-> Atom DocItem))
+
+; TODO: get-doc not showing docs for following two functions (@kind function) and (@kind atom) for some reason.
+(@doc (@kind function)
+  (@desc "Used for documentation purposes. Shows type of entity to be documented. (@kind function) in this case")
+  (@return "(@kind function) entity"))
```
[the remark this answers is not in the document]
DaddyWesker: @vsbogd 
For some reason in minimal metta for `(@kind function )` documentation returned two times:
```
Atom (@kind function): (%Undefined% (-> Atom Atom)) Used for documentation purposes. Shows type of entity to be documented. (@kind function) in this case
Atom (@kind function): DocKindFunction Used for documentation purposes. Shows type of entity to be documented. (@kind function) in this case
[(), ()]
```
In regular metta everything is fine.

lib/src/metta/runner/stdlib.metta
```diff
@@ -55,46 +62,128 @@
...
 (: (@kind function) DocKindFunction)
+
+(@doc (@kind atom)
+  (@desc "Used for documentation purposes. Shows type of entity to be documented. (@kind atom) in this case")
+  (@return "(@kind atom) entity"))
```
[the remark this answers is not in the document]
DaddyWesker: Done.

lib/src/metta/runner/stdlib.metta
vsbogd: Yeah, to me these last two changes you are talking about are also a orthogonal to each other. 

One question is about documentation and here I am starting to think that it is more logical to have documentation for the grounded atom inside its implementation like we having grounded atom's type. Thus this documentation is always available from the atom itself. It contradicts to the way I wanted to see grounded atoms in future but on the other hand it is better to have internally aligned representation and switch to another one in one step.

Second question is should we have duplicates of the operations in Python stdlib. The idea is to keep Python stdlib working with native data in a native way. Taking into account calling Python method anyway requires cloning vales I doubt implementing operations on numbers and boolean values in Python makes much sense . Strings serialization is still not implemented thus we cannot fully remove duplicates probably. Thus may be we could get rid of duplicates in Python.

DaddyWesker:
> @DaddyWesker could you please also add docs for operations added by #698 and here https://github.com/trueagi-io/hyperon-experimental/pull/691/files#diff-0cc50ef2951a3babcc599e443e489112ae31c6f8e76da2d22b97160ad85d9a32

Probably done. Please check.

vsbogd:
@DaddyWesker please merge my review comments in https://github.com/DaddyWesker/hyperon-experimental/pull/3

vsbogd (approved):

PR #695, opened 2024-05-20
Cleanup variable handling logic in minimal MeTTa interpreter

vsbogd:
Extended list of variables only in operations which make stack deeper: `chain`, `function` and `collapse`. Cleaned up variables in a single place. Added an operation which applies and cleans up `Bindings` at once. Constructed finished frames using the same `Stack::finished` method.

This PR improves minimal MeTTa test performance on 20% and fixes issues in logic.


luketpeterson (approved):
Nice work!  This speeds up the minimal interpreter by ~33% in my recursive tests, across the whole range of workload sizes

PR #698, opened 2024-05-24
New standard lib operations: union, intersection, subtraction, and unique 

Adam-Vandervorst:
I had to implement a MultiTrie key conversion for grounded types for this to be of any use.
I chose the default hasher, this is not the best choice, but it's asymptotically correct for HE's internal representation.

```
> !(unique (superpose (a a b a)))
[a, b]

> !(subtraction (superpose (a b b c)) (superpose (b c c d)))
[a, b]

> !(intersection (superpose (a b c c)) (superpose (b c c c d)))
[b, c, c]

> !(union (superpose (a b b c)) (superpose (b c c d)))
[a, b, b, c, b, c, c, d]
```

### DIFF
--- a/.github/workflows/ci-auto.yml
+++ b/.github/workflows/ci-auto.yml
@@ -14,4 +14,4 @@ jobs:
     with:
       python-version: "3.7"
       os: "ubuntu-20.04"
-      cmake-version: "3.16.x"
+      cmake-version: "3.19.x"
--- a/.github/workflows/ci-manual.yml
+++ b/.github/workflows/ci-manual.yml
@@ -11,7 +11,7 @@ on:
       os:
         description: 'JSON array of OS to be checked'
         required: true
-        default: "[\"ubuntu-20.04\", \"ubuntu-22.04\", \"macos-11\", \"macos-12\", \"macos-13\"]"
+        default: "[\"ubuntu-20.04\", \"ubuntu-22.04\", \"ubuntu-24.04\", \"macos-11\", \"macos-12\", \"macos-13\", \"macos-14\"]"
         type: string
       python-version:
         description: 'JSON array of Python versions to be checked'
@@ -21,7 +21,7 @@ on:
       cmake-version:
         description: 'JSON array of CMake versions to be checked'
         required: true
-        default: "[\"3.15.x\"]"
+        default: "[\"3.19.x\"]"
         type: string
 
 jobs:
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -10,18 +10,18 @@ on:
   workflow_call:
     inputs:
       python-version:
-        description: 'A version of a python interpreter to use'
-        default: '3.7'
+        description: "A version of a python interpreter to use"
+        default: "3.7"
         required: false
         type: string
       os:
-        description: 'A container which is used to make a build'
-        default: ubuntu-20.04
+        description: "A container which is used to make a build"
+        default: "ubuntu-20.04"
         required: false
         type: string
       cmake-version:
-        description: 'CMake version to use'
-        default: 3.16.x
+        description: "CMake version to use"
+        default: "3.19.x"
         required: false
         type: string
 
@@ -62,14 +62,14 @@ jobs:
           python-version: ${{ inputs.python-version }}
 
       - name: Install CMake
-        uses: jwlawson/actions-setup-cmake@v1.14.1
+        uses: jwlawson/actions-setup-cmake@v2
         with:
           cmake-version: ${{ inputs.cmake-version }}
 
       - name: Install Conan
         uses: turtlebrowser/get-conan@v1.2
         with:
-          version: 1.60.2
+          version: "1.64"
 
       - name: Setup Conan profile
         run: |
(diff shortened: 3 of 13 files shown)


lib/src/space/grounding.rs
```diff
@@ -7,10 +7,13 @@ use crate::atom::*;
 use crate::atom::matcher::{MatchResultIter, match_atoms};
 use crate::atom::subexpr::split_expr;
 use crate::common::multitrie::{MultiTrie, TrieKey, TrieToken};
+use crate::serial::{Error, Serializer};
```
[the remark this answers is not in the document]
luketpeterson: nm.  I did it in a4d5f7f.

lib/src/space/grounding.rs
```diff
@@ -47,7 +50,25 @@ impl<'a> Iterator for GroundingSpaceIter<'a> {
...
+    fn serialize_f64(&mut self, _v: f64) -> serial::Result { Ok(self.write_u64(_v as u64)) }
+}
+
+impl Serializer for String { // for debugging
+    fn serialize_bool(&mut self, _v: bool) -> serial::Result { Ok(self.push_str(&*_v.to_string())) }
+    fn serialize_i64(&mut self, _v: i64) -> serial::Result { Ok(self.push_str(&*_v.to_string())) }
+    fn serialize_f64(&mut self, _v: f64) -> serial::Result { Ok(self.push_str(&*_v.to_string())) }
+}
+
+impl Serializer for Vec<u8> { // for speed, but is technically unsafe because not a valid utf-8 string
```
luketpeterson, line 65: There is nothing unsafe about pushing data onto a vec of bytes.  The unsafe thing would be if you then interpreted the bytes as a str / String later on, but I don't see you doing that.

So I think this comment is unnecessarily alarming.

lib/src/space/grounding.rs
```diff
@@ -56,6 +77,14 @@ fn atom_to_trie_key(atom: &Atom) -> TrieKey<SymbolAtom> {
                 expr.children().iter().for_each(|child| fill_key(child, tokens));
                 tokens.push(TrieToken::RightPar);
             },
+            // FIXME, see below
```
luketpeterson, line 80: @vsbogd to comment.  But I think this change will mean that custom grounded matchers will not run if the type successfully serializes and the hashes aren't equal.  Which I think is against the design of HE as it currently stands.

It probably a non-issue right now, in practice, since the only types that currently support serialization are numbers and bools, and there is only one implementation of those.  But I think this cuts against the intentions of the design.

lib/src/space/grounding.rs
```diff
@@ -47,7 +50,25 @@ impl<'a> Iterator for GroundingSpaceIter<'a> {
     }
 }
 
-fn atom_to_trie_key(atom: &Atom) -> TrieKey<SymbolAtom> {
+impl Serializer for DefaultHasher { // there are much speedier hashers, but I didn't want to introduce a dependency
```
luketpeterson, line 53: I think these impls probably belong in `serial.rs` instead of `grounding.rs` since they're not specific to a GroundingSpace.

Also minor style point, but I usually reserve var names beginning with '_' for unused vars.  Unclear what it signifies here.

lib/src/metta/runner/stdlib.rs
```diff
@@ -1012,12 +1026,215 @@ impl Grounded for MatchOp {
...
+impl Grounded for UniqueOp {
+    fn type_(&self) -> Atom {
+        Atom::expr([ARROW_SYMBOL, ATOM_TYPE_ATOM, ATOM_TYPE_ATOM])
+    }
+
+    fn execute(&self, args: &[Atom]) -> Result<Vec<Atom>, ExecError> {
+        let arg_error = || ExecError::from("unique expects single executable atom as an argument");
+        let atom = args.get(0).ok_or_else(arg_error)?;
+
+        // TODO: Calling interpreter inside the operation is not too good
```
luketpeterson, line 1055: Indeed.  There are two possible solutions here, but both are beyond the scope of this PR.

1.) Make the interpreter aware of which arguments need to be interpreted / reduced, so the reduction could happen before this `execute` method is invoked. or
2.) Make this method able to essentially act lazily.  Same pattern as async Rust, which returns a future object, and is evaluated by a runtime.

My vote would be for 1, because 2 comes with a lot of overhead.

vsbogd: Actually there is a third way which is possible with Minimal MeTTa interpreter:

3. one can return a minimal MeTTa `chain` which reduces the argument first and then call second part of the operation which manipulates the evaluated value (probably it is what Adam's comment says). It is the reason why `interpret_no_error` function is inside `non_minimal_only_stdlib` module.

But for this case specifically it is enough to change the `unique` op type to the `(-> %Undefined% Atom)` and argument is evaluated before call is made.

vsbogd: For example `UniqueOp` could have the following implementation:
```rs
impl Grounded for UniqueOp {
    fn type_(&self) -> Atom {
        Atom::expr([ARROW_SYMBOL, ATOM_TYPE_UNDEFINED, ATOM_TYPE_ATOM])
    }

        let arg_error = || ExecError::from("unique expects single executable atom as an argument");
        let atom = args.get(0).ok_or_else(arg_error)?;

        let mut expr: ExpressionAtom = atom.clone().try_into()?;
        let mut set = GroundingSpace::new();
        expr.children_mut().retain(|x| {
            let not_contained = set.query(x).is_empty();
            if not_contained { set.add(x.clone()) };
            not_contained
        });
        Ok(expr.into_children())
    }

    fn match_(&self, other: &Atom) -> MatchResultIter {
        match_by_equality(self, other)
    }
}
```

Adam-Vandervorst:
@vsbogd thanks for your changes! 

> For example UniqueOp could have the following implementation:

I believe we should have a similar implementation between unique, collapse, union, etc., so I propose we merge this with the `interpret_no_error` and redo the "chain evaluation" for these ops in another PR.

lib/src/space/grounding.rs
vsbogd: Change is quite big https://github.com/Adam-Vandervorst/hyperon-experimental/compare/main...vsbogd:hyperon-experimental:grounded-matching?expand=1 so I think I will merge it and will make additional changes as separate PRs.

PR #705, opened 2024-06-03
most basic snet wrapper

Necr0x0Der:


Necr0x0Der:
`:<proto`, `:<metta` and `inherit` are indeed subtyping relations, but in different type systems (proto, metta, ai-dsl/ontology).

Issue #241, opened 2023-02-07
already mutably borrowed: BorrowError in logging


noskill:
```
from hyperon import *

def test():
        program = '''
            (X Y)
            (Y Z)
            (Y W)
            (Z W)
            (!println (match &self (, ($x $y) ($y $z) ($y $w) ($z $w)) (add-atom &self ($y $x))))
        '''
        runner = MeTTa()
        result = runner.run(program)

        print(runner.space().get_atoms())

test()
```

stacktrace:

[log.txt](https://github.com/trueagi-io/hyperon-experimental/files/10673786/log.txt)

luketpeterson:
The error in the stack trace is caused by a RefCell being duplicately borrowed - usually happens when an outer calling function holds a Ref and then an inner called function tries to re-borrow the RefCell.

This may already be resolved, or there may be more involved in reproducing it.  Running the test on current main produces expected output:
```
[GroundingSpace(addr=0x6000035cc0c8), (X Y), (Y Z), (Y W), (Z W), (!println (match GroundingSpace(addr=0x6000035cc018) (, ($x $y) ($y $z) ($y $w) ($z $w)) (add-atom GroundingSpace(addr=0x6000035cc018) ($y $x))))]
```

luketpeterson:
Following up, I enabled logging and I notices a difference in the log from the crash case vs. my log:

From the panic trace:
```
[2023-02-07T09:59:26Z DEBUG hyperon::metta::runner] Metta::load_module_space: load module space stdlib
[2023-02-07T09:59:26Z DEBUG hyperon::space::grounding] IndexTree::add(): key: (: if (-> Bool Atom Atom $t)), value: 0
[2023-02-07T09:59:26Z DEBUG hyperon::space::grounding] IndexTree::add(): key: (= (if True $then $else) $then), value: 1
[2023-02-07T09:59:26Z DEBUG hyperon::space::grounding] IndexTree::add(): key: (= (if False $then $else) $else), value: 2
[2023-02-07T09:59:26Z DEBUG hyperon::space::grounding] IndexTree::add(): key: GroundingSpace(addr=0x5629c84e3d58), value: 0
[2023-02-07T09:59:26Z DEBUG hyperon::space::grounding] IndexTree::add(): key: (X Y), value: 1
[2023-02-07T09:59:26Z DEBUG hyperon::space::grounding] IndexTree::add(): key: (Y Z), value: 2
[...]

PR #706, opened 2024-06-05
Separate custom matched grounded atoms and grounded atoms matched by equality

vsbogd:
Separate custom matched grounded atoms and grounded atoms matched by equality to address https://github.com/trueagi-io/hyperon-experimental/pull/698/files#r1623809472

We cannot understand whether custom matching is implemented on a grounded atom because of lack of the specialization feature in stable Rust, so I compared two approaches:
- returning optional matching procedure `fn Grounded::get_match(&self) -> Option<Box<dyn Fn(&Atom) -> MatchResultsIter>>`
- and adding separate `fn Grounded::as_match(&self) -> Option<&dyn CustomMatch>`
- having separate flag `Grounded::has_custom_matching`

I chosen last one because:
- it seems more clean and logical to me having each separate customization aspect as a separate trait
- if specialization is available in a stable Rust we can automatize it
- returning closures from `get_match` requires additional boxing and looks more surprising than implementing a trait

I like this approach finally and I think we could reuse it to implement `Hash` on value-like grounded atoms and to move `execute` into a separate trait.

Despite change is big most interesting files are:
- `lib/src/atom/mod.rs`
- `lib/src/common/shared.rs`
- `c/src/atom.rs`

Other files contains removed default `match_by_equality` logic and rearranging custom matching logic to match new API.

@Adam-Vandervorst FYI (cannot add you as a reviewer).

### DIFF
--- a/lib/src/atom/mod.rs
+++ b/lib/src/atom/mod.rs
@@ -303,10 +303,10 @@ pub fn make_variables_unique(mut atom: Atom) -> Atom {
 // values:
 // - type_() to return MeTTa type of the atom;
 // - execute() to represent functions as atoms;
-// - match_() to implement custom matching behaviour.
+// - as_match() to return optional custom matching behaviour API.
 
-// match_by_equality() method allows reusing default match_() implementation in
-// 3rd party code when it is not required to be customized.
+// By default Grounded::as_match() returns None which makes GroundedAtom's
+// implementation match atom by equality.
 
 /// Grounded function execution error.
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -351,7 +351,21 @@ pub trait GroundedAtom : Any + Debug + Display {
         self.as_grounded().execute(args)
     }
     fn match_(&self, other: &Atom) -> matcher::MatchResultIter {
-        self.as_grounded().match_(other)
+        match self.as_grounded().as_match() {
+            Some(match_) => match_.match_(other),
+            None => {
+                let equal = match other {
+                    Atom::Grounded(other) => self.eq_gnd(other.as_ref()),
+                    _ => false,
+                };
+                let result = if equal {
+                    matcher::BindingsSet::single()
+                } else {
+                    matcher::BindingsSet::empty()
+                };
+                Box::new(result.into_iter())
+            },
+        }
     }
     // A mutable reference is used here instead of a type parameter because
     // the type parameter makes impossible using GroundedAtom reference in the
@@ -373,17 +387,16 @@ impl dyn GroundedAtom {
     }
 }
 
-/// Trait allows implementing grounded atom with custom behaviour.
-/// [rust_type_atom], [match_by_equality] and [execute_not_executable]
-/// functions can be used to implement default behavior if requried.
-/// If no custom behavior is needed then simpler way is using [Atom::value]
-/// function for automatic grounding.
+/// Trait allows implementing grounded atom with custom behavior.
+/// [rust_type_atom] and [execute_not_executable] functions can be used to
+/// implement default behavior if required. If no custom behavior is needed
+/// then simpler way is using [Atom::value] function for automatic grounding.
 ///
 /// # Examples
 ///
 /// ```
 /// use hyperon::*;
-/// use hyperon::matcher::{Bindings, MatchResultIter, match_atoms};
+/// use hyperon::matcher::{Bindings, match_atoms};
 /// use std::fmt::{Display, Formatter};
 /// use std::iter::once;
 ///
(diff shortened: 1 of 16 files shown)


lib/src/atom/mod.rs
```diff
@@ -464,7 +531,7 @@ pub fn match_by_equality<T: 'static + PartialEq + Debug>(this: &T, other: &Atom)
     }
 }
 
-/// Returns either single emtpy [matcher::Bindings] instance if the string representing `self` and
+/// Returns either single empty [matcher::Bindings] instance if the string representing `self` and
```
luketpeterson, line 534: Good catch.

lib/src/common/shared.rs
```diff
@@ -164,15 +163,26 @@ impl<T: Grounded> Grounded for Shared<T> {
...
+            Some(_) => Some(self),
+        }
     }
 
     fn execute(&self, args: &[Atom]) -> Result<Vec<Atom>, ExecError> {
         self.0.borrow().execute(args)
     }
 }
 
+impl<T: Grounded> CustomMatch for Shared<T> {
```
luketpeterson, line 178: I don't love blanket impls for something like this.  But my issue is with the Shared type more generally, so I think the code in this PR is fine in the context of the rest of the design.

luketpeterson (approved):
I am really happy with this change.  It simplifies the grounded atom API for the common case and enables a fast path.

lib/src/space/grounding.rs
```diff
@@ -58,24 +58,16 @@ pub(crate) fn atom_to_trie_key(atom: &Atom) -> TrieKey<SymbolAtom> {
...
-            // FIXME, see below
-            Atom::Grounded(g) => {
+            Atom::Grounded(g) if g.as_grounded().as_match().is_none() => {
+                // TODO: Adding Hash on grounded atoms matched by equality is
+                // required in order to make TrieToken::Exact be generated for
+                // them.
                 let mut h = DefaultHasher::new();
                 match (*g).serialize(&mut h) {
                     Ok(()) => { tokens.push(TrieToken::Exact(SymbolAtom::new(ImmutableString::Allocated(h.finish().to_string())))) }
                     Err(_) => { tokens.push(TrieToken::Wildcard) }
```
Adam-Vandervorst, line 68: Should all (grounded) atoms that do not provide a custom match, provide a serializer?
I.e. as_match would conceptually be `Result<CustomMatch, Serializer>`.
This way HE can get rid of the linear search disambiguation logic, because a grounded item either fits into the default matcher, or it provides its own matching logic.

vsbogd: Not 100% sure I got you right. If custom matching logic is implemented for the grounded atom then in general we cannot use serialization to get `TrieToken::Exact` anyway, so probably you didn't mean this. Regarding `Serializer` for the default matching logic I am going to add this in next PR. In this PR my goal was to patch as quick as possible the behavior for the case when custom matching logic and `serialize` are implemented by the grounded atom at the same time.

lib/src/common/shared.rs
vsbogd: I am not sure what would be a better solution here. Do you mean we should require from user manually implementing `Grounded` and `CustomMatch` for `Shared<SomeType>` if user wants using `Shared` reference as a grounded atom?

lib/src/space/grounding.rs
Adam-Vandervorst: This PR LGTM, my analysis was that there are four cases:
CustomMatch, Serialize => should match using custom logic
Serialize => should match using serialized version of the atom wrapped in exact
CustomMatch => should match using custom logic
=> should never happen, let's make this illegal and unrepresentable

vsbogd: Ah, ok, so you mean `Trie` implementation could call custom matching logic when matching key. Yes it is possible to implement it. The reason it is not implemented yet is that I wanted to have the single instance of the matching procedure instead of two instances (searching in space and matching with a single atom). 

It is indeed possible to implement matching a single atom via matching with space which contains a single atom and thus still have single instance of the matching procedure. Or for instance parameterize matching procedure by a function which provides next atom to match, then implement it differently for a single atom case and for `Trie` search case.

I think it makes sense and I think I will try to implement this.

lib/src/common/shared.rs
luketpeterson: My issue is with Shared in general (not with this PR). But the internal mutability pattern (RefCell) comes with big implications for how you should access the underlying objects.  Therefore the pattern should be incorporated into the object's API itself instead of creating a wrapper that can wrap another object.

Otherwise it leads to bugs like https://github.com/trueagi-io/hyperon-experimental/issues/241 (as just one example)

So if a grounded atom needs to be a reference to a larger structure with shared mutability, (like with a Space) creating a reference object (e.g. DynSpace) that manages access is a better solution.

PR #707, opened 2024-06-05
Add notes on switch and case operations difference

vsbogd:


DaddyWesker:
Thanks for this explanation.

PR #708, opened 2024-06-07
Fix equality check for grounded ops in stdlib

vsbogd:
Fixes most of the issues with #694
Add macros to add default implementations of the `PartialEq` and `Display` for the grounded operations. Use macros to redefined `PartialEq` logic for grounded operations in standard library.


lib/src/metta/runner/stdlib.rs
```diff
@@ -24,6 +24,22 @@ use regex::Regex;
...
 
+macro_rules! grounded_op {
+    ($name:ident, $disp:literal) => {
+        impl PartialEq for $name {
+            fn eq(&self, _other: &Self) -> bool {
+                true
+            }
+        }
+
+        impl Display for $name {
```
luketpeterson, line 35:
```suggestion
        impl std::fmt::Display for $name {
```

luketpeterson: Display isn't in the preamble like PartialEq, so if we want to export this macro it will be good to specify it by full path.

luketpeterson (dismissed):
I <3 reducing boilerplate.

PR #716, opened 2024-06-14
Rename wiped variables in Bindings::retain

vsbogd:
See unit test as an example.

### DIFF
--- a/lib/src/atom/matcher.rs
+++ b/lib/src/atom/matcher.rs
@@ -736,8 +736,17 @@ impl Bindings {
             Atom::Variable(var) => {
                 if to_remove.contains(&var) {
                     match self.resolve(var) {
-                        Some(Atom::Variable(v)) if to_remove.contains(&v) => { self.rename_var(var, &to_remove).map(|v| *atom = v); },
-                        Some(value) => { *atom = value; },
+                        Some(mut value) => {
+                            value.iter_mut().for_each(|atom| match atom {
+                                Atom::Variable(var) if to_remove.contains(&var) => {
+                                    self.rename_var(var, &to_remove).map(|var| {
+                                        *atom = var;
+                                    });
+                                },
+                                _ => {},
+                            });
+                            *atom = value;
+                        },
                         None => {},
                     }
                 }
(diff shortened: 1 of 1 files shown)


lib/src/atom/matcher.rs
```diff
@@ -1801,7 +1810,7 @@ mod test {
             .add_var_equality(&VariableAtom::new("a"), &VariableAtom::new("b"))?;
         bindings.apply_and_retain(&mut atom, |_| false);
         assert!(bindings.is_empty());
-        assert_eq!(atom, expr!(a b));
+        assert_eq!(atom, expr!(a a));
```
luketpeterson, line 1813: I always thought of variable equality as commutative, but it appears I was incorrect.

Would it ever be correct for this apply operation to return `expr!(b b)`?

luketpeterson (approved):

lib/src/atom/matcher.rs
vsbogd: > Would it ever be correct for this apply operation to return expr!(b b)?

Yes, it can be one of `expr!(a a)` or `expr!(b b)` or even `expr!(n n)`. Current implementation always returns `expr!(a a)` but it is not generally required. Are you suggesting making check more universal?

luketpeterson: To be honest I haven't gotten to the level of having a suggestion yet.  I guess the test is fine given that all those are equivalent.  i.e. a variable's name only matters in the context of the other variables in an expression or bindings set.

Technically this test is overly-sensitive (will fail on a valid case) but I don't think it matters since it conforms to the implementation.

My comment was just to ask about what constitutes correct MeTTa behavior regarding variable names.

vsbogd: To add some details instance of `Bindings` plus instance of `Atom` plus scope data contains the full information about variables and their values. In this particular example we have the following pieces of the information:
- atom has two variables inside (`atom`)
- variables are equal to each other (`bindings`)
- variables don't have values (`bindings`)
- variables are already out of scope (condition argument of the `apply_and_retain`)

This is the information we would like to keep. Cleaning up the `Bindings` we can keep one of the names or invent the new one. It doesn't matter because these variables are out of scope and will not be filled by values anyway. But users like to have familiar variable names. It is the reason why `Binding` has a `var` field - the preferred var name. And it is returned when absent value is resolved.

PR #718, opened 2024-06-14
Move out executable

vsbogd:
Straightforward change to separate `Grounded::execute` method in a same way `Grounded::match_` is separated. 
```
trait Grounded {
  fn execute(...) -> ...;
}
```
by
```
trait Grounded {
  as_execute(&self) -> Option<&dyn CustomExecute>
}
trait CustomExecute {
  fn execute(...) -> ...;
}
```
Using such interface one can check whether atom is executable in advance without executing it.
Main change is inside `lib/src/atom/mod.rs` all other changes are moving old code according new interface.

### DIFF
--- a/lib/src/atom/mod.rs
+++ b/lib/src/atom/mod.rs
@@ -348,7 +348,10 @@ pub trait GroundedAtom : Any + Debug + Display {
         self.as_grounded().type_()
     }
     fn execute(&self, args: &[Atom]) -> Result<Vec<Atom>, ExecError> {
-        self.as_grounded().execute(args)
+        match self.as_grounded().as_execute() {
+            Some(execute) => execute.execute(args),
+            None => Err(ExecError::NoReduce),
+        }
     }
     fn match_(&self, other: &Atom) -> matcher::MatchResultIter {
         match self.as_grounded().as_match() {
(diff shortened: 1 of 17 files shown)


lib/src/atom/mod.rs
```diff
@@ -348,7 +348,10 @@ pub trait GroundedAtom : Any + Debug + Display {
         self.as_grounded().type_()
     }
     fn execute(&self, args: &[Atom]) -> Result<Vec<Atom>, ExecError> {
-        self.as_grounded().execute(args)
+        match self.as_grounded().as_execute() {
```
luketpeterson, line 351: This is just a question, not a suggestion.  But why not eliminate this method entirely (as well as `execute_non_executable()`) and work with `CustomExecute` objects within the interpreter?

Are you imagining some implementation of `GroundedAtom` that doesn't use a `CustomExecute` but still provides execution behavior?

luketpeterson (approved):

lib/src/atom/mod.rs
vsbogd: > Are you imagining some implementation of GroundedAtom that doesn't use a CustomExecute but still provides execution behavior?

No,  the only goal of the `GroundedAtom` is to hide the real type of the grounded data inside `Atom` enum. I tried using `Grounded` for this at first. May be I just didn't know Rust good enough, but it was difficult to keep `Grounded` simple enough for implementation in this case. This is why I added `GroundedAtom` which is automatically implemented and only hides the actual type. Now I think it can be done without additional trait, probably by implementing `dyn Grounded`.

Returning to your original question I think removing `GroundedAtom::execute()` makes perfect sense. I didn't do it because I used to work via `GroundedAtom` but in fact using `Grounded::as_execute()` makes API more clear. So I implemented it in https://github.com/trueagi-io/hyperon-experimental/pull/718/commits/57ecd884d6df623cbe006765f22d591e76ceab7c

vsbogd: I also think it makes sense to eliminate `GroundedAtom::match_` as well, but I would do it in a separate PR.

luketpeterson (approved):

vsbogd:
Luke, thanks for reviewing this!

PR #729, opened 2024-07-05
[BUGFIX] Fixing DAS vars resolution

angeloprobst:
### Rationale
While working with MeTTa and DAS Gate, I faced the following situation: the same query that works perfectly with the built-in atom space, when executed using DAS, does not resolve some variables if a function is defined and called to match some condition.
For example:
```
[(parent of Paul = $y#40)]
[(parent of Paul = John)]
```
The first line is the output of a query on DAS. And the second one comes from the same query but using the built-in space.

The issue does not happen on DAS if the match is made without calling a function.
For example, this query fails (output shows a non-resolved variable):
```lisp
(= (get_parent_from_das $x) (match &das (Parent $y $x) $y))
!(parent of Paul = (get_parent_from_das Paul))
```
However, this `match` works perfectly (it shows the value):
```lisp
!(match &das (Parent $y Paul) (parent of Paul = $y))
```
And both queries are equivalent.

### Sample Script
(two first queries do not have variables properly resolved)
```lisp
!(import! &self das_gate)

!(bind! &das (new-das))
!(bind! &local (new-space))

!(add-atom &das (Parent John Paul))  ; Paul is child of John
!(add-atom &das (Person Paul))

!(add-atom &local (Parent John Paul))  ; Paul is child of John
!(add-atom &local (Person Paul))

; getting parent from das using a function
(= (get_parent_from_das $x)
    (match &das (Parent $y $x) $y))
!(parent of Paul = (get_parent_from_das Paul))  ; person as a constant
!(match &das (Person $x)  ; person as a wildcard
    (parent of $x = (get_parent_from_das $x)))

; getting parent from das, inline, no function
!(match &das (Parent $y Paul) (parent of Paul = $y))  ; person as a constant
!(match &das (Person $x)  ; person as a wildcard
    (parent of $x = (match &das (Parent $y $x) $y)))

; getting parent from local using a function
(= (get_parent_from_local $x)
    (match &local (Parent $y $x) $y))
!(parent of Paul = (get_parent_from_local Paul))  ; person as a constant
!(match &local (Person $x)  ; person as a wildcard
    (parent of $x = (get_parent_from_local $x)))

; getting parent from local, inline, no function
!(match &local (Parent $y Paul) (parent of Paul = $y))  ; person as a constant
!(match &local (Person $x)  ; person as a wildcard
    (parent of $x = (match &local (Parent $y $x) $y)))
```
Output is:
(two first lines should show `John` instead of `$y#40` and `$y#55`)
```
[(Parent of Paul = $y#40)]
[(Parent of Paul = $y#55)]
[(Parent of Paul = John)]
[(Parent of Paul = John)]
[(Parent of Paul = John)]
[(Parent of Paul = John)]
[(Parent of Paul = John)]
[(Parent of Paul = John)]
```

### DIFF
--- a/c/src/atom.rs
+++ b/c/src/atom.rs
@@ -236,15 +236,27 @@ pub unsafe extern "C" fn atom_expr_from_vec(children: atom_vec_t) -> atom_t {
     Atom::expr(children).into()
 }
 
-/// @brief Create a new Variable atom with the specified identifier
+/// @brief Create a new Variable atom with the specified name
 /// @ingroup atom_group
-/// @param[in]  name  The identifier for the newly created Variable atom
+/// @param[in]  name  The name for the newly created Variable atom
 /// @return An `atom_t` for the Variable atom
 /// @note The caller must take ownership responsibility for the returned `atom_t`
 ///
 #[no_mangle]
 pub unsafe extern "C" fn atom_var(name: *const c_char) -> atom_t {
     Atom::var(cstr_as_str(name)).into()
+}
+
+/// @brief Create a new Variable atom with the specified name and id
+/// @ingroup atom_group
+/// @param[in]  name  The name for the newly created Variable atom
+/// @param[in]  id    The unique id for the newly created Variable atom
+/// @return An `atom_t` for the Variable atom
+/// @note The caller must take ownership responsibility for the returned `atom_t`
+///
+#[no_mangle]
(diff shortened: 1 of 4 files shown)


vsbogd (approved):
Hi @angeloprobst, thank you for the PR. It is indeed a step in a right direction. I was thinking about similar change but in a Rust core library. Also I need to finish C API rework and remove using strings as the keys in variable bindings.

I am merging it and going to raise PR with the follow up changes soon.

PR #730, opened 2024-07-05
Add Metta.evaluate_atom() method to the Python API

vsbogd:
It is required to migrate metta-motto code to the latest MeTTa interpreter (https://github.com/trueagi-io/hyperon-experimental/pull/728)


Necr0x0Der (approved):
Thanks. Seems to be working

PR #731, opened 2024-07-08
Improve variable related API

vsbogd:
Follow up #729. This PR contains two main parts:
1. Remove `Bindings` and `BindingsSet` method variants which work with strings instead of variables. This is technical debt after introducing full functional Python API for the variable bindings.
2. Add `VariableAtom::parse_name()` method to parse `VariableAtom::name()` method results into Rust and Python API. This is a technical debt after introducing unique variable instances.

New methods are used to fix DAS gate variable conversion logic.

@besSveta please check this change doesn't break `sql_space.py` logic.
@angeloprobst FYI.


CICS-Oleg (approved):
It's all clear on my side.

PR #732, opened 2024-07-09
Minimize docker image size

vsbogd:
Use minimal Python 3.10 image based on Debian `python:3.10-slim-bookworm` as a starting point.
Use multi-stage build to start from clean environment after building is finished. In addition debug information is stripped from Rust binaries. Resulting image's size is about 180 Mb which is 50Mb greater than the Python runtime size.

Rust executable file is renamed to `metta-rust` and Python executable is aliased by `metta-py` in order to remove name conflict between them. Documentation is updated.

`.dockerignore` file is added in order to make Docker builds faster. `.dockerfile` is a soft-link to the `/.gitignore` which now keeps all files ignored by Git in repo. Welcome message is added to the container to show environment version and explain basic commands.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+.gitignore
--- a/.gitignore
+++ b/.gitignore
@@ -9,8 +9,8 @@ Cargo.lock
 .DS_Store
 .vscode
 .aider*
+__pycache__
 /build
-/python/__pycache__
 /python/hyperon.egg-info
 /python/dist
 /python/build
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
-/build
+# `.gitignore` file is also used to exclued files going into Docker image
+# building context. All ignored files should be put into the same `.gitignore`
+# at the root of the repository by this reason.
 *.swp
 compile_commands.json
 .cache
@@ -7,3 +9,12 @@ Cargo.lock
 .DS_Store
 .vscode
 .aider*
+/build
+/python/__pycache__
+/python/hyperon.egg-info
+/python/dist
+/python/build
+/python/*.so
+/python/*.dylib
+/python/wheelhouse
+/python/hyperon/_version.py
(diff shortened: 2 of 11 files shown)


luketpeterson:
Sorry to bring this up after you already implemented the name change, I want to discuss the name `metta-rust`, as opposed to something else.  To me, it seems like the fact that it's implemented with Rust is a detail, and something like `metta-repl` or `metta-cli` might be more appropriate.

But it's not a big deal to me, if you feel differently.

vsbogd:
> I want to discuss the name metta-rust, as opposed to something else.

Thanks @luketpeterson I added you and @Necr0x0Der mainly to discuss this matter. 

The problem I am trying to solve we now have two execution points but users only know about one. For the end user the difference between them: one is started from Python and fully Python compatible, other is started from Rust and its Python compatibility is under question. Both could be REPL potentially but Python implementation doesn't have this feature yet.

Choosing this naming scheme I want to clearly separate them in user's mind according the criteria I see. There is a question which one should be aliased by `metta` keyword. In fact I would prefer Rust one as a default (I even raised #605) but as I wrote its compatibility with Python is under the question. In fact it doesn't even start on my environment without Python being disabled.

Saying this I don't insist on this particular naming scheme and I ready to rename one more time `metta-rust` to `metta-repl`. I just don't think it reflects the specifics good enough. I would suggest renaming Python interpreter into `metta-py` anyway to reflect its specifics.

vsbogd:
Renaming question intersects with a conceptual question: do we want separate executable for each runtime we support? On the one hand we would prefer single executable supporting all needed runtimes. On the other hand while we probably can doing this we didn't do this yet even for Python. Thus while we prefer having single `metta` executable implemented in Rust de-facto we have two different implementations for Rust and Python. And in fact I don't think that `metta-repl` is more long-term than `metta-rust` but as I said I am ready to use `metta-repl` as well.

vsbogd:
@luketpeterson , renamed to `metta-repl` in https://github.com/trueagi-io/hyperon-experimental/pull/732/commits/6b62400e7bf6c509b96c310d35870c3fb0843a2c

luketpeterson (approved):
insofar as I can see, these changes look good.  I am not a Docker expert however.

PR #734, opened 2024-07-09
Propagating python errors in loader through to MeTTa

luketpeterson:
The `test_error_pyext` test in the `test_extend.py` was aborting due to an (intentional) uncaught exception from Python.

This PR fixes it by implementing the missing error pathway from C loader function, and then using that pathway in hyperonpy


vsbogd (approved):
Thanks!

PR #738, opened 2024-07-10
Add explanation of the scope of the variables for minimal MeTTa

vsbogd:


docs/minimal-metta.md
```diff
@@ -171,6 +171,35 @@ separate alternative.
 can collect the list of alternatives using `collapse-bind` filter them and
 return filtered items to the plan using `superpose-bind`.
 
+## Scope of a variable
+
+Variable name should be unique inside a separate expression evaluated. The
```
luketpeterson, line 176: Apologies in advance if I am overstepping.   I made a few edits for English flow, but if I changed your voice or made it less clear then feel free to reject.

```suggestion
Each separately evaluated expression is a variable scope, and therefore variable names are treated as unique inside an expression.
```

docs/minimal-metta.md
```diff
@@ -171,6 +171,35 @@ separate alternative.
...
+reason is that the whole expression is a variable scope. For example one can
+write the expression `(chain (unify $parent Bob () ()) $_ $parent)`. And value
+of the `$parent` is returned correctly.
+
+When variable is passed as an argument to the function call and matched by a
```
luketpeterson, line 181:
```suggestion
When a variable is passed as an argument to a function call and matched by a
```

docs/minimal-metta.md
```diff
@@ -171,6 +171,35 @@ separate alternative.
...
+value then the value is assigned to the variable. If variable passed as an
+actual argument is matched by a formal argument variable then it is referenced
```
luketpeterson, line 183: I feel like the terms "formal argument" and "actual argument" need to be defined.  The example helps but I found this part a little bit confusing.

docs/minimal-metta.md
```diff
@@ -171,6 +171,35 @@ separate alternative.
...
+by the formal argument variable. In this case the actual argument variable can
+receive a value outsides of its scope.
+
+For example the following code (written using MeTTa runner syntax) returns `B`:
+```
+(= (foo $b) (function (chain (unify B $b () ()) $_ (return ()))))
+!(chain (eval (foo $a)) $_ $a)
+```
+
+If two separate expressions in the atomspace has the variable with the same
```
luketpeterson, line 193:
```suggestion
If two separate expressions in the space have a variable with the same
```

docs/minimal-metta.md
```diff
@@ -171,6 +171,35 @@ separate alternative.
...
+name inside then the scopes of the variables are different. Consider the
```
luketpeterson, line 194:
```suggestion
name, but the variables reside in independent scopes, then the variables are different. Consider the
```

docs/minimal-metta.md
```diff
@@ -171,6 +171,35 @@ separate alternative.
...
+following example:
+```
+(= (foo) (function (chain (unify A $a () ()) $_ (return ()))))
+!(chain (eval (foo)) $_ $a)
+```
+Here value will not be assigned to the `$a` from the caller expression because
```
luketpeterson, line 200:
```suggestion
Here the value will not be assigned to the `$a` from the caller expression because
```

docs/minimal-metta.md
```diff
@@ -171,6 +171,35 @@ separate alternative.
...
+two variables has different scopes and not referenced by each other.
```
luketpeterson, line 201:
```suggestion
each of the two variables has a different scope and they do not reference each other.
```

luketpeterson (dismissed):

docs/minimal-metta.md
vsbogd: It is a quite common terminology (for example see [Wikipedia article](https://en.wikipedia.org/wiki/Parameter_(computer_programming))). Is using word "argument" for the passed arguments and word "parameter" for the function input variables clears the thing? For example:
```
When a variable is passed as an argument to a function call and matched by a
value used in a function definition then the value is assigned to the variable.
If variable passed as an argument is matched by a function's parameter variable
then parameter variable references the variable passed as an argument. In this
case the variable passed as an argument can receive a value outsides of its
scope.
```

vsbogd:
Luke thank you very much for the edits!

docs/minimal-metta.md
luketpeterson: Ah ok.  Thanks.  I had never encountered that phrasing before.

luketpeterson (approved):

docs/minimal-metta.md
vsbogd: Well it is not only terminology and I started using it intentionally to make readers clearly see that both sides are _arguments_ but _different arguments_. I am not sure if it is easy to read this. If it is hard and there is more familiar words then I would use them.

PR #741, opened 2024-07-12
improvements to jetta wrapper and tests for lambdas

Necr0x0Der:


Necr0x0Der:
@vsbogd , FYI regarding lambdas

PR #745, opened 2024-07-23
Remove type based duplicates

### LINKED ISSUE #235: Multiple type definitions cause multiple results

vsbogd:
Fixes #235. Type checking behavior is changed (see tests `c1_`, `d2_` and `d5_`. In particular for the case like:
```
(: f (-> $t Number))
(f (+ 5 "S"))
```
previous behavior was to return the most inner type error, i.e. `(Error "S" BadType)`, now interpreter returns less inner type error, i.e. `(Error (+ 5 "S") BadType)`. I am not sure whether one should be preferred over another.


python/tests/scripts/d5_auto_types.metta
```diff
@@ -59,7 +59,11 @@ Auto type-checking can be enabled
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ; let binds without type checking, but errors when used
 (let $x (+ 5 "S") $x)
+
 (: f (-> $t Number))
-!(assertEqualToResult
-   (f (+ 5 "S"))
-   ((Error "S" BadType)))
+; TODO: This test has different behavior in old_interpreter and minimal interpreter.
```
Necr0x0Der, line 64: Yes, I like the new behavior better

python/tests/scripts/d2_higherfunc.metta
```diff
@@ -168,12 +168,14 @@
 !(assertEqualToResult
   (get-type (fmap (curry-a + 2) (Left "5")))
   ())
-!(assertEqualToResult
-  (get-type (fmap (curry-a + 2) (UntypedC "5")))
-  ())
-!(assertEqualToResult
-  (get-type (fmap (curry-a + 2) (UntypedC (Null) 5)))
-  ())
+; Two examples below are type-checked successfully because, (UntypedC "5")
```
Necr0x0Der, line 171: Yes, I agree. The type of `(UntypedC "5")` should be reducible to `%Undefined%` meaning that the whole expression can type-check. I'd add "FIXME" or "TODO", because we may need to revisit these tests and uncomment their updated versions

vsbogd, line 171:
```suggestion
; TODO: Two examples below are type-checked successfully because, (UntypedC "5")
```

python/tests/scripts/d2_higherfunc.metta
```diff
@@ -168,12 +168,14 @@
...
   (get-type (fmap (curry-a + 2) (Left "5")))
   ())
-!(assertEqualToResult
-  (get-type (fmap (curry-a + 2) (UntypedC "5")))
-  ())
-!(assertEqualToResult
-  (get-type (fmap (curry-a + 2) (UntypedC (Null) 5)))
-  ())
+; TODO: Two examples below are type-checked successfully because, (UntypedC "5")
+; can return result which has an appropriate type.
```
vsbogd, line 172:
```suggestion
; TODO: Two examples below are type-checked successfully because, (UntypedC "5")
; can return result which has an appropriate type. Uncomment when old_interpreter 
; feature is removed.
```

Necr0x0Der (approved):

PR #751, opened 2024-07-24
Remove no_python feature, make it work by default

### LINKED ISSUE #748: cargo run --features no_python --bin metta-repl before and after metta-repl
I do not know how much of a bug this is, but the user experience is suboptimal.

I just built the repo with cargo build.

1. If I launch metta-repl directly from target/debug, I get:

```console
Fatal Error: ModuleNotFoundError: No module named 'hyperon'
```

2. If I then launch `cargo run --features no_python --bin metta-repl`, I get the REPL.

3. Then, if I launch metta-repl again, directly from target/debug, it works too.

The initialisation of the hyperon module from rust seems... weird.

vsbogd:
Fixes #748

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,7 +29,7 @@ ADD --chown=user:users . ${HOME}/hyperon-experimental
 
 WORKDIR ${HOME}/hyperon-experimental
 RUN cargo test --release
-RUN cargo build --release
+RUN cargo build --features python,git --release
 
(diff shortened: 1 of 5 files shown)


luketpeterson (approved):
I am fine to switch the default...

However I still wonder if runtime detection (build with support for both) is a worthwhile feature...  Or if that just moves the user confusion from the build process to the run.

For example, if the user is expecting access to to Python atoms & spaces, but it falls back to the no_python path at runtime.

PR #764, opened 2024-08-29
metta code generation for service call

DaddyWesker:
Currently this is draft since possible there will be some sort of fixes after @Necr0x0Der review.

### DIFF
--- a/python/sandbox/snet/snet_io.py
+++ b/python/sandbox/snet/snet_io.py
@@ -51,7 +51,9 @@ class SNetSDKWrapper:
         if isinstance(atom, GroundedAtom):
             return atom.get_object().content
         return repr(atom)
-    
+
+
+
     def __call__(self, command_a, *args_a):
         command = self._unwrap_atom(command_a)
         args = []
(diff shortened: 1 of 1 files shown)


DaddyWesker:
Possible test is something like this i guess
```
!(import! &self snet_io)
!(snet-sdk init (eth_rpc_endpoint "")
                (private_key ""))
!(let $service
   (snet-sdk create_service_client "naint" "question-answering-long-seq"
        (free_call_auth_token_bin "")
        (free_call_token_expiry_block ))
   ((py-dot $service generate_metta_launch_code)))
```

Output:
```
[()]
[()]
["
(: qa (-> String String))
( = (qa $question)
	((question-answering-long-seq) "qa" "Question"
	(Kwargs (question $question))
	)
)"]
```

Necr0x0Der:
I minor thing I'd change is that `__get_inputs_outputs__` seems to have no sense to execute on each `__call__`. Can it be done on `__init__`? We create a service caller either to get meta-info about it or to make calls to it. If making calls will now require to get meta-info, then it seems reasonable to call `get_service_messages` on `__init__` and memorize its results. I'd say that we will not need to have `get_service_messages` as interface function, because we cannot do much with its output in MeTTa. So, let's move `self.service_client.get_services_and_messages_info()` to `__init__` and memorize its output and reuse it in other methods.

DaddyWesker:
@Necr0x0Der 
Done. Though currently we don't need result of get_services_and_messages_info in any other functions (only on __init__), I've still made it self.message_info just i case for future additions.

Necr0x0Der:
> currently we don't need result of get_services_and_messages_info in any other functions

My idea was to keep `get_service_messages` and return the cached result in it.

DaddyWesker:
@Necr0x0Der my bad, done.

Necr0x0Der (approved):

PR #769, opened 2024-09-11
Update and refactor of snet_io

Necr0x0Der:


Necr0x0Der:
@mvpeterson , @DaddyWesker , @besSveta , FYI

PR #771, opened 2024-09-13
current state of pretty print

DaddyWesker:
Examples of pretty_prints in current state:

```
(= (abs $x) (if
		(> $x 0)
		$x (if
			(== $x 0)
			0 (* $x -1)
			)
		))

( = (response) (if
		(== (messages) (user "match world."))
		("Hi there")
		(("Bye"))
		))

(: gen_gpt_2 (-> String String Number Number Number String))
(= (gen_gpt_2 $start_text $run_name $temperature $top_k $length)
	((text-generation) "gen_gpt_2" "Query"
		(Kwargs (start_text $start_text)
			(run_name $run_name) (temperature $temperature)
			(top_k $top_k) (length $length))))

! (assertEqual
	(match &self
		(Parent Tom $x)
		$x)
	(empty)
	)

!(assertEqual
	(exercise_1_3 1 2 5)
	29)

(= (better-sqrt-iter $oldguess $guess $x)
	(if (better-good-enough? $oldguess $guess $x)
		$guess
		(better-sqrt-iter $guess (improve $guess $x) $x)
		))

(= (first-denomination $kinds-of-coins)
	(case $kinds-of-coins
		( (1 1) (2 5) (3 10) (4 25) (5 50) )
		))

(= (cc $amount $kinds-of-coins)
	(if (== $amount 0)
		1 (if
			(or (< $amount 0) (== $kinds-of-coins 0))
			0 (+ (cc $amount (- $kinds-of-coins 1))
				(cc (- $amount (first-denomination $kinds-of-coins)) $kinds-of-coins))
			)
		))

(= (timed-prime-test $n) (let*
		( (() (println! $n)) (() (let $curtime
					(timems!)
					(start-prime-test $n $curtime)
					)))
		() ))

(= (fixed-point-print $f $first-guess)
	(let $try
		(lambda2 $guess $self-try (let $next
				($f $guess)
				(let $close-enough?
					(lambda2 $v1 $v2 (< (Abs (- $v1 $v2))
							(tolerance)))
					(if
						($close-enough? $guess $next)
						$next (let*
							( (() (println! $next)) (() ($self-try $next $self-try)) )
							() )
						)
					)
				))
		($try $first-guess $try)
		))

(= (cond $com_list) (if
		(== $com_list ())
		(empty)
		(case
			(car-atom $com_list)
			( ((Else $res) $res) (($cond $res)
					(if $cond $res
						(cond (cdr-atom $com_list))
						)) )
			)
		))
```


DaddyWesker:
I've checked, code works if launched like this:
!(bind! &generative-lms
  (snet-sdk create_service_client "naint" "generative-lms"))

!(snet-sdk get_service_callers_text &generative-lms)

python/hyperon/exts/snet_io/snet_gate.py
```diff
@@ -88,6 +88,8 @@ def __call__(self, command_a, *args_a):
             self.init_sdk()
         if command == 'get_service_callers':
             return args[0].generate_callers()
+        if command == 'get_service_callers_text':
```
Necr0x0Der, line 91: there was no need to expose this function to metta

python/hyperon/exts/snet_io/snet_gate.py
```diff
@@ -137,9 +139,105 @@ def get_service_messages(self):
...
+            len_to_previous_break += 1
+        pretty_res = pretty_res.replace("  ", " ")
+        pretty_res = pretty_res.replace("\t ", "\t")
+        pretty_res = pretty_res.replace("\n ", "\n")
+        return pretty_res
+
     def generate_callers_text(self):
-        # TODO: pretty print
-        return "\n".join([repr(e) for e in self.generate_callers()])
+        res = "\n".join([repr(e) for e in self.generate_callers()])
```
Necr0x0Der, line 239: I don't see pretty-print is used here

python/hyperon/exts/snet_io/snet_gate.py
```diff
@@ -137,9 +139,105 @@ def get_service_messages(self):
...
+        return [ValueAtom(res)]
```
Necr0x0Der, line 240: there is no need to wrap the result of this function into atom since it is supposed as a pure Python function, and it can be called via py-dot from metta

python/hyperon/exts/snet_io/snet_gate.py
```diff
@@ -137,9 +139,105 @@ def get_service_messages(self):
     def get_operation_atom(self):
         return OperationAtom(self.service_details[1], self)
 
+    def __pretty_print__(self, input_str):
```
Necr0x0Der, line 142: pretty-print was supposed for adding to stdlib, but maybe it can remain here for now....

python/hyperon/exts/snet_io/snet_gate.py
```diff
@@ -137,9 +139,105 @@ def get_service_messages(self):
...
+        list_of_functions = ["if", "match", "let", "let*", "unify",
```
Necr0x0Der, line 143: what is special about these functions?

python/hyperon/exts/snet_io/snet_gate.py
DaddyWesker: Sorry, I wasn't been aware of this. Where should I place it?

python/hyperon/exts/snet_io/snet_gate.py
DaddyWesker: Those are functions which could have several inner brackets inside their body. Probably I could forget some functions.

python/hyperon/exts/snet_io/snet_gate.py
DaddyWesker: Ok.

python/hyperon/exts/snet_io/snet_gate.py
DaddyWesker: Ok, i removed those lines.

python/hyperon/exts/snet_io/snet_gate.py
Necr0x0Der: What's about just nested expressions like `(* (- $x $y) (- $x $y))`? What is the substantial difference?

Necr0x0Der: I suspect that it is problematic to detect nested brackets in your approach, because you try to reformat a textual representation, so you need to re-parse it. If you'd started with atoms, it could be simpler...

DaddyWesker:
Okay, so I've rewrote pretty_print so it takes atoms as input not string. Below are some examples of how pretty_print prints atoms into string. @Necr0x0Der  check if it is fine with you (when you have time of course).

```
(: generate
	(-> String String))

(= (generate $request)
	(generative-lms "generate" "Query"
	 (Kwargs
			(request $request))))

(= (abs $x)
	(if (> $x 0) $x
		(if (== $x 0) 0
			(* $x -1))))

(= (response)
	(if (== (messages)
			(user "match world"))
		("Hi there")
		((Bye))))

(assertEqual
	(match &self
		(Parent Tom $x) $x)
	(empty))

(assertEqual
	(exercise_1_3 1 2 5) 29)

(= (better-sqrt-iter $oldguess $guess $x)
	(if (better-good-enough? $oldguess $guess $x) $guess
		(better-sqrt-iter $guess
			(improve $guess $x) $x)))

(= (first-denomination $kinds-of-coins)
	(case $kinds-of-coins
		((1 1) (2 5) (3 10) (4 25) (5 50))))

(= (cc $amount $kinds-of-coins)
	(if (== $amount 0) 1
		(if (or (< $amount 0)
				(== $kinds-of-coins 0)) 0
			(+ (cc $amount
					(- $kinds-of-coins 1))
				(cc (- $amount
						(first-denomination $kinds-of-coins)) $kinds-of-coins)))))

(= (timed-prime-test $n)
	(let* ((() (println! $n))
			(() (let $curtime
					(timems!)
					(start-prime-test $n $curtime))))
		()))

(= (fixed-point-print $f $first-guess)
	(let $try
		(lambda2 $guess $self-try
			(let $next
				($f $guess)
				(let $close-enough?
					(lambda2 $v1 $v2
						(< (Abs (- $v1 $v2))
							(tolerance)))
					(if ($close-enough? $guess $next) $next
						(let* ((() (println! $next))
								(() ($self-try $next $self-try)))
							()))))) ($try $first-guess $try)))

(= (cond $com_list)
	(if (== $com_list
			()) (empty) (case (car-atom $com_list)
			(((Else $res) $res)
				(($cond $res)
					(if $cond $res
						(cond (cdr-atom $com_list))))))))
```

python/hyperon/exts/snet_io/snet_gate.py
```diff
@@ -139,14 +143,58 @@ def get_service_messages(self):
     def get_operation_atom(self):
         return OperationAtom(self.service_details[1], self)
 
+    def __pretty_print_atoms__(self, input_atoms):
```
Necr0x0Der, line 146: I'd move this function out of `ServiceCall` and wouldn't introduce `len_threshold` and `current_len` as its fields.

DaddyWesker: Okay, where to move this function? As i remember, you've proposed to move it to stdlib. Should i place it there?

Necr0x0Der: You can keep it in the same file for now. Just move it out of the class.

DaddyWesker: @Necr0x0Der  Done.

python/hyperon/exts/snet_io/snet_gate.py
```diff
@@ -100,7 +101,62 @@ def __call__(self, command_a, *args_a):
             service_client = self.create_service_client(*args, **kwargs)
             return [service_client.get_operation_atom()]
         return [E(S('Error'), E(S('snet-sdk'), command_a, *args_a),
-                    ValueAtom(f'unknown command {repr(command_a)}'))]
+                  ValueAtom(f'unknown command {repr(command_a)}'))]
+
+len_threshold = 50
+current_len = 0
```
Necr0x0Der, line 107: What will happen when the function is called twice? Will `current_len` behave correctly?

DaddyWesker: Well, it should be set to 0 (line 158 sets it to 0 after every subatom was processed). I'll check just to be sure.

DaddyWesker: @Necr0x0Der  I've checked (though I needed to do some changes to init_sdk since snet-sdk was updated and now its init is slightly different, probably i need to make new pr for those changes?) and current_len is 0 on the start of pretty print. And it can be seen on line 154.

Necr0x0Der: OK, but still... Are they used only in `pretty_print_atoms` and its subfunctions? In this case, it's better to not declare them as `global`, but rather introduce them in `pretty_print_atoms` and use `nonlocal` for them in subfunctions

DaddyWesker: I'll try this out. I haven't been aware of `nonlocal`. Of course it would be better to introduce them inside function I just wasn't been able to make them work correctly. If `nonlocal `will help, I'll push changes tomorrow.

DaddyWesker: @Necr0x0Der  Okay, it works. Thank you. I've pushed new commit with changes regarding current_len and threshold

Necr0x0Der (approved):

PR #772, opened 2024-09-16
open channel and deposit added

DaddyWesker:
So, I've added possibility to open channel and deposit funds via metta sdk. Like this:
`!(snet-sdk open_channel_and_deposit &generative-lms 100 10000)`

### DIFF
--- a/python/hyperon/exts/snet_io/snet_gate.py
+++ b/python/hyperon/exts/snet_io/snet_gate.py
@@ -66,8 +66,6 @@ class SNetSDKWrapper:
         return [G(space)]
 
     def __call__(self, command_a, *args_a):
-        import time
-        time.sleep(5)
         command = self._unwrap_atom(command_a)
         args = []
         kwargs = {}
--- a/python/hyperon/exts/snet_io/snet_gate.py
+++ b/python/hyperon/exts/snet_io/snet_gate.py
@@ -66,6 +66,8 @@ class SNetSDKWrapper:
         return [G(space)]
 
     def __call__(self, command_a, *args_a):
+        import time
+        time.sleep(5)
         command = self._unwrap_atom(command_a)
         args = []
         kwargs = {}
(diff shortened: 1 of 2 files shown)


python/hyperon/exts/snet_io/snet_gate.py
```diff
@@ -66,6 +66,8 @@ def create_service_space(self, org_id, service_id, **kwargs):
         return [G(space)]
 
     def __call__(self, command_a, *args_a):
+        import time
+        time.sleep(5)
```
Necr0x0Der, line 70: Why is it needed? It doesn't look right

python/hyperon/exts/snet_io/snet_gate.py
```diff
@@ -137,6 +141,10 @@ def get_service_messages(self):
     def get_operation_atom(self):
         return OperationAtom(self.service_details[1], self)
 
+    def open_channel_and_deposit(self, amount, expiration):
+        self.service_client.deposit_and_open_channel(amount, expiration)
```
Necr0x0Der, line 145: it would be nice to have an example of usage of this function

python/hyperon/exts/snet_io/snet_gate.py
DaddyWesker: It is not needed. I've deleted it.

python/hyperon/exts/snet_io/snet_gate.py
DaddyWesker: So, should I add an example to the test_snet_call.metta?

Necr0x0Der: yes, but make it inactive (no `!`)

DaddyWesker: Ok. Ive' added. Please check

Necr0x0Der (approved):

PR #783, opened 2024-10-14
Fix newline

### LINKED ISSUE #780: String parsing with new lines is broken
**Describe the bug**
After this PR:
https://github.com/trueagi-io/hyperon-experimental/pull/777/
unit tests in metta-motto became broken.

**To Reproduce**
Run the test in metta-motto:
```
!(assertEqual
   ((echo-agent)
    (system "Ping") (user "Pong"))
"system Ping
user Pong")
```
While `echo-agent` returns
```
"system Ping
user Pong"
```
which is composed as `'\n'.join("system Ping", "user Pong")`, it cannot be matched with the string formed by parsing in Metta
```
"system Ping
user Pong"
```
This:
```
!(assertEqual
   ((echo-agent)
    (system "Ping") (user "Pong"))
"system Ping\nuser Pong")
```
also doesn't work.

**Expected behavior**
There should be a way to specify multi-line strings in metta scripts, which could be matched against corresponding strings formed in Python.

vsbogd:
Fix #780 allow recognizing `\n` by `.` inside regex using flag `s`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ edition = "2021"
 
 [workspace.dependencies]
 hyperon = { path = "./lib", version = "0.2.0" }
-regex = "1.5.4"
+regex = "1.11.0"
 log = "0.4.0"
 env_logger = "0.8.4"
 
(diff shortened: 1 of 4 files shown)


Necr0x0Der (approved):
Thanks!

PR #784, opened 2024-10-21
dummy threads in agents

Necr0x0Der:


Necr0x0Der:
@besSveta , could you please check that this PR breaks nothing in metta-motto

besSveta (approved):
metta-motto works

PR #785, opened 2024-10-21
Fix unique

### LINKED ISSUE #760: Subtraction, union, intersection and unique doesn't work after switching to minimal MeTTa
**Describe the bug**

`subtraction`, `union`, `intersection` and `unique` doesn't work after switching to minimal MeTTa:

**To Reproduce**
1. run `!(subtraction (superpose (a b d)) (intersection (superpose (a b c)) (superpose (a b d))))`

**Expected behavior**
Result is `[d]`

**Actual behavior**
Result is `[d, b, a]`

**Additional context**
 - `metta-repl 0.2.0`

vsbogd:
Fixes #760 
Also removes interpreter from the functions implementations.

### DIFF
--- a/lib/src/metta/runner/stdlib.metta
+++ b/lib/src/metta/runner/stdlib.metta
@@ -67,6 +67,47 @@
 (: empty (-> %Undefined%))
 (= (empty) (let a b never-happens))
 
+(@doc unique
+  (@desc "Function takes non-deterministic input (first argument) and returns only unique entities. E.g. (unique (superpose (a b c d d))) -> [a, b, c, d]")
+  (@params (
+    (@param "Non-deterministic set of values")))
+  (@return "Unique values from input set"))
+(: unique (-> Atom Atom))
+(= (unique $arg) (let $c (collapse $arg) (let $u (unique-atom $c) (superpose $u))))
+
+(@doc union
+  (@desc "Function takes two non-deterministic inputs (first and second argument) and returns their union. E.g. (union (superpose (a b b c)) (superpose (b c c d))) -> [a, b, b, c, b, c, c, d]")
+  (@params (
+    (@param "Non-deterministic set of values")
+    (@param "Another non-deterministic set of values")))
+  (@return "Union of sets"))
+(: union (-> Atom Atom Atom))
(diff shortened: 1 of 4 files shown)


ngeiswei:
Works like a charm, thanks!

ngeiswei (dismissed):

ngeiswei:
Unfortunately I spoke too fast :-(, it does not work in my case, but it is a complex one, I want to simplify it a bunch before pasting it here.  But to give you an idea, the following
```scheme
!(unique (bc &kb
             (fromNumber 5)
             (: $prg (-> (: $x (SongIn "English")) (SongIn "Chinese")))))
```
does not work, while the following
```
!(let* (($results (collapse
                   (bc &kb
                       (fromNumber 5)
                       (: $prg (-> (: $x (SongIn "English")) (SongIn "Chinese"))))))
        ($uniq-results (unique (superpose $results))))
   $uniq-results)
```
does.  Meaning that I need to explicitly collapse the results, then call `unique` on their superposition, and then it works.

I am not sure if it is a bug or a feature, I can still use that, but of course it is not as elegant as just calling `unique` on my non-deterministic function call.

ngeiswei:
Please find a simplified example
```scheme
(= (f) a)
(= (f) a)

!(unique (f))
!(let $y (collapse (f)) (unique (superpose $y)))
```

vsbogd:
Thanks @ngeiswei. Seems like it is a bug which prevented me to implement it last time. `unique` is implemented in stdlib and thus when `collapse` is called from implementation it uses `stdlib` space as a context. As consequence `(f)` is not evaluated and stays `(f)` because stdlib doesn't have definition of `(f)`. It is the reason why `unique` returns `(f)` and after collapse it is evaluated in two instances of `a`.

vsbogd:
Well, I think we can merge it, because it is better then previous state. Then I will think how to fix it properly and raise another PR.

ngeiswei (approved):

vsbogd:
Btw, @ngeiswei , the following code works:
```
(= (f) a)
(= (f) a)

!(unique (capture (f)))
```

PR #789, opened 2024-10-25
Index atom

DaddyWesker:
Index-atom function added. Usage:
```
!(index-atom (5 2) 1)
[2]
```

So it allows to get atom from expression (created by cons-atom) using index.


lib/src/metta/runner/stdlib.rs
```diff
@@ -1160,6 +1160,40 @@ impl CustomExecute for IntersectionAtomOp {
...
+        let index = atom_to_string(atom).parse::<i32>().unwrap();
+        if index < 0 {
+            println!("Negative indexes currently not supported");
+            return unit_result()
+        }
+        else if (index as usize) >= children.len() {
+            println!("Index is out of bounds");
+            return unit_result()
+        }
+        Ok(vec![children.get(index as usize).clone().expect("Unknown error").clone()])
```
vsbogd, line 1193: 1. I would use `Number` as an index argument instead of converting atom via string
2. I would use internal logic of the `Vec::get` method to check index bounds
3. It is more proper to return `Error` instead of printing it on the screen. In the former case caller can check for the error automatically:
```suggestion
        let children = TryInto::<&ExpressionAtom>::try_into(args.get(0).ok_or_else(arg_error)?)?.children();
        let index = AsPrimitive::from_atom(args.get(1).ok_or_else(arg_error)?).as_number().ok_or_else(arg_error)?;
        match children.get(Into::<i64>::into(index) as usize) {
            Some(atom) => Ok(vec![atom.clone()]),
            None => Err(ExecError::from("Index is our of bounds")),
        }
```
In order to make this code compile one should also make `AsPrimitive` public inside `arithmetics.rs`

lib/src/metta/runner/stdlib.rs
```diff
@@ -1160,6 +1160,40 @@ impl CustomExecute for IntersectionAtomOp {
...
 }
 
+#[derive(Clone, Debug)]
+pub struct IndexAtomOp {}
+
+grounded_op!(IndexAtomOp, "index-atom");
+
+impl Grounded for IndexAtomOp {
+    fn type_(&self) -> Atom {
+        Atom::expr([ARROW_SYMBOL, ATOM_TYPE_EXPRESSION, ATOM_TYPE_ATOM, ATOM_TYPE_ATOM])
```
vsbogd, line 1170: I would use `Number` as an index type like in all arithmetic operations.
```suggestion
        Atom::expr([ARROW_SYMBOL, ATOM_TYPE_EXPRESSION, ATOM_TYPE_NUMBER, ATOM_TYPE_ATOM])
```

lib/src/metta/runner/stdlib.rs
vsbogd, line 1: Please add:
1. Rust unit tests to check it works
2. Documentation into both `stdlib.metta` and `stdlib_minimal.metta`

vsbogd:
Why do you have commits from other branches in this one?

DaddyWesker:
> Why do you have commits from other branches in this one?

I'm working in several branches simultaneously and somehow it got really messy =(

DaddyWesker:
Thanks for your comments. I'll look into them and make suggested changes.

PR #791, opened 2024-10-31
random-float and random-int added

DaddyWesker:
I've added random-float and random-int to stdlib. Also I've added documentation and tests. Though I'm not sure how to test anything except error since number is random. Any possible way?

Usage is simple:
!(random-int 0 5)
!(random-float 0 10)

### DIFF
--- a/lib/src/metta/runner/stdlib.rs
+++ b/lib/src/metta/runner/stdlib.rs
@@ -1259,6 +1259,7 @@ impl CustomExecute for SubtractionAtomOp {
 // big. In future releases (0.9+) of rand signature of sample_single will be changed and it will be
 // possible to use match construction to cover overflow and other errors. So after library will be
 // upgraded RandomInt and RandomFloat codes should be altered.
+// see comment https://github.com/trueagi-io/hyperon-experimental/pull/791#discussion_r1824355414
 #[derive(Clone, Debug)]
 pub struct RandomIntOp {}
 
--- a/lib/src/metta/runner/stdlib.rs
+++ b/lib/src/metta/runner/stdlib.rs
@@ -1254,6 +1254,11 @@ impl CustomExecute for SubtractionAtomOp {
     }
 }
 
+
+//TODO: In the current version of rand it is possible for rust to hang if range end's value is too
+// big. In future releases (0.9+) of rand signature of sample_single will be changed and it will be
+// possible to use match construction to cover overflow and other errors. So after library will be
+// upgraded RandomInt and RandomFloat codes should be altered.
 #[derive(Clone, Debug)]
 pub struct RandomIntOp {}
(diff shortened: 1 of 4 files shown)


lib/src/metta/runner/stdlib.rs
```diff
@@ -1253,6 +1254,64 @@ impl CustomExecute for SubtractionAtomOp {
...
+    fn execute(&self, args: &[Atom]) -> Result<Vec<Atom>, ExecError> {
+        let arg_error = || ExecError::from("random-float expects two arguments: number (start) and number (end)");
+        let start = AsPrimitive::from_atom(args.get(0).ok_or_else(arg_error)?).as_number().ok_or_else(arg_error)?;
+        let end = AsPrimitive::from_atom(args.get(1).ok_or_else(arg_error)?).as_number().ok_or_else(arg_error)?;
+        let range = Into::<f64>::into(start)..Into::<f64>::into(end);
+        if range.is_empty() {
+            return Err(ExecError::from("Wrong range"));
+        }
+        let mut rng = rand::thread_rng();
+        Ok(vec![Atom::value(rng.gen_range::<f64,_>(range))])
```
vsbogd, line 1311: Version `0.9.0` of `rand` changes api by adding `Result` into `SampleRange::sample_single` signature (see [docs](https://docs.rs/rand/0.9.0-alpha.2/rand/distr/uniform/trait.SampleRange.html)). Thus it will be possible to do `match range.sample_single(rng) {` and check the result of sampling. It will cover overflow error as well as some others. I would suggest adding comment about it and waiting for the library upgrade.

lib/src/metta/runner/stdlib.rs
```diff
@@ -1253,6 +1254,69 @@ impl CustomExecute for SubtractionAtomOp {
     }
 }
 
+
+//TODO: In the current version of rand it is possible for rust to hang if range end's value is too
+// big. In future releases (0.9+) of rand signature of sample_single will be changed and it will be
+// possible to use match construction to cover overflow and other errors. So after library will be
+// upgraded RandomInt and RandomFloat codes should be altered.
```
vsbogd, line 1261:
```suggestion
//TODO: In the current version of rand it is possible for rust to hang if range end's value is too
// big. In future releases (0.9+) of rand signature of sample_single will be changed and it will be
// possible to use match construction to cover overflow and other errors. So after library will be
// upgraded RandomInt and RandomFloat codes should be altered.
// see comment https://github.com/trueagi-io/hyperon-experimental/pull/791#discussion_r1824355414
```

vsbogd (dismissed):

vsbogd (approved):
Merging PR to fix the issue with Rust/Python conversion afterwards

PR #793, opened 2024-11-07
BaseListeningAgent

besSveta:
BaseListeningAgent is an example of an agent that processes incoming messages in a separate thread.


python/hyperon/exts/agents/agent_base.py
```diff
@@ -40,7 +40,40 @@ def __next__(self):
             raise StopIteration
         return self._result.get()
 
-class AgentObject:
+
+class BaseListeningAgent:
```
Necr0x0Der, line 44: It doesn't look like `BaseListeningAgent` will be used by itself. Why not create a `ListeningAgent` which inherits from `AgentObject` instead of inheriting `AgentObject` from `BaseListeningAgent`? I'm not sure if the functionality from `BaseListeningAgent` is always needed and will not disturb some uses of `AgentObject`. Inheriting `ListeningAgent` from `AgentObject` looks more flexible to me. Or am I missing something?

python/hyperon/exts/agents/agent_base.py
```diff
@@ -166,6 +170,41 @@ def __metta_call__(self, *args):
             return [E()]
         return st
 
+class BaseListeningAgent(AgentObject):
+    def __init__(self, path=None, atoms={}, include_paths=None, code=None):
+        super().__init__(path, atoms, include_paths, code)
+        self.messages = Queue()
+        self.running = True
```
Necr0x0Der, line 177: Should the agent be `running` after `__init__`, or should this happens after `start`?

PR #794, opened 2024-11-07
Fix native interface of the Python serializer

vsbogd:
Fix names of the functions inside pybind11 macro stub.
@DaddyWesker should fix example from #791


DaddyWesker (approved):
Works for me. Thank you.

PR #802, opened 2024-11-18
Math operations

DaddyWesker:
So, I've added some math functions, also is_nan and is_infinite functions. It is draft for now since I want @CICS-Oleg to tell me if he wants something else here. 
Anyway, @vsbogd please check if everything is fine.

### DIFF
--- a/lib/src/metta/runner/stdlib_minimal.rs
+++ b/lib/src/metta/runner/stdlib_minimal.rs
@@ -1299,6 +1299,29 @@ impl CustomExecute for SqrtOp {
 }
 
 #[derive(Clone, Debug)]
+pub struct AbsOp {}
+
+grounded_op!(AbsOp, "abs");
+
+impl Grounded for AbsOp {
+    fn type_(&self) -> Atom {
+        Atom::expr([ARROW_SYMBOL, ATOM_TYPE_NUMBER, ATOM_TYPE_NUMBER])
+    }
+
+    fn as_execute(&self) -> Option<&dyn CustomExecute> {
+        Some(self)
+    }
+}
+
+impl CustomExecute for AbsOp {
+    fn execute(&self, args: &[Atom]) -> Result<Vec<Atom>, ExecError> {
+        let arg_error = || ExecError::from("abs expects one argument: number");
+        let input = Into::<f64>::into(AsPrimitive::from_atom(args.get(0).ok_or_else(arg_error)?).as_number().ok_or_else(arg_error)?);
+        Ok(vec![Atom::gnd(Number::Float(input.abs()))])
+    }
+}
+
+#[derive(Clone, Debug)]
 pub struct SubtractionAtomOp {}
 
 grounded_op!(SubtractionAtomOp, "subtraction-atom");
@@ -1851,6 +1874,8 @@ pub fn register_common_tokens(tref: &mut Tokenizer, _tokenizer: Shared<Tokenizer
     tref.register_token(regex(r"powf"), move |_| { powf_op.clone() });
     let sqrt_op = Atom::gnd(SqrtOp{});
     tref.register_token(regex(r"sqrt"), move |_| { sqrt_op.clone() });
+    let abs_op = Atom::gnd(AbsOp{});
+    tref.register_token(regex(r"abs"), move |_| { abs_op.clone() });
     let index_atom_op = Atom::gnd(IndexAtomOp{});
     tref.register_token(regex(r"index-atom"), move |_| { index_atom_op.clone() });
     let random_int_op = Atom::gnd(RandomIntOp{});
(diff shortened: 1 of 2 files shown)


CICS-Oleg (commented):
I think it would be awesome if you can add some math constants like pi and e too.

CICS-Oleg (dismissed):

lib/src/metta/runner/stdlib_math.rs
```diff
@@ -0,0 +1,730 @@
...
+
+    fn as_execute(&self) -> Option<&dyn CustomExecute> {
+        Some(self)
+    }
+}
+
+impl CustomExecute for SqrtMathOp {
+    fn execute(&self, args: &[Atom]) -> Result<Vec<Atom>, ExecError> {
+        let arg_error = || ExecError::from("sqrt-math expects one argument: number");
+        let input = Into::<f64>::into(args.get(0).and_then(Number::from_atom).ok_or_else(arg_error)?);
```
vsbogd, line 94: Just to make code a little bit more readable you can use here and  below:
```suggestion
        let input: f64 = args.get(0).and_then(Number::from_atom).ok_or_else(arg_error)?.into();
```

lib/src/metta/runner/stdlib_minimal.rs
```diff
@@ -1785,6 +1786,44 @@ pub fn register_common_tokens(tref: &mut Tokenizer, _tokenizer: Shared<Tokenizer
...
+    let atan_math_op = Atom::gnd(stdlib_math::AtanMathOp{});
+    tref.register_token(regex(r"atan-math"), move |_| { atan_math_op.clone() });
+    let isnan_math_op = Atom::gnd(stdlib_math::IsNanMathOp{});
+    tref.register_token(regex(r"isnan-math"), move |_| { isnan_math_op.clone() });
+    let isinf_math_op = Atom::gnd(stdlib_math::IsInfMathOp{});
+    tref.register_token(regex(r"isinf-math"), move |_| { isinf_math_op.clone() });
+    tref.register_token(regex(r"const_pi"),
+                        |_| { Atom::gnd(Number::Float(std::f64::consts::PI)) });
+    tref.register_token(regex(r"const_e"),
+                        |_| { Atom::gnd(Number::Float(std::f64::consts::E)) });
```
vsbogd, line 1826: I would use more intuitive naming. May be `PI` and `EXP`?

lib/src/metta/runner/stdlib_math.rs
```diff
@@ -0,0 +1,730 @@
...
+        Ok(vec![Atom::gnd(Number::Float(input.sqrt()))])
```
vsbogd, line 95: I would suggest returning error rather than if input is negative number.

lib/src/metta/runner/stdlib_math.rs
```diff
@@ -0,0 +1,730 @@
...
+    fn as_execute(&self) -> Option<&dyn CustomExecute> {
+        Some(self)
+    }
+}
+
+impl CustomExecute for AbsMathOp {
+    fn execute(&self, args: &[Atom]) -> Result<Vec<Atom>, ExecError> {
+        let arg_error = || ExecError::from("abs-math expects one argument: number");
+        let input = Into::<f64>::into(args.get(0).and_then(Number::from_atom).ok_or_else(arg_error)?);
+        Ok(vec![Atom::gnd(Number::Float(input.abs()))])
```
vsbogd, line 118: `abs` in Rust is defined for both `i64` and `f64` thus you can again check `Number` variant to select proper implementation.

lib/src/metta/runner/stdlib_math.rs
DaddyWesker: Currently sqrt returns NaN if number is below zero. I thought we've discussed that we will leave functions to return NaN but add is_nan function. Or...?

lib/src/metta/runner/stdlib_math.rs
DaddyWesker: Difference between abs for int and abs for float is in performance?

lib/src/metta/runner/stdlib_math.rs
vsbogd: Yeah, we have discussed it but probably it was a bit abstract to me. I am not sure why Rust API looks in this way, probably because it is a traditional implementation. But for `sqrt` specifically I think returning error looks more natural.

lib/src/metta/runner/stdlib_math.rs
vsbogd: No, just because I think it is better to keep `i64` as `i64` not promote it to `f64`. Promoting it may lead to the precision loss.

DaddyWesker: Ok.

lib/src/metta/runner/stdlib_math.rs
vsbogd: After discussing it offline, we decided to keep it as is, to make it behavior similar to native implementation.

lib/src/metta/runner/stdlib_minimal/mod.rs
```diff
@@ -1785,6 +1790,42 @@ pub fn register_common_tokens(tref: &mut Tokenizer, _tokenizer: Shared<Tokenizer
...
+    let atan_math_op = Atom::gnd(stdlib_math::AtanMathOp{});
+    tref.register_token(regex(r"atan-math"), move |_| { atan_math_op.clone() });
+    let isnan_math_op = Atom::gnd(stdlib_math::IsNanMathOp{});
+    tref.register_token(regex(r"isnan-math"), move |_| { isnan_math_op.clone() });
+    let isinf_math_op = Atom::gnd(stdlib_math::IsInfMathOp{});
+    tref.register_token(regex(r"isinf-math"), move |_| { isinf_math_op.clone() });
+    tref.register_token(regex(r"consts_PI"),
+                        |_| { Atom::gnd(Number::Float(std::f64::consts::PI)) });
+    tref.register_token(regex(r"consts_EXP"),
+                        |_| { Atom::gnd(Number::Float(std::f64::consts::E)) });
```
vsbogd, line 1828: In fact I suggested removing prefix `const_`. Don't see if we need it actually.
```suggestion
    tref.register_token(regex(r"PI"),
                        |_| { Atom::gnd(Number::Float(std::f64::consts::PI)) });
    tref.register_token(regex(r"EXP"),
                        |_| { Atom::gnd(Number::Float(std::f64::consts::E)) });
```

DaddyWesker: Ok, done.

vsbogd (approved):

PR #806, opened 2024-11-21
Separation of stdlib

DaddyWesker:
So, some modularity for stdlib is here. I've created two new modules for common tokens and runner tokens and extracted functions to those modules alongside with tests for those functions. 

I've also made run_program in stdlib_minimal/mod.rs::tests pub to use it in separate modules' tests instead of copy-pasting run_program to each new module. I'm not sure if this is right decision, but it seems more elegant to me.

### DIFF
--- a/lib/src/metta/interpreter_minimal.rs
+++ b/lib/src/metta/interpreter_minimal.rs
@@ -6,7 +6,7 @@ use crate::atom::matcher::*;
 use crate::space::*;
 use crate::metta::*;
 use crate::metta::types::*;
-use crate::metta::runner::stdlib_minimal::IfEqualOp;
+use crate::metta::runner::stdlib_minimal::stdlib_common_tokens::IfEqualOp;
 use crate::common::collections::CowArray;
 
 use std::fmt::{Debug, Display, Formatter};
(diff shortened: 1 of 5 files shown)


vsbogd (commented):
Probably I should be more specific and provide some examples when we were discussing it. Separating functions in `stdlib` on "common tokens" and "runner tokens" sets doesn't make much sense because it is technical difference. "common" set of tokens is registered only once on loading root MeTTa runner because they doesn't depend on `space` and `metta`. "runner tokens" should be reloaded each time new MeTTa module is imported. Thus for example `assert...` and `import!` are both in "runner tokens" set while semantically they are different.

Thus I would suggest starting from scratch and consider functions from both "common" and "runner" token sets as a single list. I would suggest then make a commit to finally separate "math" module. We could add a `register_common_tokens()` function implementation into `stdlib_math.rs` and put all math tokens into it. then call this new `stdlib_math::register_common_tokens()` from `stdlib::register_common_tokens()`. After this `stdlib_math` module will be more or less separated. 

Then I would start trying to figure out which new kinds of modules can be added. For example:
- `random` module: `random-...` and `flip`
-  `atom` module: `get-...type...`, `...-atom`, `sealed`
-  `module` module: `import!`, `include`, `bind!`, `mod-space!`, `print-mods!`
- `package` module: `pkg_mgmt` functions;
- `string` module: `println!`, `format-args`
- `debug` module: `trace!`, `print-alternatives!`
- `space` module: `new-space`, `add-atom`, etc, `get-atoms`

BTW, `stdlib_minimal/mod.rs` includes unit tests for `stdlib_minimal.metta` part of library. Thus tests for functions like `cons-atom` could also be moved into `atom` module, etc

PR #807, opened 2024-11-22
final separation of math

DaddyWesker:
So, now math should be fully separated from mod.rs. And some more.


lib/src/metta/mod.rs
```diff
@@ -1,8 +1,8 @@
 //! Contains MeTTa specific types, constants and functions.
 
 pub mod text;
-pub mod interpreter_minimal;
-pub use interpreter_minimal as interpreter;
+pub mod interpreter;
+// pub use interpreter_minimal as interpreter;
```
vsbogd, line 5:
```suggestion
```

vsbogd (approved):
Looks good to me, except one commented line could be removed (see other comment).

vsbogd (approved):

PR #809, opened 2024-11-22
Use rust primitive types in Python runtime

### LINKED ISSUE #797: c1_grounded_basic.metta and d2_higherfunc.metta MeTTa test scenarios fail with Rust interpreter
**Describe the bug**
MeTTa test scenarios `c1_grounded_basic.metta` and `d2_higherfunc.metta` fails when being executed using Rust `metta-repl` tool.

**To Reproduce**
1. `cargo run python/test/scripts/c1_grounded_basic.metta`
2. `cargo run python/test/scripts/d2_higherfunc.metta`
3. See output

**Expected behavior**
All checks are passed.

**Actual behavior**
Some checks fail:
```
$ metta-repl c1_grounded_basic.metta`
[()]
[()]
[()]
[()]
[()]
[()]
[(Error (assertEqualToResult (> 4 (+ ln 2)) ((> 4 (+ ln 2)))) 
Expected: [(> 4 (+ ln 2))]
Got: [(Error (+ ln 2) + expects two number arguments)]
Missed result: (> 4 (+ ln 2)))]

$ metta-repl d2_higherfunc.metta`
[()]
[()]
[()]
[()]
[()]
[()]
[()]
[()]
[()]
[()]
[()]
[(Error (assertEqual ((lambda $x (+ $x 1)) 2) 3) 
Expected: [3]
Got: [(Error (+ $x 1) + expects two number arguments)]
Missed result: 3)]
```

**Additional context**
 - MeTTa version: 0.2.0+gd89d1073.d20240820
 - `f1_imports.metta` scenario fails also, but this is expected because it depends on Python stdlib representation in memory

vsbogd:
Remove Python numbers and boolean values from Python standard library and use Rust primitives.

Introduce `IncorrectArgument` error to return from grounded operations to say interpreter that arguments of the operation cannot be matched properly. Interpreter doesn't reduce such operation similarly to the case when pure MeTTa function is not matched.

Some improvements of the Rust stdlib modules structure.

Fixes #797


vsbogd:
Fixes #797

PR #825, opened 2024-12-18
Return second argument of trace! without evaluation

vsbogd:
Evaluating second argument before calling trace! leads to unexpected sequence of traces. Making type of the second argument Atom means argument is evaluated after trace! prints the first argument.


vsbogd:
@ngeiswei adding you as a reviewer for this one-liner because you are an author of the original code.

PR #826, opened 2024-12-19
Don't copy imported module to filter out its dependencies

vsbogd:
This PR is replacement of #776 without issue which were caused by old interpreter implementation. New commits start from https://github.com/trueagi-io/hyperon-experimental/commit/4277146f621100f3f408bb3bfae724617a57a226

In this PR `ModuleSpace` is introduced which keeps main space and its dependencies separately. This allows querying them separately and when dependent space is queried it doesn't queried recursively. Thus all transitive dependencies are queried only once from the top level space.

For example if we have space `A` which imports space `B` and `C` and `D` is imported from `B` and `C` then `A` contains `B`, `C` and `D` as dependencies in a separate collection. Query to `A` goes through its atoms first then goes to `B`, `C` and `D` but dependencies of these spaces are not queried.

`ModuleSpace::atom_iter()` iterates through main atomspace only. This breaks tests in `f1_imports.metta` which are disabled in this change. There are two ways of solving this: (1) adding dependency spaces as atoms into `ModuleSpace` iterator which resembles previous behavior, (2) add separate function which gets dependencies from space. I would like to implement solution after discussing a way. Option (2) is not possible with current `DynSpace` used internally as space representation may be it can be implemented after additional refactoring.

New `Space::visit` method is added because it is impossible to implement `Space::atom_iter` for `DynSpace` safely (see @luketpeterson  comment https://github.com/trueagi-io/hyperon-experimental/pull/776#issuecomment-2375449225). Internal usages of `Space::atom_iter` are replaced by `Space::visit` usages.

### DIFF
--- a/c/src/space.rs
+++ b/c/src/space.rs
@@ -732,53 +732,23 @@ impl Space for CSpace {
             None
         }
     }
-    fn atom_iter(&self) -> Option<Box<dyn Iterator<Item=&Atom> + '_>> {
-        struct CSpaceIterator<'a>(&'a CSpace, *mut c_void);
-        impl<'a> Iterator for CSpaceIterator<'a> {
-            type Item = &'a Atom;
-            fn next(&mut self) -> Option<&'a Atom> {
-                let api = unsafe{ &*self.0.api };
-                if let Some(next_atom) = api.next_atom {
-                    let atom_ref = next_atom(&self.0.params, self.1);
-                    if atom_ref.is_null() {
-                        None
-                    } else {
-                        Some(atom_ref.into_ref())
-                    }
-                } else {
-                    panic!("next_atom function must be implemented if new_atom_iterator_state is implemented");
-                }
-            }
-        }
-        impl Drop for CSpaceIterator<'_> {
-            fn drop(&mut self) {
-                let api = unsafe{ &*self.0.api };
-                if let Some(free_atom_iterator_state) = api.free_atom_iterator_state {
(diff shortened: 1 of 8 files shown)


Necr0x0Der (approved):

luketpeterson (approved):
Wholeheartedly approve these changes.  :-)

luketpeterson:
> ModuleSpace::atom_iter() iterates through main atomspace only. This breaks tests in f1_imports.metta which are disabled in this change. There are two ways of solving this: (1) adding dependency spaces as atoms into ModuleSpace iterator which resembles previous behavior, (2) add separate function which gets dependencies from space. I would like to implement solution after discussing a way. Option (2) is not possible with current DynSpace used internally as space representation may be it can be implemented after additional refactoring.

As to this question, it gets into the behavior expected from the MeTTa language / modules.  I was assuming behavior 1 was "correct", based on how things worked before the module separation was added.  But a case can be made for either behavior.

vsbogd:
> As to this question, it gets into the behavior expected from the MeTTa language / modules. I was assuming behavior 1 was "correct", based on how things worked before the module separation was added. But a case can be made for either behavior.

Yes, we discussed it with @Necr0x0Der as well. We still can add atomspace which is an atom and at the same time contains functions which are available for execution from the space which contains this atom. Thus one still can do something like "manual import".

Introspection of imported modules is important from my perspective. But it is not critical right now and I hope clarifying this behavior later.

PR #828, opened 2024-12-28
python's repr fixed

### LINKED ISSUE #818: (repr "A") is incorrect
**Describe the bug**

I think the output of `(repr "A")` is incorrect.

**To Reproduce**
Run

```
(repr "A")
```

**Expected behavior**

```
["\"A\""]
```

**Actual behavior**

```
[""A""]
```

**Additional context**
 - MeTTa version: 0.2.1+g8795f59b.d20241129
 - `(repr 'A')` or `(repr A)` is correct, only `(repr "A")` isn't.

DaddyWesker:
Fix #818

### DIFF
--- a/lib/src/metta/runner/stdlib/string.rs
+++ b/lib/src/metta/runner/stdlib/string.rs
@@ -2,7 +2,7 @@ use crate::*;
 use crate::metta::*;
 use crate::metta::text::Tokenizer;
 use crate::metta::runner::str::*;
-use super::{grounded_op, atom_to_string, unit_result, regex};
+use super::{grounded_op, unit_result, regex};
 
 use std::convert::TryInto;
 
@@ -25,7 +25,7 @@ impl CustomExecute for PrintlnOp {
     fn execute(&self, args: &[Atom]) -> Result<Vec<Atom>, ExecError> {
         let arg_error = || ExecError::from("println! expects single atom as an argument");
         let atom = args.get(0).ok_or_else(arg_error)?;
-        println!("{}", atom_to_string(atom));
+        println!("{}", atom_into_string(atom.clone()));
         unit_result()
     }
 }
@@ -50,10 +50,10 @@ impl Grounded for FormatArgsOp {
 impl CustomExecute for FormatArgsOp {
     fn execute(&self, args: &[Atom]) -> Result<Vec<Atom>, ExecError> {
         let arg_error = || ExecError::from("format-args expects format string as a first argument and expression as a second argument");
-        let format = atom_to_string(args.get(0).ok_or_else(arg_error)?);
+        let format = atom_into_string(args.get(0).ok_or_else(arg_error)?.clone());
         let args = TryInto::<&ExpressionAtom>::try_into(args.get(1).ok_or_else(arg_error)?)?;
         let args: Vec<String> = args.children().iter()
-            .map(|atom| atom_to_string(atom))
+            .map(|atom| atom_into_string(atom.clone()))
             .collect();
         let res = format.format(args.as_slice());
         Ok(vec![Atom::gnd(Str::from_string(res))])
--- a/lib/src/metta/runner/stdlib/string.rs
+++ b/lib/src/metta/runner/stdlib/string.rs
@@ -25,7 +25,7 @@ impl CustomExecute for PrintlnOp {
     fn execute(&self, args: &[Atom]) -> Result<Vec<Atom>, ExecError> {
         let arg_error = || ExecError::from("println! expects single atom as an argument");
         let atom = args.get(0).ok_or_else(arg_error)?;
-        println!("{}", atom_into_string(atom.clone()));
+        println!("{}", atom_to_string(atom));
         unit_result()
     }
 }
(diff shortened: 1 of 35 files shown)


python/tests/test_examples.py
```diff
@@ -368,6 +368,7 @@ def test_char_vs_string(self):
         self.assertEqual(repr(metta.run('!("A")')), '[[("A")]]')
         self.assertEqualMettaRunnerResults(metta.run("!(get-type 'A')"), [[S('Char')]])
         self.assertEqualMettaRunnerResults(metta.run('!(get-type "A")'), [[S('String')]])
+        self.assertEqual(metta.run('!(repr "A")')[0][0].get_object(), ValueObject("\"A\""))
```
vsbogd, line 371: Please move test into `test_grounded_type.py` as this is a change in a grounded atom's behaviour.

python/hyperon/atoms.py
```diff
@@ -279,7 +279,8 @@ def __repr__(self):
         # Overwrite Python default representation of a string to use
         # double quotes instead of single quotes.
         if isinstance(self.content, str):
-            return f'"{self.content}"'
+            newstr = self.content.translate(str.maketrans({'"' : r'\"'}))
```
vsbogd, line 282: Here, all symbols which needs to be escaped should be escaped: for instance `"`, `'`, `\`, `<CR>`, `<LF>`, `<TAB>` other symbols which are not visible. I would suggest finding standard Python function to escape them all.

python/tests/test_stdlib.py
```diff
@@ -19,8 +19,8 @@ def test_text_ops(self):
         #self.assertEqualMettaRunnerResults(metta.run('!(parse "$X")'),
         #                                  [[(V("X"))]])
 
-        self.assertEqualMettaRunnerResults(metta.run('!(parse "\\"A\\"")'),
-                                           [[(ValueAtom("A"))]])
+        self.assertEqualMettaRunnerResults(repr(metta.run('!(parse "\\"A\\"")')),
+                                           '[[\\"A\\"]]')
```
vsbogd, line 23: 1. Test should check which atoms are returned by `parse` not what their string representation is
2. At the moment test is passing in the following form:
```
        self.assertEqualMettaRunnerResults(metta.run('!(parse "\\"A\\"")'),
                                           [[(S('\\"A\\"'))]])
```
One can see that now `parse` returns symbol instead of string. It is because `parse` parses `str(atom)` instead of internal string. In this particular case atom is a string `"A"` but `str()` from this atom gives `\"A\"` (this is what is changed by the PR). This sequence is parsed as symbol `\"A\"` because escape sequences are processed by MeTTa parser only inside strings and `\"A\"` is not a string because it starts from `\` symbol. Thus the result of the test is unexpected. To make it proper we need to change `parse` implementation to work with original string not with the result of `str()`. Test should be passed in its original form:
```suggestion
        self.assertEqualMettaRunnerResults(metta.run('!(parse "\\"A"\\")'),
                                           [[(ValueAtom("A"))]])
```

vsbogd (changes requested):

python/tests/test_stdlib.py
vsbogd: To properly solve it, I would suggest change `parse` implementation to extract Python string from grounded atom on input. Should be as simple as calling `.get_object()` on an input atom and checking that this is a really string.

python/tests/test_stdlib.py
```diff
@@ -19,8 +19,8 @@ def test_text_ops(self):
         #self.assertEqualMettaRunnerResults(metta.run('!(parse "$X")'),
         #                                  [[(V("X"))]])
 
-        self.assertEqualMettaRunnerResults(metta.run('!(parse "\\"A\\"")'),
-                                           [[(ValueAtom("A"))]])
+        self.assertEqualMettaRunnerResults(metta.run('!(parse "\\"A"\\")'),
+                                                   [[(ValueAtom("A"))]])
```
vsbogd, line 23: Quote is located in incorrect place:
```suggestion
        self.assertEqualMettaRunnerResults(metta.run('!(parse "\\"A\\"")'),
                                                   [[(ValueAtom("A"))]])
```

lib/src/metta/runner/stdlib/string.rs
```diff
@@ -25,7 +25,7 @@ impl CustomExecute for PrintlnOp {
     fn execute(&self, args: &[Atom]) -> Result<Vec<Atom>, ExecError> {
         let arg_error = || ExecError::from("println! expects single atom as an argument");
         let atom = args.get(0).ok_or_else(arg_error)?;
-        println!("{}", atom_to_string(atom));
+        println!("{}", atom_into_string(atom.clone()));
```
vsbogd, line 28: Cloning here is ineffective.
1. Remove old `atom_to_string` implementation
2. Rename `atom_into_string` to `atom_to_string`
3. Change `atom_to_string` signature to `atom_to_string(atom: &Atom)` to eliminate cloning

DaddyWesker: Done

python/tests/test_stdlib.py
DaddyWesker: Done

vsbogd (approved):

PR #830, opened 2024-12-31
Improve handling of inline comments

AndyZhu1991:
Break the word parsing loop when encountering a semicolon. This change disallows semicolons in symbols.
Additional test cases have been added to validate this behavior.

### DIFF
--- a/lib/src/metta/text.rs
+++ b/lib/src/metta/text.rs
@@ -491,7 +491,7 @@ impl<'a> SExprParser<'a> {
         let start_idx = self.cur_idx();
 
         while let Some((_idx, c)) = self.it.peek() {
-            if c.is_whitespace() || *c == '(' || *c == ')' {
+            if c.is_whitespace() || *c == '(' || *c == ')' || *c == ';' {
                 break;
             }
             token.push(*c);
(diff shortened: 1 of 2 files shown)


lib/src/metta/text.rs
vsbogd, line 1: I would add a unit test here inside `tests` module because core rust library can be used not only from Python:
```rs
    #[test]
    fn test_comment_in_symbol() {
        let program = "(a; 4)
                  5)";
        let expected = vec![expr!("a" "5")];
        let res = parse_atoms(program);
        assert_eq!(res, expected);
    }

```

vsbogd (commented):
@AndyZhu1991 thank you for help. Please add the unit test in a Rust module and I will be happy to merge it.

lib/src/metta/text.rs
AndyZhu1991: Thanks, I will add this test case.

AndyZhu1991:
> @AndyZhu1991 thank you for help. Please add the unit test in a Rust module and I will be happy to merge it.

I have added the unit test for parse_word.

vsbogd (approved):

vsbogd:
Thanks @AndyZhu1991 , happy to merge it.

PR #838, opened 2025-01-20
Unicode unescape

DaddyWesker:
So, unicode parsing is added for `\u{...}` sequence. Also some tests for parser.


lib/src/metta/text.rs
```diff
@@ -486,6 +493,30 @@ impl<'a> SExprParser<'a> {
...
+            cur_idx += 1;
+            if cur_idx > 8 { return None }
+            let char_to_digit = match next_char.to_digit(16) {
+                Some(digit) => digit,
+                None => return None,
+            };
+            result_u32 = (result_u32 << 4) | char_to_digit;
+        }
+        let res_char = char::from_u32(result_u32).unwrap();
+        Some(res_char)
```
vsbogd, line 517:
```suggestion
       char::from_u32(result_u32)
```

lib/src/metta/text.rs
```diff
@@ -486,6 +493,30 @@ impl<'a> SExprParser<'a> {
         }).and_then(|code_val| (code_val <= 0x7F).then(|| code_val))
     }
 
+    fn parse_unicode_sequence(&mut self) -> Option<char> {
+        // unicode sequence presumably looks like this '\\u{0123}'
```
vsbogd, line 497:
```suggestion
        // unicode sequence presumably looks like this '\u{0123}'
```

lib/src/metta/text.rs
```diff
@@ -664,6 +695,23 @@ mod tests {
...
+        let mut parser = SExprParser::new("\"\\u{0123\"");
+        let node = parser.parse_string();
+        assert!(!node.is_complete);
+        assert_eq!("Invalid escape sequence", node.message.unwrap());
+
+        let mut parser = SExprParser::new("\"\\u{012345678\"");
+        let node = parser.parse_string();
+        assert!(!node.is_complete);
+        assert_eq!("Invalid escape sequence", node.message.unwrap());
+
```
vsbogd, line 712: I would also add the following cases:
```
r#""\u0123}""#
r#""\u0123""#
r#""\u0{123}""#
r#""\u{defg}""#
r#""\u{130AC}""#
r#""\u{130ac}""#
```
The expected behaviour can be compared with Rust compiler parser.

PR #839, opened 2025-01-24
Return strip_quotes to prevent unnecessary folder creation

DaddyWesker:
Subj


lib/src/metta/runner/environment.rs
```diff
@@ -471,6 +471,8 @@ fn include_path_from_cfg_atom(atom: &ExpressionAtom, env: &Environment) -> Resul
         None => return Err(format!("Error in environment.metta. #includePath missing path value"))
     };
     let path = <&crate::SymbolAtom>::try_from(path_atom)?.name();
+    //It is crucial to return strip_quotes here since otherwise running Metta will create folder {$cfgdir} and some subfolders.
```
vsbogd, line 474:
```suggestion
    // At this stage stdlib is not loaded and thus path is parsed as a symbol not string. 
    // This is why we should manuall stip quotes from the symbol's name.
```

vsbogd (approved):

PR #848, opened 2025-02-10
set-random-seed added

DaddyWesker:
~~So, I've figured out how to properly use thread_local for creating thread_local global variable in Rust. Usage of `random-int` and `random-float` is the same.~~ ~~Base seed is coming from `StdRng::from_os_rng()`. New `set-random-seed` is used to set a random seed.~~ ~~So when this operation is called, `GLOBAL_RNG` (which is thread local) will be reinitialized with a new seed.~~ 

~~I'm not sure what kind of tests should I provide for `set-random-seed` operation so I've created kinda simple one.~~  

~~Example is quite simple:~~
~~!(random-int 0 5) # just a regular call without pre-defined seed as before
!(set-random-seed! 5) # setting a seed to a random generator
!(random-int 0 5) # now calling random-int after reinitializing generator with a seed.~~

Update. It was decided to refactor API and add some additional functions like `new-random-generator` etc. So API is know looks like this:
1. `new-random-generator` creates random generator instance for further usage. Takes seed as an argument;
2. `random-int\random-float` now takes not only range but also instance of random generator;
3. there is a metta-level (token) random generator with default behavior, e.g. `StdRng::from_os_rng();` which called `&rng` and which can be used just like user-creeated random generator instance;
4. `set-random-seed` accepts number and random generator instance and changes its behavior using random seed;
5. `reset-random-generator` discards all changes to random generator instance and reinitialize it using  `StdRng::from_os_rng();`.

Example of usage
```
!(bind! &newrng (new-random-generator 9)) # 9 is a seed for creating random generator instance
!(random-int &newrng 0 10) # using just created &newrng to generate a number
!(random-float &rng 0 10) # &rng is a global metta-level token
!(set-random-seed &newrng 5) # re-seeding random number generator instance
!(set-random-seed &rng 5) # we can also re-seed global &rng
!(reset-random-generator &rng) # resetting &rng to the default behavior
```


lib/src/metta/runner/stdlib/random.rs
```diff
@@ -66,11 +69,35 @@ impl CustomExecute for RandomFloatOp {
...
-        let mut rng = rand::thread_rng();
-        Ok(vec![Atom::gnd(Number::Float(rng.gen_range(range)))])
+        Ok(vec![Atom::gnd(Number::Float(GLOBAL_RNG.with(|generator| generator.borrow_mut().random_range(range))))])
     }
 }
 
+#[derive(Clone, Debug)]
+pub struct SetRandomSeedOp {}
+
+grounded_op!(SetRandomSeedOp, "set-random-seed");
```
vsbogd, line 79: Name of the operation should end with `!` because it has side effects.

lib/src/metta/runner/stdlib/random.rs
```diff
@@ -6,7 +7,10 @@ use crate::metta::runner::number::*;
 use crate::metta::runner::bool::*;
 
 use std::fmt::Display;
-use rand::Rng;
+use std::cell::RefCell;
+use rand::{Rng, SeedableRng, rngs::StdRng};
+
+thread_local!(static GLOBAL_RNG: RefCell<StdRng> = RefCell::new(StdRng::from_os_rng()));
```
vsbogd, line 13: Lets name it `THREAD_LOCAL_RNG` then

vsbogd (changes requested):
Please add a test which checks that `set-random-seed` leads to deterministic results. Also I would add a code proposed by @Necr0x0Der if you have one.

lib/src/metta/runner/stdlib/random.rs
DaddyWesker: Done

lib/src/metta/runner/stdlib/random.rs
DaddyWesker: Done

lib/src/metta/runner/stdlib/random.rs
```diff
@@ -66,11 +69,35 @@ impl CustomExecute for RandomFloatOp {
...
+    }
+
+    fn as_execute(&self) -> Option<&dyn CustomExecute> {
+        Some(self)
+    }
+}
+
+impl CustomExecute for crate::metta::runner::stdlib::random::SetRandomSeedOp {
+    fn execute(&self, args: &[Atom]) -> Result<Vec<Atom>, ExecError> {
+        let arg_error = || ExecError::from("set-random-seed expects one argument: number (seed)");
```
vsbogd, line 93:
```suggestion
        let arg_error = || ExecError::from("set-random-seed! expects one argument: number (seed)");
```

vsbogd (approved):

Necr0x0Der (dismissed):

vsbogd (approved):

Necr0x0Der (approved):

PR #852, opened 2025-02-20
agent with events

Necr0x0Der:


besSveta (approved):

astroseger (commented):
Review from point view of multi-threading implementation: 

I.
If I understood correctly new thread is created inside
`AgentObject.__metta_call`
and inside
`EventObject.start`

Does it mean that a separate thread is created for each event (and for each __metta_call_ of object)? If yes, then it might be relatively expensive for some particular use cases (on my machine creating/destruction of  1000 threads costs ~0.05 sec + some memory footprint). However I'm not ready right now to propose an alternative.

II.
In
`EventAgent.event_processor` if self.events is empty the thread will consume 100% CPU by running
```
while True:
```

If I understood correctly this code should work, but it will not exit after self.running is set to False. However it should be noted that the current code (code from PR, not what is proposed here) does not guarantee it neither if event_processor can be called from different threads (if two threads pass self.events.empty but only one consume get)  . 
```python
    def event_processor(self, *args):
        # `*args` received on `start`
        while self.running:
                    (event_id, func, args) = self.events.get()
                    # Wrapping into ValueAtom if arg is not an atom yet
                    with self.lock:
                          resp = self._metta.evaluate_atom(E(func,
                            *[a if isinstance(a, Atom) else ValueAtom(a) for a in args]))
                           for r in resp:
                                 self.outputs.put(r)
```

If we want that self.events.get() is gracefully (without running in daemon mode for example) terminated immediately after `self.running` is set to `False` then it is tricky. Possible solution (for proposed version of the code):
1. Use timeout in get
2. Use "sentinel" object (put termination object in queue), but if there are many threads, we should put as many sentinel objects as there are threads

Necr0x0Der:
@astroseger ,
I. Well, yeah.... There are different scenarios, though:
I.1) The agent is not a daemon, and doesn't have an event loop. Its methods are called directly, and `start` is not called. In this case, `__metta_call__` returns a stream, which is typically consumed right away, and there is no real need to create a thread for it. However, someone may still want to not consume the stream immediately.
I.2) The agent is a daemon. In this case, the stream for the method called within `__metta_call__` is not returned directly, and the method is evaluated in the background in a separate thread, which is the intended behavior. There is no `start` method, though, unless it is `EventAgent`, so no double work on stream creation is done.
I.3) The agent is an `EventAgent` and has an event loop created by `start`. In this case, it should not typically be a daemon. Initially, I created it as a daemon by default, but it instantly became obvious that there is no need for this. In this case, events are handle asynchronously, while other methods are evaluated immediately if being called directly, unless someone doesn't consume the stream they create right away.
I.4) The agent is both an `EventAgent` and a daemon. Here, a thread will be created for the event processor as well as for each `__metta_call__`.  I don't see use cases for this atm. But if there are, this behavior should be intentional.
Concluding, it may make sense to create a stream in `__metta_call__` only for daemons. But I'm not precisely sure. The current way was implemented, when there was no `EventAgent`, and wrapping all calls into stream might be redundant now. But this way is more flexible. It is up to the caller wether to wait for the stream to be finished on not. OTOH, there is indeed an overhead, which might be unnecessary (but it's not really the case that a separate thread is created for each **event**).

Necr0x0Der:
II.
> EventAgent.event_processor if self.events is empty the thread will consume 100% CPU by running

Yes, indeed.

> this code should work, but it will not exit after self.running is set to False.

Why not?...

In any case, using `get` with timeout seems a reasonable solution. Putting termination event into the queue can also work, but we need to prepend a queue if we want `stop` to work immediately, and yes, it will work only for the event loop, while other operation modes and threads will require a separate handling of `stop`, which is not that convenient. Another way is to add `sleep` if the queue is empty, but it will increase latency in reacting to all events, while timeout in get will add latency only to `stop`. So, `get` with timeout is more preferable.

Necr0x0Der:
@astroseger , do you think the last commit is the right way to address the termination issue?

astroseger:
@Necr0x0Der I have the following comments.
1. I would propose to move `StopEvent = object()` inside the class (as class attribute), to not have dangling object outside class.
2. Make sure that event_processor is running only by one thread, so forbid to run start if `self.running == True`  (or put StopEvent back to the queue to stop event_processor in other threads)
3. Remove self.lock completely, and rewrite  `clear_outputs` and `get_outputs` in the following way:
```
import queue
....
    def clear_outputs(self):
        try:
            while True:
                self.outputs.get_nowait()
        except queue.Empty:
            pass
    def get_output(self):
        try:
            while True:
                yield self.outputs.get_nowait()
        except queue.Empty:
            pass
```
otherwise in the current code there is a possibility of `deadlock` if clear_outputs and get_output can be called from different threads (code can be stuck in self.outputs.get).



Also I would like to add that multi thread codding is more trickier then it look like, so it could be more hidden problems...

Necr0x0Der:
1) yeah, I thought about it, but I supposed that events will evolve into a separate class. But I can move it there for now.
2) makes sense. Not sure if we need multiple threads for a single event agent, so I'll add exception for an attempt to run it twice
3) OK, thanks.

astroseger (approved):
From the point of view of multi threading seems fine.

PR #865, opened 2025-03-04
Add unit test for cases from issue #481

### LINKED ISSUE #481: Bug regarding superpose remaining unevaluated on empty tuples
Clearly there is no way for the following collapse to have any item:
```
!(== () (collapse (let* (($L ()) 
                         ($x (superpose $L)))
                        $x)))
```
**Yet:**
Output: [False]
Expected: [True]

MeTTa leaves the let* expression unevaluated in this case, which leaves the collapse with a "pseudo-item", an expression with variables, which leads to the comparison to return false even though there cannot be any variable assignment that actually satisfies the let* expression.

vsbogd:
Fixes #481 Issue seems to be fixed in minimal MeTTa based interpreter, this PR is to fix behavior by adding tests.

@Necr0x0Der adding you to check whether test asserts the results you meant in https://github.com/trueagi-io/hyperon-experimental/issues/481#issuecomment-1894049126

### DIFF
--- a/lib/src/metta/runner/mod.rs
+++ b/lib/src/metta/runner/mod.rs
@@ -1189,6 +1189,7 @@ fn wrap_atom_by_metta_interpreter(space: DynSpace, atom: Atom) -> Atom {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use super::bool::Bool;
 
     #[test]
     fn test_space() {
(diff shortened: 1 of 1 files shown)


Necr0x0Der (approved):
Yes, it seems so. I don't have a strong opinion, but this behavior looks most consistent

PR #890, opened 2025-03-18
Filter out results with variable loops

### LINKED ISSUE #889: Unification in match query succeeds while it should fails
**Describe the bug**
An elementary unification that should fail, succeeds in a match query.

**To Reproduce**
Run the following MeTTa code
```scheme
!(bind! &kb (new-space))
!(add-atom &kb (R $a $a))
!(match &kb (R $b (S $b)) $b)
```

**Expected behavior**
```scheme
[]
```

**Actual behavior**
```scheme
[$b]
```

**Additional context**
 - MeTTa version: 0.2.1+g0cf8adae.d20250113
 - The following `!(let (R $b (S $b)) (R $a $a) $b)` works as expected, thus the problem is with match.

vsbogd:
Fixes #889


ngeiswei (approved):
It works for me, thanks!

Issue #875, opened 2025-03-06
Conan compiles hyperonc dependencies twice on Windows platform


vsbogd:
**Describe the bug**
When CI build is processed on Windows container Conan compiles hyperonc dependencies twice.

**To Reproduce**
Run CI manual using `windows-latest` as an OS

**Expected behavior**
Conan builds dependencies only once

**Actual behavior**
Conan build dependencies twice:
```
...
2025-03-06T16:33:23.8694197Z   -------- Installing package libcheck/0.15.2 (1 of 1) --------
2025-03-06T16:33:23.8694632Z   libcheck/0.15.2: Building from source
...
2025-03-06T16:34:29.2245483Z   -------- Installing package libcheck/0.15.2 (1 of 1) --------
2025-03-06T16:34:29.2246011Z   libcheck/0.15.2: Building from source
...
```

DaddyWesker:
Firstly I thought that issue is related to two conanfile.txt files (in python and c folders). Though that is not the case. Though, I've found that this problem is also persist for python part of build. I've uploaded output of cmake build on local windows machine. It can be seen that `Computing dependency graph` step occurs four times. Two for c/conanfile.txt and two for python/conanfile.txt which is weird. I'll keep investigating conan behavior on windows platform. 

[win_build.txt](https://github.com/user-attachments/files/19308460/win_build.txt)

DaddyWesker:
@vsbogd 
I got it. It is related to:

`-- CMake-Conan: Installing both Debug and Release`

So first install builds Debug version, second install builds Release version. That is why we've got dependencies being installed twice.

vsbogd:
Build log shows that only one configuration of the hyperon libraries is built actually (Debug in case of CI scripts). Thus it is still not clear why Conan builds both types of dependencies while it should build only Debug one.

vsbogd:
For example let's look at today's nightly build https://github.com/trueagi-io/hyperon-experimental/actions/runs/13914535521/job/38936642894

Log:
```
...
2025-03-18T04:11:47.7193032Z ##[group]Run cmake --build . --config Debug
...
2025-03-18T04:11:52.4804863Z   -- CMake-Conan: Installing both Debug and Release
...

And one can see that `hyperonc` is built only once in Debug mode.

DaddyWesker:
[...]

PR #891, opened 2025-03-19
fix duplicate installation of dependencies using conan

DaddyWesker:
This should fix [issue ](https://github.com/trueagi-io/hyperon-experimental/issues/875) . Here is [manual launch](https://github.com/DaddyWesker/hyperon-experimental/actions/runs/13939848073/job/39014500121) on windows.


conan_provider.cmake
```diff
@@ -588,9 +588,8 @@ macro(conan_provide_dependency method package_name)
             message(STATUS "CMake-Conan: Installing single configuration ${CMAKE_BUILD_TYPE}")
             conan_install(${_host_profile_flags} ${_build_profile_flags} ${CONAN_INSTALL_ARGS} ${generator})
         else()
-            message(STATUS "CMake-Conan: Installing both Debug and Release")
-            conan_install(${_host_profile_flags} ${_build_profile_flags} -s build_type=Release ${CONAN_INSTALL_ARGS} ${generator})
-            conan_install(${_host_profile_flags} ${_build_profile_flags} -s build_type=Debug ${CONAN_INSTALL_ARGS} ${generator})
+            message(STATUS "CMake-Conan: Installing single configuration ${CMAKE_BUILD_TYPE}")
```
vsbogd, line 591: Did you check `CMAKE_BUILD_TYPE` has some value in case of Release multiconfiguration build? I suspect it is always empty. According to the [CMake documentation](https://cmake.org/cmake/help/latest/variable/CMAKE_CONFIGURATION_TYPES.html#variable:CMAKE_CONFIGURATION_TYPES) it is related to a single-config generators.

DaddyWesker: You can see this line in manual runs

`-- CMake-Conan: Installing single configuration Debug` 

It is messaged using `message(STATUS "CMake-Conan: Installing single configuration ${CMAKE_BUILD_TYPE}")`

CMakeLists.txt
vsbogd, line 1: Please add before line 26 comment:
```
Change in conan-provider.cmake to solve #875 depends on using ExternalProject (see https://github.com/conan-io/cmake-conan/issues/687#issuecomment-2736031979). If ExternalProject is replaced by usual CMake add_subdirectory function then conan-provider.cmake doesn't have information about current configuration because $<CONFIG> is not available at the CMake configuration stage. Possible solution in this case is using manual conan install command in CI.
```

vsbogd (approved):

PR #894, opened 2025-03-19
Allow overriding tokens from imported module

vsbogd:
Previously tokens were duplicated by modules. It happens because each module imports `corelib` and `stdlib` and module's tokenizer contains their tokens. For example we have module `A` which imports `corelib` and `stdlib` and module `B` which also imports `corelib` and `stdlib` (these imports happen automatically). Now what happens if module `A` imports module `B`. All tokens from `B` are imported into `A` and tokens of `corelib` and `stdlib` as well. Import of the tokens is done via copying tokens from imported module's tokenizer. Thus importing (`A`) module's tokenizer now contains two copies of `corelib` and `stdlib` tokens. Tokens of the imported module always added to the beginning of the importing tokenizer because we don't want to override first copy of `corelib` and `stdlib` tokens and tokenizer always search for tokens starting from the end.

First of all this PR removes duplication of the tokens. It is done by adding additional `own_tokenizer` to the module. `own_tokenizer` contains only tokens which are owned and exported by the module. For example `own_tokenizer` doesn't contain `corelib` and `stdlib` tokens. This approach doesn't work with `corelib` because `corelib` functions create tokens specific to the importing module. For example `&self` is a token which contains a link to the importing module space. Thus we need a special handling for the `corelib` tokens import. In this PR it is implemented as a conditional block in the `export_all_tokens_into` method.

`export_all_tokens_into` method replaces `import_all_tokens_from_dependency` method because last method requires importing module to know how to import tokens from exporting module. This PR replaces this logic. Now each module is responsible for exporting its tokens into importing module. This required separating `corelib` and `stdlib`. Now they are two separate modules each of it is responsible for exporting its own tokens. 

In case of Rust `stdlib` library `corelib` is an alias to the `stdlib` and both of them are imported by other modules. Duplication doesn't happens in this case because every module except `corelib` is exporting tokens from `own_tokenizer` (`stdlib`  as well). As `stdlib` is an alias and `corelib` `own_tokenizer` is empty the import of the `stdlib` doesn't add tokens to the importing module. On the other hand this will not allow us to move logic from condition inside `export_all_tokens_into` method to the module abstraction when each module ether copy tokens (default behaviour) or create tokens custom behaviour. Thus this should be fixed in future.

PR makes possible adding new tokens to the end of the tokenizer which effectively allows overriding tokens by the imported module. Also when token is registered in the module it is added into both runtime `tokenizer` and `own_tokenizer`. It is implemented for the Python, but isn't implemented for Rust modules. The reason is that in Rust tokenizer is used instead of `RunContext` to register the tokens. Taking into account this change blocks metta-lang.dev website I suggest merging it and make further changes in additional PR.

In addition some cleanup of the `corelib` (`stdlib`) is made. All tokens are separated into two groups: (1) tokens which depend on importing module's context like space, tokenizer or runner; (2) tokens which are independent of the module context. Last group of tokens could be added into `own_tokenizer` of the `corelib` but it requires additionally copying `own_tokenizer` tokens in `export_all_tokens_into` method. Also it will lead to the token duplication as `stdlib` is an alias for `corelib`. Thus it is not done yet and it is a subject to do something in the future.


vsbogd:
@luketpeterson I am not sure if you have time or will to review this. But as you are original author it would be great to have your comments about this.

luketpeterson:
I remember we talked about something like this about a year ago.  So I am happy to see you have done it. 😄

luketpeterson (approved):
I'll be honest, my head isn't in the code to the level that would be needed to spot subtle issues.  But in the big strokes, I agree with the changes you have made.

besSveta (approved):

PR #895, opened 2025-03-19
move legacy minecraft tests to metta-examples

Necr0x0Der:
`test_minecraft.py` is a nice example of functional inference with grounded operations. However, it is not a unit test per se, and checks for nothing new, which is already covered by other unit tests, and `test_minecraft_planning_with_abstractions` even has no asserts. `test_minelogy.py` has a legacy non-conventional representation as well as a representation based on the built-in type system. Nothing really unique is tested either, although it worked as a sort of integration test for some time. atm, these tests have more merits being repurposed as examples in `metta-examples`.


Necr0x0Der:
@vsbogd , just want to make sure that you are happy with this

Necr0x0Der:
https://github.com/trueagi-io/metta-examples/pull/67

vsbogd (approved):

Issue #881, opened 2025-03-07
Some operators can no longer operate on py-atoms upwards of hyperon-0.2.2


JungeWerther:
**Describe the bug**

Previously (confirmed to work in 0.2.1) you could do:

```
!(bind! now (py-atom datetime.datetime.now))
(= (A) (now))
(= (B) (now))
!(- (A) (B))
```
The `-` sign would point to the python infix operator `-`. However, in 0.2.3 for example, I get the following error:

```
thread '<unnamed>' panicked at lib/src/atom/serial.rs:65:56:
ConvertingSerializer is not expected returning error: NotSupported
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

thread '<unnamed>' panicked at library/core/src/panicking.rs:218:5:
panic in a function that cannot unwind
stack backtrace:
   0:     0x7d5e64b7535a - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::he089f96442833f67
   1:     0x7d5e64bc48d3 - core::fmt::write::h2f210ed4c94745cb
   2:     0x7d5e64b69673 - <unknown>
   3:     0x7d5e64b751a2 - <unknown>
   4:     0x7d5e64b77880 - <unknown>
   5:     0x7d5e64b77660 - std::panicking::default_hook::h24e207139139d40a
   6:     0x7d5e64b78052 - std::panicking::rust_panic_with_hook::ha9131beeb2ddc506
   7:     0x7d5e64b77dc6 - <unknown>
   8:     0x7d5e64b75859 - std::sys::backtrace::__rust_end_short_backtrace::h1d1ca3eade483f4c
   9:     0x7d5e64b77a8d - rust_begin_unwind
  10:     0x7d5e64770b2d - core::panicking::panic_nounwind_fmt::h0d5ff668f956fac4
  11:     0x7d5e64770bc2 - core::panicking::panic_nounwind::h385b7d9bda51382d
  12:     0x7d5e64770d86 - core::panicking::panic_cannot_unwind::h757b6ea37bf9b60a
  13:     0x7d5e647b08cf - metta_run
  14:     0x7d5e64795138 - <unknown>
  15:     0x7d5e6477b6de - <unknown>
  16:           0x58208f - <unknown>
  17:           0x549185 - _PyObject_MakeTpCall
  18:           0x5d73c9 - _PyEval_EvalFrameDefault
[...]

PR #899, opened 2025-03-24
Add Python operations [back] for manual import

Necr0x0Der:


Necr0x0Der:
Old code, which was in Python stdlib earlier, is returned in the form of module deployed with hyperon package, but which needs manual import. It will fix #881 (in particular), as well as some examples on metta-lang.dev. A few comments:
- it's not late to give a different name for the module if there are better suggestions
- types are removed because the main reason of adding this module is to make operations work on non-core types
- tokens for core types (numbers, strings, bools) are NOT returned. We can do this later if it proves itself to be needed, but atm I'm not sure if it is really needed

besSveta (approved):

PR #871, opened 2025-03-06
Changes to make ssl work on windows

DaddyWesker:
Subj

PR #905, opened 2025-04-03
Json dict

DaddyWesker:
So, I've added rust implementation of json dict discussed [here](https://github.com/trueagi-io/hyperon-experimental/pull/871) and in person with @Necr0x0Der . It is draft since I haven't finished writing tests and also there will be additions I'm sure.

Also, I've got metta implementation of get-all-keys and get-value-by-key functions

```
(= (get-all-keys $jsondict) (match $jsondict ($key $value) $value))

(= (get-value-by-key $jsondict $key) (match $jsondict ($key $value) $value))
```

I don't know if we need them and where to put them. Anyway, here is the usage of current functions:

```
!(bind! &jsondict (json-dict ((key1 value1) (key2 value2))))

!(assertEqual 
    (get-all-keys &jsondict) 
    (superpose (key1 key2)))

!(assertEqual 
    (get-value-by-key &jsondict key1) 
    value1)
```


Necr0x0Der:
A minor comment
`(= (get-all-keys $jsondict) (match $jsondict ($key $value) $value))`
should be
`(= (get-all-keys $jsondict) (match $jsondict ($key $value) $key))`

Necr0x0Der:
> I've got metta implementation of get-all-keys and get-value-by-key functions

It would be more interesting to have metta implementation of creating a space with the corresponding content (maybe, adding the content via `foldl-atom`) and compare performance with Rust implementation

DaddyWesker:
> > I've got metta implementation of get-all-keys and get-value-by-key functions
> 
> It would be more interesting to have metta implementation of creating a space with the corresponding content (maybe, adding the content via `foldl-atom`) and compare performance with Rust implementation

Okay, I'll test. Though I thought about using add atom function.

Necr0x0Der:
On a conceptual level, the current implementation has nothing specific about jsons. Creation of a space with certain content (underwent reduction) is generally applicable. Such the function can be created and added to stdlib (say, `new-space-reduct` or something like that). It can be done separately from json-related stuff.
The next question is if we want to use such non-specific representation for spaces as dicts (just adding pairs and treat them as key-value pairs). Or we want it to be more specific, e.g. `((Key key1) (Value value1))`. The latter will be less efficient, but it will prevent misuse of json-related functions.

Necr0x0Der:
> Though I thought about using add atom function.

Yes, folding with add-atom over the given tuple.

DaddyWesker:
It was decided to make new pr and leave only metta-implemented functions. Closing this one.

PR #907, opened 2025-04-04
Add add-reducts

DaddyWesker:
So, new pr. I've added add-reducts function after discussing with @Necr0x0Der which uses foldl-atom in it's core. Usage:

```
!(bind! &newspace (new-space))

!(add-reducts &newspace ((k1 v1) (k2 v2) (k3 v3)))

!(assertEqual
    (match &newspace (k1 $v) $v)
    v1)
```

I've also added doc and tests (@vsbogd please check if tests are in right place, I'm not sure). 

Strange thing I've noticed when I was writing tests. This one works:
```
!(chain (eval (new-space)) $newspace
      (let $f (add-reducts $newspace ((k1 v1) (k2 v2) (k3 v3)))
          (assertEqual
              (match $newspace (k1 $v) $v)
              v1)))
```

This one - not:
```
!(chain (eval (new-space)) $newspace
    (chain (eval (add-reducts $newspace ((k1 v1) (k2 v2) (k3 v3)))) $f
        (assertEqual
            (match $newspace (k1 $v) $v)
            v1)))
```

I can't quite understand why...

### DIFF
--- a/lib/src/metta/runner/stdlib/atom.rs
+++ b/lib/src/metta/runner/stdlib/atom.rs
@@ -548,7 +548,23 @@ mod tests {
         assert_eq!(run_program("!(eval (foldl-atom (1 2 3) 0 $a $b (eval (+ $a $b))))"), Ok(vec![vec![expr!({Number::Integer(6)})]]));
     }
 
+    #[test]
+    fn metta_add_reducts() {
+        assert_eq!(run_program(&format!("!(chain (eval (new-space)) $newspace (add-reducts key1))")), Ok(vec![vec![expr!("Error" ("add-reducts" "key1") "IncorrectNumberOfArguments")]]));
 
+        assert_eq!(run_program(&format!("!(chain (eval (new-space)) $newspace
+                                                 (assertEqual
+                                                     (add-reducts $newspace ((key1 value1) (key2 value2) (key3 value3)))
+                                                     ()))")),
+                   Ok(vec![vec![UNIT_ATOM]]));
+
+        assert_eq!(run_program(&format!("!(chain (eval (new-space)) $newspace
+                                                    (let $f (add-reducts $newspace ((k1 v1) (k2 v2) (k3 v3)))
+                                                        (assertEqual
+                                                            (match $newspace (k1 $v) $v)
+                                                            v1)))")),
+                   Ok(vec![vec![UNIT_ATOM]]));
+    }
 
     #[test]
     fn size_atom_op() {
--- a/lib/src/metta/runner/stdlib/atom.rs
+++ b/lib/src/metta/runner/stdlib/atom.rs
@@ -550,11 +550,11 @@ mod tests {
 
     #[test]
     fn metta_add_reducts() {
-        assert_eq!(run_program(&format!("!(chain (eval (new-space)) $newspace (add-reducts key1))")), Ok(vec![vec![expr!("Error" ("add-reducts" "key1") "IncorrectNumberOfArguments")]]));
+        assert_eq!(run_program(&format!("!(chain (eval (new-space)) $newspace (add-reducts k1))")), Ok(vec![vec![expr!("Error" ("add-reducts" "k1") "IncorrectNumberOfArguments")]]));
 
         assert_eq!(run_program(&format!("!(chain (eval (new-space)) $newspace
                                                  (assertEqual
-                                                     (add-reducts $newspace ((key1 value1) (key2 value2) (key3 value3)))
+                                                     (add-reducts $newspace ((k1 v1) (k2 v2) (k3 v3)))
                                                      ()))")),
                    Ok(vec![vec![UNIT_ATOM]]));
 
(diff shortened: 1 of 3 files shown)


DaddyWesker:
So, @Necr0x0Der  asked me to add test to check if add-reducts is really reducing input values inside tuple and add-atoms (which I forgot to add) - not. I've tried this one:

```
(= (pair4) (k4 v4))
(= (pair5) (k5 v5))
(= (pair6) (k6 v6))

!(bind! &reductspace (new-space))
!(add-reducts &reductspace ((pair4) (pair5) (pair6)))
!(get-atoms &reductspace)
```

And got some strange results here. So, in this example output is
`[(k4 v4), (k4 v4), (k5 v5), (k5 v5), (k6 v6), (k6 v6)]`

Also, I've got the same result with add-atoms:
(: add-atoms (-> Grounded Expression (->)))
(= (add-atoms $space $tuple)
    (foldl-atom $tuple () $a $b (eval (add-atom $space $b))))

...
!(add-atoms &reductspace ((pair4) (pair5) (pair6)))
!(get-atoms &reductspace)
;[(k4 v4), (k4 v4), (k5 v5), (k5 v5), (k6 v6), (k6 v6)]
```

So, changes in type of function does not change result of function in this particular case - input is still being reduced and for some reason every pair is added twice into space.

As an experiment I've tried naive implementation using recursion:
```
(= (dict-reduct $dictspace $tuple)
    (let $car (car-atom $tuple)
        (let $cdr (cdr-atom $tuple)
            (let $unit (add-atom $dictspace $car)
                (dict-reduct $dictspace $cdr)))))
...
!(dict-reduct &reductspace ((pair4) (pair5) (pair6)))
!(get-atoms &reductspace)
; [(k4 v4), (k5 v5), (k6 v6)]
```

and it worked fine. Also, I've tried rust implementation from previous closed PR (which was called json-dict, though it is not right name).

Rust:
```
#[derive(Clone, Debug)]
pub struct JsonDictOp {}

grounded_op!(JsonDictOp, "json-dict");

impl Grounded for JsonDictOp {
    fn type_(&self) -> Atom {
        Atom::expr([ARROW_SYMBOL, ATOM_TYPE_UNDEFINED, rust_type_atom::<DynSpace>()])
    }

    fn as_execute(&self) -> Option<&dyn CustomExecute> {
        Some(self)
    }
}

impl CustomExecute for JsonDictOp {
        let arg_error = || ExecError::from("json-dict expects expression as an argument");
        let expr = TryInto::<&ExpressionAtom>::try_into(args.get(0).ok_or_else(arg_error)?)?;

        let atoms = expr.children();
        let mut json_space = GroundingSpace::new();
        for atom in atoms.iter() { json_space.add(atom.clone()); }
        Ok(vec![Atom::gnd(DynSpace::new(json_space))])
    }
}
```

Metta:
```
...
!(bind! &reductspace (json-dict ((pair4) (pair5) (pair6))))
!(get-atoms &reductspace)
;[(k4 v4), (k5 v5), (k6 v6)]
```

Rust implementation not duplicating atoms also. 

So, I'm not sure where exactly problem is. Maybe in foldl, maybe in a way I'm using foldl. I'll try to investigate.

File: lib/src/metta/runner/stdlib/stdlib.metta [hunk 1]
```diff
@@ -667,6 +667,26 @@
    (let $c1 (collapse $arg1) (let $c2 (collapse $arg2)
      (let $u (subtraction-atom $c1 $c2) (superpose $u)))))
 
+(@doc add-reducts
+  (@desc "Function takes space and expression which contains key-value pairs, reduces them and adds result into given space")
```
Necr0x0Der, line 671: Expression should not necessarily contain key-value pairs. In case of this function, these can be arbitrary atoms

lib/src/metta/runner/stdlib/stdlib.metta
```diff
@@ -667,6 +667,26 @@
...
+  (@params (
+    (@param "Space")
+    (@param "Key-value pairs in form of expression")))
+  (@return "Unit atom"))
+(: add-reducts (-> Grounded %Undefined% (->)))
+(= (add-reducts $space $tuple)
+    (foldl-atom $tuple () $a $b (eval (add-atom $space $b))))
+
+(@doc add-atoms
+  (@desc "Function takes space and expression which contains key-value pairs and adds them into given space. No reducing")
```
Necr0x0Der, line 681: key-values are also irrelevant here

vsbogd:
> This one - not:
> 
> ```
> !(chain (eval (new-space)) $newspace
>     (chain (eval (add-reducts $newspace ((k1 v1) (k2 v2) (k3 v3)))) $f
>         (assertEqual
>             (match $newspace (k1 $v) $v)
>             v1)))
> ```
> 
> I can't quite understand why...

The issue with this is that `eval` does the only step of a function evaluation. You sometimes cannot see the difference in MeTTa interpreter because `eval` results are evaluated further. For example the following code:
```
(= (foo) (bar))
(= (bar) Ok)

!(chain (eval (foo)) $x $x) ; returns Ok
```
But if you add quote around `$x` you will see the difference:
```
!(chain (eval (foo)) $x (quote $x)) ; returns (quote (bar))
```
It is not a case with `let`, because second argument is fully evaluated before body is evaluated:
```
!(let $x (foo) $x) ; returns Ok
!(let $x (foo) (quote $x)) ; returns (quote Ok)
```

In the example above `(eval (add-reducts $newspace ((k1 v1) (k2 v2) (k3 v3))))` returns `(foldl-atom ...)` and atom is not actually added into the space. This explains the result of the test.

lib/src/metta/runner/stdlib/atom.rs
vsbogd, line 1: I would add tests into `stdlib/mod.rs` file because the implementation is located in `stdlib.metta` but old `foldl-atom` tests are kept here thus lets leave it as is.

lib/src/metta/runner/stdlib/atom.rs
```diff
@@ -548,7 +548,61 @@ mod tests {
         assert_eq!(run_program("!(eval (foldl-atom (1 2 3) 0 $a $b (eval (+ $a $b))))"), Ok(vec![vec![expr!({Number::Integer(6)})]]));
     }
 
+    #[test]
+    fn metta_add_reducts() {
+        assert_eq!(run_program(&format!("!(chain (eval (new-space)) $newspace (add-reducts k1))")), Ok(vec![vec![expr!("Error" ("add-reducts" "k1") "IncorrectNumberOfArguments")]]));
```
vsbogd, line 553: Taking into account https://github.com/trueagi-io/hyperon-experimental/pull/907#issuecomment-2783637378 I am not sure if `chain/eval` is a good pattern here. On the one hand it is safe because `(new-space)` is a grounded call and thus `eval` evaluates it in one step. On the other hand it is not generally evaluates `eval`'s argument till the end thus it is not possible using it everywhere.

DaddyWesker: So, your suggestion is to replace chain-eval with let here just in case?

lib/src/metta/runner/stdlib/stdlib.metta
DaddyWesker: @Necr0x0Der  Okay, so just `Function takes space and expression, evaluates atoms in it and adds them into given space` then?

DaddyWesker:
@Necr0x0Der @vsbogd  Done.

vsbogd (approved):

Necr0x0Der (approved):

PR #908, opened 2025-04-07
Add assert and assert contains functions

### LINKED ISSUE #900: Add more assert function types
## Implement assert

We have `assertEqual` and so, but we don't have `assert`.  It is a trivial addition, but my point is that it should probably be part of the stdlib.

To be clear what I mean

```scheme
!(assert (or True False))
```

would output

```
[()]
```

While

```scheme
!(assert (and True False))
```

would raise an error.

## Implement assertContain

Another very convenient one would be `assertContain` (and maybe `assertAlphaContain` as well) which would check if the expected result is contained in a superposition of results.

For instance

```scheme
(= (foo) 41)
(= (foo) 42)
(= (foo) 43)

;; Make sure that 42 is part of the solution set of foo
!(assertContain (foo) 42)
```

Bitseat:
Fixes #900

### DIFF
--- a/lib/src/metta/runner/stdlib/stdlib.metta
+++ b/lib/src/metta/runner/stdlib/stdlib.metta
@@ -679,6 +679,22 @@
             ()
             (Error $arg (Expected True, evaluated to  $atom)))))
 
+(@doc assertContains
+  (@desc "Checks if the second argument is one of the results of the first argument's evaluation, rather than requiring all the results")
+  (@params (
+    (@param "First expression")
+    (@param "Second expression")))
+  (@return "Unit atom if the second argument is contained in the results of the first argument's evaluation, error - otherwise"))
+(: assertContains (-> Expression Expression (->)))
+(= (assertContains $atom $content)
+    (let $diff (collapse (subtraction (superpose ($content)) $atom))
+        (if (== $diff ())
+            ()
+            (Error $diff ("assertContains error:" $diff "not contained in" $atom))
+        )
+    )
+)
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ; Documentation formatting functions
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
--- a/lib/src/metta/runner/stdlib/stdlib.metta
+++ b/lib/src/metta/runner/stdlib/stdlib.metta
@@ -667,6 +667,18 @@
    (let $c1 (collapse $arg1) (let $c2 (collapse $arg2)
      (let $u (subtraction-atom $c1 $c2) (superpose $u)))))
 
+(@doc assert
+  (@desc "Asserts that the result of evaluating an expression is True")
+  (@params (
+    (@param "Expression to evaluate")))
+  (@return "Unit atom if the expression evaluates to True, error - otherwise"))
+(: assert (-> Expression Atom))
+(= (assert $arg)
+   (let $atom (eval $arg)
+        (if (== $atom True)
+            ()
+            (Error $arg (Expected True, evaluated to  $atom)))))
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ; Documentation formatting functions
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
--- a/python/tests/scripts/a1_symbols.metta
+++ b/python/tests/scripts/a1_symbols.metta
@@ -20,6 +20,20 @@
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; `assertContains` Checks if the second argument is 
+; one of the expected results of the first   
+; expression's evaluation
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+!(assertContains
+  (match &self (($x leaf2) leaf3) $x)
+  leaf1)
+
+!(assertContains
+  (match &self (($x leaf2) leaf3) $x)
+  (leaf0 leaf1))
+  
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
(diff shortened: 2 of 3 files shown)


Bitseat:
Help - Calling assertContains function in the test file (a1_symbols.metta) is not working.

lib/src/metta/runner/stdlib/stdlib.metta
```diff
@@ -667,6 +667,34 @@
    (let $c1 (collapse $arg1) (let $c2 (collapse $arg2)
      (let $u (subtraction-atom $c1 $c2) (superpose $u)))))
 
+(@doc assert
+  (@desc "Asserts that the result of evaluating an expression is True")
+  (@params (
+    (@param "Expression to evaluate")))
+  (@return "Unit atom if the expression evaluates to True, error - otherwise"))
+(: assert (-> Expression Atom))
```
ngeiswei, line 675: Note that with that type signature, the following
```scheme
!(assert True)
```
might fail, because `True` is not an expression (an expression is basically a tuple), thus you probably want
```scheme
(: assert (-> Atom Atom))
```
or maybe
```scheme
(: assert (-> Atom (->))
```
where `(->)` seems to denote the unit type (I'm not completely sure about that though, maybe @vsbogd could shed some light).

lib/src/metta/runner/stdlib/stdlib.metta
```diff
@@ -667,6 +667,34 @@
...
+(= (assert $arg)
+   (let $atom (eval $arg)
+        (if (== $atom True)
+            ()
+            (Error $arg (Expected True, evaluated to  $atom)))))
```
ngeiswei, line 680: I've never hacked the stdlib so I'm not sure at all, but I wonder, is `eval` really needed?

Also, what would
```scheme
!(assert (empty))
```
do?  Would it raise an error, or silently output `[]`?  I believe it should raise an error, but I'm not sure your code behaves like that.  You should add a unit test for it.  Which is another question (to @vsbogd): where should the unit tests of the stdlib be?

lib/src/metta/runner/stdlib/stdlib.metta
```diff
@@ -667,6 +667,34 @@
...
+
+(@doc assertContains
+  (@desc "Checks if the second argument is one of the results of the first argument's evaluation, rather than requiring all the results")
+  (@params (
+    (@param "First expression")
+    (@param "Second expression")))
+  (@return "Unit atom if the second argument is contained in the results of the first argument's evaluation, error - otherwise"))
+(: assertContains (-> Expression Expression (->)))
```
ngeiswei, line 688: Same comment about `Expression`, it should probably be `Atom`.

python/tests/scripts/a1_symbols.metta
```diff
@@ -18,6 +18,20 @@
...
 
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; `assertContains` Checks if the second argument is 
+; one of the expected results of the first   
+; expression's evaluation
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+!(assertContains
+  (match &self (($x leaf2) leaf3) $x)
+  leaf1)
```
ngeiswei, line 29: I don't think the tests should be here, they should probably be in their own place, but I don't know where.  @vsbogd, what do you think?

ngeiswei (changes requested):
Beside a few things, see my comments, I think it looks good.  Could you implement as well `assertIncludes`, as you suggested?

ngeiswei:
> Help - Calling assertContains function in the test file (a1_symbols.metta) is not working.

I believe it is because of the type signature.  If you replace `Expression` by `Atom` it might get fixed.

python/tests/scripts/b1_equal_chain.metta
```diff
@@ -71,3 +71,11 @@
...
    (eq Green Blue)
   ((eq Green Blue)))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; `assert` checks if the result of evaluating an
+;  expression is True
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+!(assert (or True False))
+!(assert (< 5 8))
```
ngeiswei, line 81: Very minor, but the POSIX standard recommends to end text files with an empty new line.  That's why Github is showing this red forbidden symbol.

python/tests/scripts/a1_symbols.metta
vsbogd: Usually I write unit tests for the stdlib in Rust. The Rust MeTTa runner is called with MeTTa program written as a string and result is compared to the expected atom. The only tricky part there is using `expr!` macros to create an expected atom. The macros has the following syntax:
- string in quotes means symbol
- parenthesis add nested expression
- identifier (without quotes) creates variable
- curly brackets add grounded atom

For instance: `expr!("sym" var ("nested" "expression"))` corresponds to the MeTTa `(sym $var (nested expression))`.

For example see link below or any other test which calls `run_program`: https://github.com/trueagi-io/hyperon-experimental/blob/bde1b596f4e63a26faf377902cda00915908d27e/lib/src/metta/runner/stdlib/mod.rs#L675-L703

Necr0x0Der: I definitely agree that tests for different asserts should not be in `a1_symbols.metta`. If one really wants these tests to be in `scripts` folder, one should possibly create a separate file and systematically cover different asserts there, so this file would simultaneously serve as a sort of tutorial about asserts and provide a good coverage on them. A stand-alone test for a particular assert should be somewhere else, e.g. in Rust tests for stdlib as @vsbogd said.

Bitseat:
Thanks for all the feedback. Fixed all the issues with assert and assertContains. assertIncludes is still a work in progress and I will explore how to include assertEqualMsg with this PR. I am also writing tests for these functions and I was wondering how to handle the non deterministic behavior of metta in the case of assert?
for example, given
(= (foo) 1)
(= (foo) 2)

should !(assert (== (foo) 1)) return unit atom?

vsbogd:
@Bitseat I see commits merged from the main branch. I would suggest re-create the branch from scratch and add your code in it and then raise new PR. Otherwise it is difficult to review it.

Bitseat:
@vsbogd , I agree. Closing this one and creating a new PR.

PR #909, opened 2025-04-07
Limit recursion depth

vsbogd:
Return error if stack depth limit is reached.

Current implementation has the following pecularities:

1. The stack depth is calculated in terms of the minimal MeTTa instructions not MeTTa instructions thus the actual stack depth can be bigger than expected as minimal MeTTa instructions are more detailed.

2. Execution is interrupted and returns error only on the next MeTTa function call. This is required to correctly return error from the
minimal MeTTa stack when MeTTa program is executed.

3. Case/capture/superpose/collapse/assert operations call interpreter from the interpreter stack and it leads to the creation of the new stack. Thus if `case` (or other similar operation) is used the limit counter is started from the beginning for the nested expression. In stack where `case` calls `case` the limit is not reached.

### DIFF
--- a/lib/src/atom/mod.rs
+++ b/lib/src/atom/mod.rs
@@ -784,7 +784,7 @@ impl Clone for Box<dyn GroundedAtom> {
 
 /// Atoms are main components of the atomspace. There are four meta-types of
 /// atoms: symbol, expression, variable and grounded.
-#[derive(Debug, Clone)]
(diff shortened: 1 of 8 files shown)


vsbogd:
I don't see an easy way to fix the use-case when `case` is used in recursive function. I am going to fix it in a further PR.

Necr0x0Der (approved):

PR #912, opened 2025-04-10
write/read file added

DaddyWesker:
So, I've added function file-read and file-write which allows to read/write file. Usage:

```
!(file-read "path/to/file.txt")
!(file-write "path/where/to/write/file.txt" "content in form of string")
```

I've also added docs and tests.
@vsbogd  please check if everything is right.
@Necr0x0Der  please check if this is what you've asked me to do.


Necr0x0Der:
Yes, it looks like what we need to read/write json files. I wonder, though, if these functions:
1) should handle only strings, or work with binary data as well
2) should be atomic functions or should we have `open` (with r/w, t/b options)
In the latter case, I also wonder if unused atoms are automatically deleted and the file can be closed in a construction like:
```
(let $f (fopen "rt" "my.txt")
   (fwrite $f "something))
```
@vsbogd ?

vsbogd:
> 2\. should be atomic functions or should we have `open` (with r/w, t/b options)
>     In the latter case, I also wonder if unused atoms are automatically deleted and the file can be closed in a construction like:
> 
> ```
> (let $f (fopen "rt" "my.txt")
>    (fwrite $f "something))
> ```

Reading the content of the file into one big string can be difficult when file is really big. Thus I would vote for open/close/read/write and we can build read all into a string on top of these primitives.

I am not sure how easy is to implement `(let $f (fopen ...) ...)` one can wrap file handle into smart pointer and then put this pointer into a grounded atom. Thus when all atoms which hold the link to the handle are destroyed handle is destroyed as well. I would suggest try it and check how does it work.

vsbogd:
General comment: I would suggest rework it and create a built-in module using `skel.rs` in #913 as an example.

Necr0x0Der:
> General comment: I would suggest rework it and create a built-in module using `skel.rs` in #913 as an example.

Yes, this was our intention.

DaddyWesker:
@Necr0x0Der @vsbogd 
I've done refactoring to skel-like module. Please check if everything is fine now. For now, I'll write some tests.

lib/src/metta/runner/builtin_mods/fileio.rs
```diff
@@ -0,0 +1,190 @@
...
+            Ok(file) => Ok(Self(Rc::new(RefCell::new(file)))),
+            Err(_) => Err(ExecError::from("Failed to open file with provided arguments"))
+        }
+    }
+
+    fn read_as_str(&self) -> String
+    {
+        self.0.borrow_mut().seek(SeekFrom::Start(0)).unwrap();
+        let mut contents = String::new();
+        let _ = self.0.borrow_mut().read_to_string(&mut contents);
```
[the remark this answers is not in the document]
DaddyWesker: As i know, write and read can provide errors only in case when provided path to filename doesn't exist. Though in our case it can't happen since we are using OpenOptions and if file doesn't exist and create option is false it will fail on OpenOptions.open() step. Or maybe you mean something else?

DaddyWesker:
@vsbogd  I've fixed what you've asked, added doc (previous one was a stub from first version and I forgot to change it) and wrote a test.

lib/src/metta/runner/builtin_mods/fileio.rs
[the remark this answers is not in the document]
vsbogd: > As i know, write and read can provide errors only in case when provided path to filename doesn't exist.

There are many possibilities for the errors during reading/writing. For example space on disk can be exhausted during write operation. Detachable storage can be detached during reading.

lib/src/metta/runner/builtin_mods/fileio.rs
```diff
@@ -0,0 +1,156 @@
...
+pub const ATOM_TYPE_FILE_HANDLE: Atom = sym!("FileHandle");
+
+#[derive(Clone, Debug)]
+pub struct FileHandle(Rc<RefCell<fs::File>>);
+
+impl FileHandle {
+    fn open(path: &str, options: &str) -> Result<Self, ExecError> {
+
+        let mut opened_file = OpenOptions::new();
+        let opened_file= opened_file.read(options.contains("r")).write(options.contains("w")).create(options.contains("c")).append(options.contains("a")).truncate(!options.contains("a"));
```
vsbogd, line 26: It is more readable in this way:
```suggestion
        let opened_file = opened_file.read(options.contains("r"))
            .write(options.contains("w"))
            .create(options.contains("c"))
            .append(options.contains("a"))
            .truncate(!options.contains("a"));
```

lib/src/metta/runner/builtin_mods/fileio.metta
```diff
@@ -0,0 +1,19 @@
+(@doc file-open
```
vsbogd, line 1: Functions with side-effects should have `!` exclamation mark at the end of the name. Please add it to `file-open!`, `file-read!` and `file-write!`

lib/src/metta/runner/builtin_mods/fileio.rs
```diff
@@ -0,0 +1,156 @@
...
+#[cfg(test)]
+mod tests {
+    use crate::metta::*;
+    use crate::metta::runner::run_program;
+
+    #[test]
+    fn test_filehandle() {
+        let program = "
+            !(import! &self fileio)
+            !(bind! &fhandle (file-open \"test.txt\" \"rwc\"))
```
vsbogd, line 144: I would recommend using temporary file name instead. You can get it using this library https://docs.rs/tempfile/latest/tempfile/struct.NamedTempFile.html

vsbogd (changes requested):

DaddyWesker:
@vsbogd  Done

lib/src/metta/runner/builtin_mods/fileio.rs
```diff
@@ -0,0 +1,168 @@
...
+            Ok(_) => Ok(contents),
+            Err(_) => Err(ExecError::from("Failed to read file contents"))
+        }
+    }
+
+    fn write(&self, content: &str) -> Result<(), ExecError>
+    {
+        match self.0.borrow_mut().write(&content.as_bytes()) {
+            Ok(_) => Ok(()),
+            Err(_) => Err(ExecError::from("Failed to read file contents"))
```
vsbogd, line 54: It is a write function:
```suggestion
            Err(_) => Err(ExecError::from("Failed to write file contents"))
```
Also it is worth to add the reason of the error into the error message (for both read and write). Because user should understand what caused the issue to prevent it next time.

DaddyWesker: > Also it is worth to add the reason of the error into the error message (for both read and write). Because user should understand what caused the issue to prevent it next time.

You mean something like that?
`Err(message) => Err(ExecError::from(format!("Failed to write file contents: {}", message)))`

vsbogd: Yeah

DaddyWesker:
@vsbogd  Done.

vsbogd (approved):

vsbogd:
This API doesn't have functions to read part of the file or for binary reading. Am I right it is planned to be added by additional PRs?

DaddyWesker:
@vsbogd 
Question is - what will be better? To add changes to this pr or make another?

vsbogd:
I would merge this PR and make next one.

PR #917, opened 2025-04-23
Allow direct tokens loading into importing module

vsbogd:
This PR is a followup after #894. It adds `ModuleLoader::load_tokens` method and uses it to export tokens from module A into importing module B. Thus only exported tokens are copied into the importing module. This change is simple on a Rust side but required significant work on Python side as fully functional `ModuleLoader` implementation is required there.

On core library side `ModuleLoader::load_tokens` method is added and implemented for `corelib` and built-in modules. It is used to export tokens into importing module and old `own_tokenizer` is removed.

On `hyperonc` side new module `module.rs` is added. Most of the module handing code and structure are moved there. New `module_loader_t` is implemented which allows implementing `ModuleLoader` interface from a C program. `fs_module_format_t` structure is modified to reflect Rust `FsModuleFormat` API. Old `ModuleLoader` implementations are removed and replaced by the one.

On `hyperonpy` side the `module_loader_t` API is used to implement loading `stdlib` and direct loading other Python modules by name or via function call. `fs_module_format_t` API is used to implement loading Python modules by name using `import!` from MeTTa programs. `MettaModRef` is added to allow `load_tokens` implementation getting tokenizer from importing MeTTa module. 

The rework of the Python module importing code is done. Now functions which register MeTTa tokens are not wrapped but instead marked by `metta_type` attribute which also points how the returned atoms should be registered. Also `metta_pass_metta` attribute is added to know whether the function expects metta as a parameter or not. Then module loading function `_priv_register_module_tokens` goes through all marked functions and calls them passing results to the tokenizer.

### DIFF
--- a/c/src/lib.rs
+++ b/c/src/lib.rs
@@ -5,3 +5,4 @@ pub mod atom;
 pub mod space;
 pub mod metta;
 pub mod serial;
+pub mod module;
--- a/c/src/metta.rs
+++ b/c/src/metta.rs
@@ -858,29 +858,6 @@ pub extern "C" fn metta_new_with_space_environment_and_stdlib(space_ref: *mut sp
     metta.into()
 }
 
-/// Internal wrapper to turn a mod_loader_callback_t into an trait object that implements [ModuleLoader]
-#[derive(Debug)]
-struct CModLoaderWrapper {
-    //NOTE: This function type matches the internals of mod_loader_callback_t.  This is necessary because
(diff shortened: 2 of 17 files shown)


besSveta (approved):

Necr0x0Der (approved):
@besSveta , have you checked it with metta-motto?

vsbogd:
> @besSveta , have you checked it with metta-motto?

Sveta wrote me that she checked already, so I am merging it.

Issue #177, opened 2022-08-22
In-place type definition/restriction


Necr0x0Der:
There are different cases, when it is desirable to indicate the type or meta-type of a symbol (especially, variable) in a concrete expression.
1) Some examples come from dependent type usage, e.g. Idris definition `Refl : (x : a) -> Equal x x` supposes that we have a variable of type `a`, and the output type is defined by this variable. `Refl 42` is of type `Equal 42 42`. In MeTTa, we specify only types. For `(: Refl (-> a (Equal a a)))` , `(Refl 42)` will be of type `(Equal Int Int)`. **In type definitions, we are lacking the possibility to use the argument value passed to the constructor**. There is a possible workaround, which partly works, if we define: `(: (Refl $x) (=== $x $x))`. `!(get-type (Refl 42))` returns `[(%Undefined% Number), (=== 42 42)]`. The problem is also that the type of `Refl` itself is `%Undefined%`. Whether we should allow (and interpret in a special way) in-place definitions like `(: $x $a)` or should support definitions like `(: (Refl $x) (=== $x $x))` more cleanly, is the question to be discussed.

2) **In equalities, we are lacking the possibility to specify the type of the argument (or bind it to a variable)**.  For example, one may want to return the type of the argument, e.g. `(= (type (: $x $t)) $t)`. It should be noted that we may need to specify a type (e.g., `(: $x $t)` instead of just `$x`), which is normally missing, while in type definitions we may need to specify the argument (e.g., `(: $x $a)` instead of just `$a`), which is normally missing. The example of using the type of the argument can be worked around with `get-type` grounded function: `(= (type $x) (get-type $x))`. 

2b) We may want to use types of arguments as additional restrictions / conditions. For example, if we have a generic fmap type definition
`(: fmap (-> (-> $a $b) ($F $a) ($F $b)))`
it's tempting to define
`(= (fmap $f $C0) $C0)`
to handle `(: Nothing (Maybe $t))` or `(: Nil (List $t))`. Unfortunately, the pattern `(fmap $f $C0)` will be matched against any expression to be bound to `$C0`. We may want to restrict `$C0` meta-type to be `Symbol`. Something like
```
(= (fmap $f $C0)
   (if (== (get-meta-type $C0) Symbol)
       $C0
       (superpose ())))
```
may work, but looks cumbersome. It would be much more convenient to be able to write `(= (fmap $f (: $C0 Symbol)) $C0)`.
[...]

PR #918, opened 2025-04-24
Stop evaluating atom if it is returned by the function which return type is Atom

### LINKED ISSUE #914: Convenient way to stop evaluations
**Describe the issue**
There are many cases, especially, in meta-programming, when we want to prevent evaluation of expressions. Currently, with Minimal MeTTa instructions, it seems to be possible in principle, but rather inconvenient.

**To Reproduce**
Let's say we want to construct an expression `(+ 1 2)` without evaluating it with the use of `cons-atom`: `(cons-atom + (1 2))`. Apparently, `(cons-atom + (1 2))` will return `(+ 1 2)`, which will be immediately reduced to `3`. Even we would be ok with wrapping the result into `quote`, it would still be problematic. Obviously, `(quote (cons-atom + (1 2)))` will reduce nothing at all. One could try
```
!(let $r (cons-atom + (1 2))
   (quote $r))
```
But the result of `(cons-atom + (1 2))` will be reduced to `3` before getting binded to `$r`.

One workaround is
```
!(chain (cons-atom + (1 2)) $q (quote $q))
```
which requires `quote`, but at least works with it. However, it's annoying to pass `quote` everywhere. Also, this construction will not interpret subexpressions. For example, one could try to write:
```
(= (one-step (quote $expr))
   (chain $expr $q (quote $q)))
```
or
```
(= (one-step (quote $expr))
   (chain (eval $expr) $q (quote $q)))
```
`(one-step (quote (cons-atom + (1 2))))` will produce `(quote (+ 1 2))` in both cases. And `(one-step (one-step (quote (cons-atom + (1 2)))))` will produce `3` in the second case (but will remain `(quote (+ 1 2))` in the first case).

`(one-step (quote (cons-atom (+ 1 2) (+ 3 4))))` will  produce `(quote ((+ 1 2) + 3 4))` in both cases, but adding one more `one-step` will not change the result in the first case and produce `(quote NotReducible)` in the second case. Maybe, there are more complicated ways to handle this. However, it would be nice to have simpler ways.

As it was suggested earlier, if the return type of any function is `Atom`, it's result should not be evaluated further.

vsbogd:
Fixes #914. Tested on the following examples:
```
(= (bar) X)
(= (foo) (bar))

(: stop (-> Atom))
(= (stop) (foo))

!(stop) ; returns (foo)

!(cons-atom + (1 2)) ; returns (+ 1 2)
```

Evaluation of the expression stops if function return type is `Atom`. First I added `%Undefined%` return type to the all of atoms of the standard library here https://github.com/trueagi-io/hyperon-experimental/pull/918/commits/b98e3dc4ec50d6e72c116d700152eadc8245e985. Then I used `Atom` for some functions here https://github.com/trueagi-io/hyperon-experimental/pull/918/commits/1478e8b915ad2c0d6276e13776ff61d298bc7e7d and here https://github.com/trueagi-io/hyperon-experimental/pull/918/commits/edee1e5a1ea9a4f5be3fc2d0ce8902748cf2265d


ngeiswei (approved):
I don't completely understand the fix but it works for me, thanks a lot!

vsbogd:
Just in case fix is in this commit https://github.com/trueagi-io/hyperon-experimental/pull/918/commits/99dd9911f85cbbd76acbc8390a54d35e2210d42c
I check the return type of the function and send flat to the `metta_call` to stop evaluate further.

Necr0x0Der:
I'm definitely happy with this change, especially if it (almost) doesn't affect the existing tests. I've got a few minor questions, though:
* JIC, is there a test, which checks that `Atom`-typed result will not be further evaluated.  I didn't notice it.
* I wonder why `=` should have `Atom` as the result type (I'm referring to the change in d4_type_prop.metta). I don't think we frequently execute `=` as a function. If `Atom` doesn't work for it, I'd like to know why. If it really doesn't work for a reason, I still don't like `%Undefined%` here too much. And, maybe, the result of `=` should have a dedicated type (although this can be addressed in a separate issue/PR).
* I wonder if we can redefine `quote` from
```
(: quote (-> Atom Atom))
(= (quote $atom) NotReducible)
```
to
```
(: quote (-> Atom Atom))
(= (quote $atom) $atom)
```
as it was originally envisioned. If we need a strong quotation constructor, we can have both quote functions (maybe we can keep `quote` as is and introduce a new one under a different name). This can be discussed separately, but I just want to mention this.
* If we want to be more specific about return types, we need to be able to define input arguments as both specific types or varialbe types and `Atom` meta-type simultaneously, say, `(: f (-> (: $t Atom) $t))` or something like this instead of `(: f (-> Atom %Undefined%))`. But this is just a side remark. It should definitely not influence this PR. Just `%Undefined%` return type makes it more obvious

vsbogd:
> * I wonder why `=` should have `Atom` as the result type (I'm referring to the change in d4_type_prop.metta). I don't think we frequently execute `=` as a function. If `Atom` doesn't work for it, I'd like to know why. If it really doesn't work for a reason, I still don't like `%Undefined%` here too much. And, maybe, the result of `=` should have a dedicated type (although this can be addressed in a separate issue/PR).

Making `=` return `Atom` fails this test: https://github.com/trueagi-io/hyperon-experimental/blob/58891e8131de3bc5328ffa185556dc4c1375d574/python/tests/scripts/d4_type_prop.metta#L112-L114
It can be fixed by adding additional `(: = (-> $t $t %Undefined%))` type for the `=` in the test though.

vsbogd:
> * I wonder if we can redefine `quote`

It is not safe to redefine exactly `quote` because it significantly affects the already written code (even part of the interpreter):
1. Some code matches `quote` to unquote it
2. In the following case the effect is different:
```
(= (bar) X)
(= (foo) (bar))
(: q (-> Atom Atom))
(= (q $a) $a)
(= (e $a) $a)

!(e (q (foo)))  ; returns X
!(e (quote (foo))) ; returns (quote (foo))
```

vsbogd:
> * JIC, is there a test, which checks that `Atom`-typed result will not be further evaluated.  I didn't notice it.

I am sorry, forgot to add it. Added in https://github.com/trueagi-io/hyperon-experimental/pull/918/commits/191407634b49fd034235d7b4e2b63e4545057fb5

Necr0x0Der:
> Making `=` return `Atom` fails this test:
> 
> https://github.com/trueagi-io/hyperon-experimental/blob/58891e8131de3bc5328ffa185556dc4c1375d574/python/tests/scripts/d4_type_prop.metta#L112-L114
> 
> 
> It can be fixed by adding additional `(: = (-> $t $t %Undefined%))` type for the `=` in the test though.

So, this test really executed `=`. And it supposes that execution of`=` returns `T` of `Type` type. So, it's reasonable that adding this behavior requires adding a type definition as well. I'm not sure what behavior will be, when there are two type definitions - one with `Atom` result and another with ordinarily typed result. If the test works with such addition, then maybe it's a way to go. But maybe we should consider this more carefully separately. I'm not voting for `Atom` or `Expression` as a return type. But I also don't like `%Undefined%` there. I'm leaning towards a specific type like `Type`, `Prop`, `Equality` or whatever else. OK, let's discuss this in a separate issue.

Necr0x0Der:
> I agree, am I right that #177 is the same issue?

Yes. I just wanted to point out that when the return type becomes `%Undefined%`, which is not a meta-type, it becomes even more tempting to have a parameterized return type there. So, it's not a new issue. Just a remark.

Necr0x0Der:
> ```
> !(e (q (foo)))  ; returns X
> !(e (quote (foo))) ; returns (quote (foo))
> ```

Well, the very reason we used `quote` this way was that it was not possible to make it work as `q` in your example works now. If we can redefine both `quote` and `unquote`, e.g. `unquote` is just ordinary `id` with type `(: unquote (-> $t $t))`, then it could be preferable and more efficient. But I agree that this change should be considered carefully, so it would be better to consider this as a separate issue.

Necr0x0Der (approved):

PR #931, opened 2025-05-14
random to module

DaddyWesker:
I moved random from stdlib to built-in module as we've discussed with @vsbogd . Currently, two commented-out tests not working since 
`!(assertEqual (random-int &rng 5 0) (Error (random-int &rng 5 0) RangeIsEmpty))`

now gives
`[(Error (assertEqual (random-int RandomGenerator-0x62962c56f410 5 0)  (Error (random-int RandomGenerator-0x62962c5758f0 5 0) RangeIsEmpty)) IncorrectNumberOfArguments)]`

Probably I'm missing something since this test was working properly in previous state of random.rs.


lib/src/metta/runner/builtin_mods/random.rs
```diff
@@ -0,0 +1,292 @@
...
+            r"reset-random-generator".into(),
+            Atom::expr([ARROW_SYMBOL, ATOM_TYPE_RANDOM_GENERATOR, UNIT_ATOM]),
+            reset_random_generator));
+
+        tref.register_function(GroundedFunctionAtom::new(
+            r"flip".into(),
+            Atom::expr([ARROW_SYMBOL, ATOM_TYPE_BOOL]),
+            flip));
+
+        tref.register_token(Regex::new(r"&rng").unwrap(),move |_| { Atom::gnd(RandomGenerator::from_os_rng()) });
```
vsbogd, line 127: First reason why test fails is that you changed the way `&rng` is instantiated. Before it was a shared link to the same instance:
```
    let generator = RandomGenerator::from_os_rng();
    tref.register_token(regex(r"&rng"), move |_| { Atom::gnd(generator.clone()) });
```
Now it is new instance each time.
```suggestion
    let generator = RandomGenerator::from_os_rng();
    tref.register_token(Regex::new(r"&rng").unwrap(), move |_| { Atom::gnd(generator.clone()) });
```

vsbogd: Added this and common code fix in https://github.com/DaddyWesker/hyperon-experimental/pull/20

vsbogd (changes requested):

vsbogd:
@DaddyWesker please also fix `python/sandbox/test_gnd_conv.metta` which uses `flip` function and we can merge it.

DaddyWesker:
@vsbogd Done.

vsbogd (approved):

PR #941, opened 2025-05-28
Add json encode

DaddyWesker:
So, I've added json-encode code in metta. There is no tests or documentaion for it yet since I need to know if it is what was asked for by @Necr0x0Der . And there is no decode since I'm currently thinking on how to properly write string parser on metta (no idea right now). 
How to use encode right now:
```
> !(import! &self json)
[()]
> !(bind! &dictspace (dict-space ((k1 v1) (k2 v2) (k3 (v3 5 v4 6)))))
[()]
> !(json-encode &dictspace)
["{ 'k3' : ['v3', 5, 'v4', 6], 'k2' : 'v2', 'k1' : 'v1' }"]
```

Code is not efficient as i understand, so if any advice on making this code more efficient I'd appreciate it.

Currently what is supported - encoding of dicts (spaces on metta), lists (expressions), everything else (if number then as it is just encoded into string, other stuff is surrounded by ' '). What should I add?


vsbogd:
I am not sure we need JSON encoder in MeTTa. I would suggest JSON <-> Atomspace encoder/decoder in Rust and functions to get/set data from atomspace in MeTTa. But I am not sure what are the use-cases. @DaddyWesker @Necr0x0Der could you add use-case descriptions? Do we need ability to parameterize JSON encoding/decoding by MeTTa function?

PR #943, opened 2025-05-30
Separate hyperon atom

vsbogd:
Move `atom::gnd::tests` into a `lib/tests/metta.rs` and move the `atom` module into a separate `hyperon-atom` crate. 
This allows using `hyperon-atom` as a dependency of the external builtin library implementation (such as DAS integration) and using this builtin library as a dependency of the `hyperon` crate.

Dependency diagram:
diagram


arturgontijo (approved):
Great one, thanks... LGTM!

PR #944, opened 2025-06-03
Add MeTTa DAS Bus Client

arturgontijo:
~~DO NOT MERGE IT!!! As we plan to have a tag/version of the DAS repo to use it here (currently using `master`)~~

This PR adds a new built-in module called `das` that uses the `metta-bus-client` rust package to send pattern matching queries to a DAS Service Bus.

In order to test it, one could follow the steps from [Setup DAS locally](https://github.com/singnet/das/tree/master/3rd_party_slots/rust_metta_bus_client#setup-das-locally) and then:

```sh
cargo build --release --features=das
```

Then run the MeTTa REPL binary:
```sh
./target/release/metta-repl 
Visit https://metta-lang.dev/ for tutorials.
Execute !(help!) to get list of the standard library functions.
> !(import! &self das)
[()]
> !(bind! &das (new-das! (localhost:8080) (localhost:35700)))
[()]
> !(match &das (Similarity "human" $S) ($S))
[(a408f6dd446cdd4fa56f82e77fe6c870), (3225ea795289574ceee32e091ad54ef4), (181a19436acef495c8039a610be59603)]
> 
```

@vsbogd let me know if I'm in the right direction about adding new modules here.


lib/src/space/distributed/mod.rs
vsbogd, line 1: DAS atomspace could also be implemented in `https://github.com/singnet/das` repository. Content of the top-level `space` module (i.e `./lib/src/space/mod.rs` file) should be moved into a separate crate for doing this. @arturgontijo may I ask you to help making this change?

vsbogd: As `index` is also used inside DAS atomspace implementation then `./lib/src/space/grounding/index` should also be moved into the same `hyperon-space` crate.

lib/src/space/distributed/mod.rs
```diff
@@ -0,0 +1,145 @@
...
+    fn as_match(&self) -> Option<&dyn CustomMatch> {
+        Some(self)
+    }
+}
+
+impl CustomMatch for DistributedAtomSpace {
+    fn match_(&self, other: &Atom) -> MatchResultIter {
+        Box::new(self.query(other).into_iter())
+    }
+}
```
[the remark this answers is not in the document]
vsbogd: See https://github.com/trueagi-io/hyperon-experimental/pull/947

lib/src/space/distributed/mod.rs
arturgontijo: > may I ask you to help making this change?

Sure, I can do that!

About `index`, I'm sure that I copied that from another code/example...I'll clean that up.

lib/src/space/distributed/mod.rs
```diff
@@ -0,0 +1,145 @@
...
+    Space,
+    SpaceCommon,
+    SpaceEvent,
+    SpaceMut,
+    SpaceVisitor
+};
+
+#[derive(Clone)]
+pub struct DistributedAtomSpace {
+    index: AtomIndex,
```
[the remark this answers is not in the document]
arturgontijo: I'm pretty sure we can remove it...I added it based on another example.

Cargo.toml
```diff
@@ -23,5 +23,10 @@ regex = "1.11.0"
 log = "0.4.0"
 env_logger = "0.8.4"
 
+[patch.'https://github.com/trueagi-io/hyperon-experimental.git']
+hyperon-common = { path = "./hyperon-common", version = "0.2.5" }
+hyperon-atom = { path = "./hyperon-atom", version = "0.2.5" }
+hyperon-space = { path = "./hyperon-space", version = "0.2.5" }
+
```
vsbogd, line 30: Not sure if this section is needed. Is there a reason to add it?
```suggestion
```

lib/src/metta/runner/builtin_mods/das.rs
```diff
@@ -0,0 +1,92 @@
...
+
+grounded_op!(NewDasOp, "new-das!");
+
+impl Grounded for NewDasOp {
+    fn type_(&self) -> Atom {
+        Atom::expr([
+            ARROW_SYMBOL,
+            ATOM_TYPE_SYMBOL,
+            ATOM_TYPE_SYMBOL,
+            rust_type_atom::<DynSpace>(),
```
vsbogd, line 61: I am sorry, missed it on the first review, actually type should be represented by the `ATOM_TYPE_SPACE` constant which I just moved into `hyperon_space` crate (see #954 )
```suggestion
            ATOM_TYPE_SPACE,
```

lib/src/metta/runner/builtin_mods/das.rs
```diff
@@ -0,0 +1,92 @@
+use std::sync::{Arc, Mutex};
+
+use hyperon_atom::{rust_type_atom, Atom, CustomExecute, ExecError, Grounded};
+use hyperon_space::DynSpace;
```
vsbogd, line 4: See comment about type below.
```suggestion
use hyperon_space::{DynSpace, ATOM_TYPE_SPACE};
```

vsbogd (commented):
Last couple of things I missed on the previous review.

Cargo.toml
arturgontijo: The issue here is when we run `cargo b -r --features=das` the compiler will use two difference sources for these dependencies, one will be the locals `./xxx` and the other (as set [here](https://github.com/singnet/das/blob/master/3rd_party_slots/rust_metta_bus_client/Cargo.toml#L14-L16)) will come from github. Even both codes are identical, `cargo` will fail to compile.

By using `patch` we force any dependency that uses github as "source" to fallback to what we have locally, that then allow `cargo` to use the same source code to compile.

Another option would be publishing these deps on crates.io so we fetch source from it here and in the `rust_metta_bus_client` code.

IMO using `patch` is ok for now, as both codebases are under heavily development. Once we get stable versions we can publish things with version tags on crates.io

arturgontijo:
Hey @vsbogd sorry for the huge delay on this one...We finally got a version tag that exposes the necessary code for this module.
When you get a chance, please do another review round here.

vsbogd (approved):

vsbogd:
Thanks @arturgontijo , looks good to me and I am ready to merge it.

There are two alternatives:
1. make it available by default, then it should be added into default feature list in Cargo.toml
2. make it optional, then I would add it as a default into Release builds

For the second alternative there are two places to edit:
- first one is the `Dockerfile` https://github.com/trueagi-io/hyperon-experimental/blob/500d7777c72032255555af5ca5dd335def7121c9/Dockerfile#L32 to add it into the Docker `metta-repl`
- second one is more tricky add it as an option to the CMake files, see https://github.com/trueagi-io/hyperon-experimental/blob/500d7777c72032255555af5ca5dd335def7121c9/python/CMakeLists.txt#L4 and all usages of the 'GIT' in code.

I don't see why we cannot make it default choice (i.e. option (1)), but I would like to hear your thoughts on this. Option (2) can be implemented as a separate PR.

vsbogd:
BTW probably this PR obsoletes:
- https://github.com/trueagi-io/hyperon-experimental/tree/main/python/integration
- https://github.com/trueagi-io/hyperon-experimental/tree/main/python/hyperon/exts/das_gate

@CICS-Oleg , what do you think?

Cargo.toml
vsbogd: It makes sense, thanks!

arturgontijo:
> make it available by default, then it should be added into default feature list in Cargo.toml

@vsbogd I vote for making it default (as this option seems more simple then `2`, and we can change that in future if needed).

> check that DAS integration is a part of the Release build

Will do and let you know

arturgontijo:
Hey @vsbogd, I was able to create the image (from the repo root's `Dockerfile`) and also to create the python wheel:
```
-rw-r--r--  1 ... ...  13M Jul 23 20:35 hyperon-0.2.6+gded17330-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
```

I had to download `protoc` from their github release URL as `yum install -y protobuf-compiler` was installing a version 2 of `protoc` and DAS requires `proto3` ([ref](https://github.com/singnet/das-proto/blob/master/atom_space_node.proto#L1))

vsbogd (approved):

vsbogd:
@arturgontijo thanks for trying, ok, I think I can merge it as is and if I can improve it using Conan I will do it.

PR #946, opened 2025-06-04
Add json

DaddyWesker:
So, currently I've added json-decode part of json library. While I'm thinking on json-encode, I'd like to hear if this json-decode is what was asked for. Usage:

```
!(import! &self json)
; json dicts <-> metta space:
!(bind! &decdict (json-decode "{\"key1\" : \"value1\", \"key2\" : \"value2\", \"key3\" : [5, 4, 3]}")) 
!(match &decdict ($k $v) ($k $v))
; [("key1" "\"value1\""), ("key2" "\"value2\""), ("key3" (5 4 3))]
```

No documentation or tested added yet. I'll wait till those changes approved.


Necr0x0Der:
> if this json-decode is what was asked for

Yes, I guess so. Will it also work with nested dicts (turning them into nested spaces)?

> `!(match &decdict ($k $v) ($k $v))`

There were functions to access space-dict elements. `match` is ok for testing, but are these functions available?

lib/src/metta/runner/builtin_mods/json.rs
```diff
@@ -0,0 +1,126 @@
...
+        Value::Object(_) => decode_object(&v.clone()),
+        Value::Number(_) => {
+            if v.is_i64() {
+                Atom::gnd(Number::Integer(v.as_i64().unwrap()))
+            }
+            else if v.is_f64() {
+                Atom::gnd(Number::Float(v.as_f64().unwrap()))
+            }
+            else {
+                Atom::gnd(Number::Integer(v.as_u64().unwrap() as i64))
```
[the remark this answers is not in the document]
DaddyWesker: @vsbogd Probably you right. But should we use panic! or just throw an error?

DaddyWesker:
@vsbogd @Necr0x0Der PR is ready for review.

lib/src/metta/runner/builtin_mods/json.rs
```diff
@@ -0,0 +1,223 @@
...
+        let mut tref = target.tokenizer().borrow_mut();
+
+        tref.register_function(GroundedFunctionAtom::new(
+            r"json-encode".into(),
+            Atom::expr([ARROW_SYMBOL, ATOM_TYPE_ATOM, ATOM_TYPE_STRING]),
+            json_encode));
+
+        tref.register_function(GroundedFunctionAtom::new(
+            r"json-decode".into(),
+            Atom::expr([ARROW_SYMBOL, ATOM_TYPE_STRING, ATOM_TYPE_SPACE]),
```
vsbogd, line 46: It returns not space but Space, String, Number or Bool depends on first argument:
```suggestion
            Atom::expr([ARROW_SYMBOL, ATOM_TYPE_STRING, ATOM_TYPE_ATOM]),
```

lib/src/metta/runner/builtin_mods/json.rs
```diff
@@ -0,0 +1,223 @@
...
+        res_str = res_str + ", " + &key + ": " + &value_str;
+    }
+    res_str.remove(0);
+    res_str.remove(0);
+    res_str = format!("{{{}}}", res_str);
+    Ok(res_str)
+}
+
+fn encode_atom(input: &Atom) -> Result<String, ExecError> {
+    let metatype = atom_to_string(&get_meta_type(&input));
```
vsbogd, line 91: It is unnecessary to convert the meta-type into a string. If you need to check if it is expression you even don't need a meta-type. You can just use `TryInto::<&ExpressionAtom>` results to check whether it is expression or not.

lib/src/metta/runner/builtin_mods/json.rs
```diff
@@ -0,0 +1,223 @@
...
+    let space = DynSpace::new(GroundingSpace::new());
+    let typ = get_atom_types(&space, &input);
+    if metatype == "Expression" {
+        let exp = TryInto::<&ExpressionAtom>::try_into(input)?;
+        match encode_list(exp.clone().into_children()) {
+            Ok(encoded_list) => Ok(encoded_list),
+            Err(err) => Err(err),
+        }
+    }
+    else if atom_to_string(&typ[0]) == "Number" {
```
vsbogd, line 101:
```suggestion
    else if typ[0] == ATOM_TYPE_NUMBER {
```

lib/src/metta/runner/builtin_mods/json.rs
```diff
@@ -0,0 +1,223 @@
...
+        let number_atom_string = atom_to_string(&input);
+        Ok(number_atom_string)
+    }
+    else if atom_to_string(&typ[0]) == "SpaceType" {
```
vsbogd, line 105:
```suggestion
    else if typ[0] == ATOM_TYPE_SPACE {
```

lib/src/metta/runner/builtin_mods/json.rs
```diff
@@ -0,0 +1,223 @@
...
+        match encode_dictspace(&input) {
+            Ok(str) => Ok(str),
+            Err(err) => Err(err)
+        }
+    }
+    // String should be encoded using serde_json since it will encode "a" with additional escape
+    // characters. Symbols like just 'a' are not supported by serde_json so proposition is to encode
+    // them like strings.
+    else {
```
vsbogd, line 114: Not sure why separate branch for boolean is absent here?

lib/src/metta/runner/builtin_mods/json.rs
```diff
@@ -0,0 +1,223 @@
...
+fn encode_list(input: Vec<Atom>) -> Result<String, ExecError> {
+    let mut res_str = "".to_string();
+    for atom in input {
+        match encode_atom(&atom) {
+            Ok(encoded) => {res_str = res_str + ", " + &encoded;},
+            Err(err) => return Err(err),
+        }
+    }
+    res_str.remove(0);
+    res_str.remove(0);
```
vsbogd, line 62: These two removes require copying content of the string twice. It is better to not add ", " before first atom in a string. You can do it by getting first atom from array and after that processing tail of the array in a loop.

lib/src/metta/runner/builtin_mods/json.rs
```diff
@@ -0,0 +1,223 @@
...
+    Ok(format!("[{}]", res_str))
```
vsbogd, line 63: Simpler to add `[` before and `]` after processing atoms. This line requires allocating memory for the whole JSON string twice.

lib/src/metta/runner/builtin_mods/json.rs
```diff
@@ -0,0 +1,223 @@
...
+}
+
+fn encode_dictspace(input: &Atom) -> Result<String, ExecError> {
+    let space = Atom::as_gnd::<DynSpace>(input).unwrap();
+    let result = space.borrow().subst(&Atom::expr([Atom::var("k"), Atom::var("v")]),
+                                      &Atom::expr([Atom::var("k"), Atom::var("v")]));
+    if result.len() == 0 {
```
vsbogd, line 70: Not sure why we cannot encode empty dictionary as JSON `{}`?

lib/src/metta/runner/builtin_mods/json.rs
```diff
@@ -0,0 +1,223 @@
...
+            Ok(str) => str.to_string(),
+            Err(err) => return Err(err),
+        };
+        let value = res_expr.children().get(1).unwrap();
+        let value_str = encode_atom(value)?;
+        res_str = res_str + ", " + &key + ": " + &value_str;
+    }
+    res_str.remove(0);
+    res_str.remove(0);
+    res_str = format!("{{{}}}", res_str);
```
vsbogd, line 86: Same as in `encode_list` above.

lib/src/metta/runner/builtin_mods/json.rs
```diff
@@ -0,0 +1,223 @@
...
+    }
+    res_str.remove(0);
+    res_str.remove(0);
+    Ok(format!("[{}]", res_str))
+}
+
+fn encode_dictspace(input: &Atom) -> Result<String, ExecError> {
+    let space = Atom::as_gnd::<DynSpace>(input).unwrap();
+    let result = space.borrow().subst(&Atom::expr([Atom::var("k"), Atom::var("v")]),
+                                      &Atom::expr([Atom::var("k"), Atom::var("v")]));
```
vsbogd, line 69: I would use `Space::visit` instead, because it should be more effective at the moment for a big dictionaries.

lib/src/metta/runner/builtin_mods/json.rs
```diff
@@ -0,0 +1,223 @@
...
+    }
+}
+
+fn decode_value(v: &Value) -> Result<Atom, ExecError> {
+    match v {
+        Value::Array(_) => {match decode_array(v.as_array().unwrap()) {
+            Ok(atom) => Ok(atom),
+            Err(err) => Err(err),
+        }},
+        Value::Null => Ok(Atom::gnd(Str::from_str("null"))),
```
vsbogd, line 139: I would suggest using `Null` or `null` symbol instead as basically `null` in JSON means no value. String is not distinguishable from `"null"` string as a value.

lib/src/metta/runner/builtin_mods/json.rs
```diff
@@ -0,0 +1,223 @@
...
+}
+
+fn decode_object(obj: &Value) -> Result<Atom, ExecError> {
+    let dict = obj.as_object().unwrap();
+    let keys = dict.keys();
+    let space = DynSpace::new(GroundingSpace::new());
+    for key in keys {
+        let value = dict.get(key).unwrap();
+        match decode_value(value) {
+            Ok(decoded) => space.borrow_mut().add(Atom::expr([Atom::gnd(Str::from_string(key.clone())), decoded])),
```
vsbogd, line 167: I would not wrap `GroundingSpace` into a `DynSpace` from very beginning. Instead you can fill `GroundingSpace` instance and then wrap it into a `DynSpace` before making the atom. It will make borrowing unnecessary:
```suggestion
            Ok(decoded) => space.add(Atom::expr([Atom::gnd(Str::from_string(key.clone())), decoded])),
```

lib/src/metta/runner/builtin_mods/json.rs
```diff
@@ -0,0 +1,223 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::metta::*;
+    use crate::metta::runner::run_program;
+
+    #[test]
+    fn test_json() {
```
vsbogd, line 201: No boolean or null values are tested.

lib/src/metta/runner/builtin_mods/json.rs
DaddyWesker: @vsbogd because it could be not dictionary but just some space without any key-value pairs. I've asked @Necr0x0Der what to do in such cases and wrote this code relying on his words. Maybe you right and it will be better but I think it should be discussed since I'm not sure this is the right way to return `{}` for every space without key-value pairs.

DaddyWesker:
@vsbogd  everything is ready except replacing match with visit. I'm still figuring this out.

vsbogd (dismissed):

Necr0x0Der:
I have a few questions regarding `get-keys` and `dict-space`.
First of all, why `(: get-keys (-> Grounded Expression))`? Apparently, `get-keys` can return `Atom`, not `Expression`. And it should be a more typical case. This needs to be fixed.
Then, I wonder, why its implementation is
```
(= (get-keys $dictspace)
   (function
     (chain (unify $dictspace ($key $value) $key Empty) $t (return $t)) ))
```
instead of just
```
(= (get-keys $dictspace)
   (match $dictspace ($key $value) $key))
```
?
Is there any difference, @vsbogd , @DaddyWesker ?

Interestingly, with `Expression` return type, the following code returns `[3, C]`
```
!(bind! &space (new-space))
!(add-atoms &space (
    ((+ 1 2) B)
    (C D)
))
(: get-keys (-> Grounded Expression))
(= (get-keys $dictspace)
   (function
     (chain (unify $dictspace ($key $value) $key Empty) $t (return $t)) ))
!(get-keys &space)
```
While with `Atom` it returns `[(+ 1 2), C]`.
At the same time, `(= (get-keys_ $dictspace) (match $dictspace ($key $value) $key))` works differently (it doesn't work with `Atom` return type, because `match` is not executed, but one can have `%Undefined%` return type and use `id` to avoid reduction).
In any case, I'm ok with this implementation (with `Atom` return type it makes more sense), but I wonder if there are real reasons for it.

As per `dict-space`, I wonder if we need to use `add-reducts` or `add-atoms` there. I don't have a strong opinion. On the one hand, one may want to store atoms as is, say,
```
!(bind! &space (dict-space
    (((+ 1 2) (+ 2 1))
     (C D))))
```
to store `((+ 1 2) (+ 2 1))` as a non-reduced pair. OTOH, one may also want to create a space filled in with reduct, feeding `dict-space` with expressions to be calculated. It's not too important, since one may resort to explicit `add-reducts` and `add-atoms` with `new-space`, but I wonder what will be a more frequent case. Nevertheless, we can keep `add-reducts` for now. 

However, I wonder why the implementation is
`(= (dict-space $expr) (chain (eval (new-space)) $dictspace (let $res (add-reducts $dictspace $expr) $dictspace)))`
Why not
```
(= (dict-space $expr)
   (let* (($dictspace (new-space))
          ($res (add-reducts $dictspace $expr)))
        $dictspace))
```
Since there is `let` in any case, it seems more readable to use `let*`.

DaddyWesker:
@Necr0x0Der 

About reasons on why get-keys implementation looks like this.
As you probably remember, I've started to write encode part of json purely on Metta. And I had subfunctiuon encode-dict which takes space and encodes it. It was needed to check if get-keys returns any keys or not (something like this, but not exactly this `(let $keys (get-keys &self) (== $keys Empty))`). But in case of match implementation (which looks more adequate at the first glance) I couldn't write regular `(if (== $keys Empty) ...)` since when match gives Empty it just stops further interpretation and return []. I've asked Vitaly how to bypass it and one of his suggestions was to "use minimal Metta". And this means following code:
```
(: get-keys (-> Grounded Expression))
(= (get-keys $dictspace) 
   (function
     (chain (unify $dictspace ($key $value) $key Empty) $t (return $t)) ))


!(chain (context-space) $space
   (chain (eval (get-keys &self)) $keys
     (unify Empty $keys True False) ))
```

It worked back in the day for that json-encode-pure-Metta-implementation and I've just left it as it is. I'm not sure if I should return previous implementation. Maybe @vsbogd  will make his suggestion on that?


>  I wonder what will be a more frequent case

Well, we sure can have two dict-space functions (with atoms being reduced or not). Or just make one function with bool flag (though I'm not sure on performance cost of using additional if there).


> Since there is let in any case, it seems more readable to use let*

Well, I can replace this implementation If you think let* variant more readable, no rpoblem.

vsbogd:
> I'm not sure if I should return previous implementation. Maybe @vsbogd will make his suggestion on that?

Whatever code works and looks more logical. Now probably `match` is more compact, but I don't have strong opinion here.

> Well, we sure can have two dict-space functions (with atoms being reduced or not). Or just make one function with bool flag (though I'm not sure on performance cost of using additional if there).

In general I am against two versions of functions: reducing and not reducing. I would say in normal programming language this should be solved via function composition. I.e. we have function which creates dict without reduction (`dict-space`), we have function which reduces atoms (`metta` or whatever) and if we need to create dict with reduced atoms then we first apply `metta` to the list of pairs and then pass the result to the `dict-space`. Having two versions of functions make API unreasonably large and difficult for awkward.

DaddyWesker:
> we have function which creates dict without reduction (dict-space), we have function which reduces atoms (metta or whatever) and if we need to create dict with reduced atoms then we first apply metta to the list of pairs and then pass the result to the dict-space

So, your suggestion is to make dict-space add atoms without reduction? And if user wants to add them reducted, he need to reduce them first and then use dict-space, right?

vsbogd:
> So, your suggestion is to make dict-space add atoms without reduction? And if user wants to add them reducted, he need to reduce them first and then use dict-space, right?

Yes, it is my view on this. I am not sure whether @Necr0x0Der is agree with this.

Necr0x0Der:
> Yes, it is my view on this. 

Well, tbh, I'd keep `add-reducts`. The reason is the following usage, which I think will be the most frequent one:
```
(dict-space
   (A (dict-space (C D) (E F)))
   (B (dict-space (G H) (I G)))
)
```
Basically, creating nested dicts is very convenient with reduction, while it will be awkward without it.

PR #963, opened 2025-07-03
Fix stdlib

vsbogd:
1. Use `SpaceType` where it is appropriate
2. Replace `if-decons-expr` by `(chain (decons-atom $expr) $ht (unify ($head $tail) $ht $then $else))` which works faster and looks more compact.

### DIFF
--- a/lib/src/metta/runner/stdlib/stdlib.metta
+++ b/lib/src/metta/runner/stdlib/stdlib.metta
@@ -47,7 +47,7 @@
     (@param "Atom to be evaluated, can be reduced via equality expression (= ...) or by calling a grounded function")
     (@param "Space to evaluate atom in its context")))
   (@return "Result of evaluation"))
-(: evalc (-> Atom Grounded Atom))
+(: evalc (-> Atom SpaceType Atom))
 
 (@doc chain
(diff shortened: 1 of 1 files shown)


DaddyWesker:
Looks fine, but I've got a question. You've removed `if-decons-expr` entirely and replaced its usage with  `(chain (decons-atom $expr) $ht (unify ($head $tail) $ht $then $else))` everywhere `if-decons-expr` was used previously. But why not just replace `if-decons-expr` implementation and leave its usage in other functions? Or `if-decons-expr` is entirely inner function which shouldn't be used by metta user so we don't need it anymore?

DaddyWesker (approved):

zariuq:
In my HE code, if-decons-expr wasn't wrapped in unify so it looks much cleaner than with this substitution.

If one uses the, "one additional call makes code slower", reasoning, one would almost always program in the lowest-level lanugage possible and would not use APIs.  One would unwrap almost all API calls with the internal functions for optimized performance.

My experience using HE-MeTTa is that of desiring more high-level stdlib functions so that I don't need to spend time on lower level control logic.  I'd be curious to see a pool of the "stdlib-like utility functions" MeTTa developers implement for their own convenience (so that I don't need to reinvent the wheel).

I'm not sure that if-decons-expr should necessarily survive, however 🤓

PR #966, opened 2025-07-07
Eval cleaning

DaddyWesker:
So, I've cleaned some evals and `cargo test` runs well for me now.


lib/src/metta/runner/stdlib/stdlib.metta
vsbogd, line 1: As discussed, I would suggest rolling this changes back. If we need to use MeTTa representation we should rewrite these functions in MeTTa instead.

PR #968, opened 2025-07-08
Divide refactored

DaddyWesker:
So, now `!(/ 6 0)` will not result in rust panic, but in metta error instead.

### DIFF
--- a/lib/src/metta/runner/stdlib/arithmetics.rs
+++ b/lib/src/metta/runner/stdlib/arithmetics.rs
@@ -50,7 +50,6 @@ macro_rules! def_binary_number_op {
 def_binary_number_op!(SumOp, +, ATOM_TYPE_NUMBER, Number);
 def_binary_number_op!(SubOp, -, ATOM_TYPE_NUMBER, Number);
 def_binary_number_op!(MulOp, *, ATOM_TYPE_NUMBER, Number);
-def_binary_number_op!(DivOp, /, ATOM_TYPE_NUMBER, Number);
 def_binary_number_op!(ModOp, %, ATOM_TYPE_NUMBER, Number);
 def_binary_number_op!(LessOp, <, ATOM_TYPE_BOOL, Bool);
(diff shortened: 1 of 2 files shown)


lib/src/metta/runner/stdlib/arithmetics.rs
```diff
@@ -125,6 +124,38 @@ impl CustomExecute for NotOp {
...
+    fn as_execute(&self) -> Option<&dyn CustomExecute> {
+        Some(self)
+    }
+}
+
+impl CustomExecute for DivOp {
+    fn execute(&self, args: &[Atom]) -> Result<Vec<Atom>, ExecError> {
+        let arg_error = || ExecError::from("Divide expects two numbers: dividend and divisor");
+        let dividend: f64 = args.get(0).and_then(Number::from_atom).ok_or_else(arg_error)?.into();
+        let divisor: f64 = args.get(1).and_then(Number::from_atom).ok_or_else(arg_error)?.into();
```
vsbogd, line 150: 1. Previous implementation didn't always promote `Number` to `f64`. Thus in `(/ 5 2)` returns `2` as a result not `2.5`. 
I would keep this behavior if there is no issue which asks for changes.

2. It is incorrect to match float with `0.0`, there are at least two zeroes `+0.0` and `-0.0`. `f64` doesn't have an issue with dividing something by `0.0` anyway, it returns `inf` instead. Thus I would check for zero only for the `Integer` case.

Thus the code could be the following (it is a pseudo code):
```
                let a = args.get(0).and_then(Number::from_atom).ok_or_else(arg_error)?;
                let b = args.get(1).and_then(Number::from_atom).ok_or_else(arg_error)?;

                let (a, b) = Number::promote(a, b);
                let res: Number = match (a, b) {
                    (Number::Integer(a), Number::Integer(0)) => Err(ExecError::from("DivisionByZero")),
                    (Number::Integer(a), Number::Integer(b)) => (a / b).into(),
                    (Number::Float(a), Number::Float(b)) => (a / b).into(),
                    _ => panic!("Unexpected state"),
                };
```

Issue #861, opened 2025-02-27
assertEqualMsg would be useful


Necr0x0Der:
While `assertEqual` is ok for unit tests, asserts with semantically meaningful error messages in real code would be helpful Thus, a version of `assertEqual`, which also accepts a message could be added.

Necr0x0Der:
> This message should be returned on assert fail as i understand?

I'd say, it should be returned as a part of `Error` expression. It doesn't make sense to return the message itself without raising a error. Otherwise, it is not an assert, right?

Necr0x0Der:
closing via #974

PR #974, opened 2025-07-11
assert and asserts with messages added

DaddyWesker:
So, I've added `assert` as @ngeiswei  asked [here](https://github.com/trueagi-io/hyperon-experimental/issues/900) , please check if it is what you've asked.

I've also added some asserts with custom messages as @Necr0x0Der asked [here](https://github.com/trueagi-io/hyperon-experimental/issues/861) . Please also check if it is fine with you. 

One thing about asserts with custom message. Regular assert will stop program from further interpretation on failure, but -Msg asserts won't. I'm not sure about this behavior.  As an alternative, instead of just outputting custom message it could be wrapped into Error like regular output of assert. So some 
`!(assertEqualMsg (+ 1 2) 4 "Test message")` will output `(Error (assertEqualMsg (+ 1 2) 4 "Test message") "Test message")` instead of just ("Test message"). What behavior is preferable?

### DIFF
--- a/lib/tests/test_stdlib.metta
+++ b/lib/tests/test_stdlib.metta
@@ -49,3 +49,43 @@
 !(assertEqualToResult
    (id (noeval (+ 1 2)))
    (3))
+
+; assert
+
+!(assert True)
+!(assert (or True False))
+!(assertEqual (assert (and True False)) (Error (assert (and True False)) ((and True False) not True)))
+
+; asserts with custom messages
+; first part of tests is to check if asserts works as intended when assert should output [()].
+; So those tests are duplicate of tests for original asserts
+
+!(assertEqualToResultMsg
+   (noeval (+ 1 2))
+   ((+ 1 2))
+   "Test message")
+
+!(assertEqualToResultMsg
+   (= (f) (+ 1 2))
+   ((= (f) 3))
+   "Test message")
+
+(= (f1 $x) $x)
+(= (f2 $x) $x)
+
+!(assertEqualMsg (+ 1 2) 3 "Test message")
+!(assertAlphaEqualMsg (f1 $x) (f2 $x) "Test message")
+
+(= (f3) $x)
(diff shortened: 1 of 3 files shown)


lib/tests/test_stdlib.metta
```diff
@@ -49,3 +49,43 @@
...
+
+!(assertEqualMsg (+ 1 2) 3 "Test message")
+!(assertAlphaEqualMsg (f1 $x) (f2 $x) "Test message")
+
+(= (f3) $x)
+!(assertAlphaEqualToResultMsg ((f3) (f3)) (($x $y)) "Test message")
+
+; next part of tests is to test if message is being output on assert failure
+
+!(assertEqual (assertEqualMsg (+ 1 2) 4 "Test message") ("Test message"))
```
Necr0x0Der, line 84: Do I understand correctly that `(assertEqualMsg (+ 1 2) 4 "Test message")` returns `("Test message")`? I supposed it should return a `Error` expression with the given message. Otherwise, it doesn't work as an assert?

DaddyWesker: @Necr0x0Der well currently yes. As i said in the description:
```
One thing about asserts with custom message. Regular assert will stop program from further interpretation on failure, but -Msg asserts won't. I'm not sure about this behavior. As an alternative, instead of just outputting custom message it could be wrapped into Error like regular output of assert. So some
!(assertEqualMsg (+ 1 2) 4 "Test message") will output (Error (assertEqualMsg (+ 1 2) 4 "Test message") "Test message") instead of just ("Test message"). What behavior is preferable?
```

DaddyWesker:
@Necr0x0Der  I've fixed what asserts with message returns. It looks like this currently:
`(assertEqualMsg (+ 1 2) 4 "Test message") ; (Error (assertEqualMsg (+ 1 2) 4) "Test message")`

Is this fine?

Necr0x0Der:
> Is this fine?

Looks acceptable to me.

Issue #923, opened 2025-04-30
panic in quoted tuple


Necr0x0Der:
**Describe the bug**
Panic happens in any operation, which wraps a quoted tuple with `println!` call.

**To Reproduce**
```
! (== (quote ((println! "Test"))) (quote ()))
```

**Expected behavior**
`False`

**Actual behavior**
Panic

**Additional context**
```
! (== ((println! "Test")) ())
```
also panics, while
`! ((println! "Test"))`
produces "incorrect number of arguments error"

The following examples don't panic:
```
!(== (quote (println! "Test")) (quote ()))
!(== (quote (_ (println! "Test"))) (quote ()))
!(quote ((println! "Test")))
```

Other cases:
```
!(decons-atom ((println! "hello2") (println! "world2")))
```
doesn't panic, while
```
!(let ($h $t) (decons-atom ((println! "hello2") (println! "world2")))
 OK)
```
panics.

```
(: eqt (-> Atom Atom Atom))
(= (eqt $x $x) True)
(= (eqt $x $y) Empty)
(: eq (-> Atom Atom Bool))
(= (eq $x $y)
   (case (eqt $x $y)
    ((True True)
     ($_ False))))
!(eq (quote ((println! "Test"))) (quote ()))
```
doesn't panic.
```
!(if (eq (quote ((println! "Test"))) (quote ())) NOOK OK)
```
panics again.

I tried to replace `println!` with some other grounded functions, and they seem to be OK.
Wrapping `println!` into an ordinary function removes panic.

Necr0x0Der:
> could also comment on why UNIT_TYPE is (->) whether we can change it.

Well, `()` is formally an application of an empty function to empty arguments. Syntactically, its type is `(->)`. `()` can be viewed as a morphism from empty set to empty set. There is only one such morphism, and thus it's unit. As a morphism, it should have a type signature, and you cannot assign any other type signature besides `(->)`... Well, after we introduced `Empty` as an explicit symbol, another option would be to have `(: () (-> Empty Empty))`. It is not precise, because `Empty` stands not for an empty set, but an empty (false/unpopulated) type. But I'm not sure if we are trying to follow category-theoretic semantics in MeTTa.
I wonder if it panics, though. There can be different solutions. E.g., `()` could have `(-> Atom)` type, because `()` is not reduced (it doesn't really return a non-existent element of empty set on application). Thus, `(-> Atom)` makes perfect sense for `()` within MeTTa, if it works.

[...]

PR #991, opened 2025-08-04
fixes for issue 923

DaddyWesker:
This fixes issue #923 . Now type (->) won't be interpreted as function.


vsbogd (approved):

Necr0x0Der:
This looks like a conservative fix for #923 , and I'm definitely OK with it. The issue is that it obscures the question about `()` type. Should we keep it as `(->)`, or not introducing this fix, but treating `()` as `(-> Atom)` makes more sense. @vsbogd ?

vsbogd:
I think we can keep the type of `()` as `(->)`. `(->)` is not callable anyway and `is_func` is more about being callable. `(-> Atom)` implies returning some value while `()` doesn't return one. Thus I would just merge this PR.

PR #993, opened 2025-08-06
Allow returning bindings from executable grounded atoms

### LINKED ISSUE #911: match doesn't return bindings
**Description**
`match` doesn't bind variables from patterns, which causes confusion.

**To Reproduce**
```
(a A B)
(= (isa $x $y) (match &self (a $x $y) True))
!((isa $a B) $a)
```

**Expected behavior**
`[(True A)]`

**Actual behavior**
`[(True $a)]`

**Additional context**
It's not a big deal since the following code works as expected:
```
(a A B)
(= (isa $x $y) (unify &self (a $x $y) True Empty))
!((isa $a B) $a)
```

However, it is confusing that `unify` works as needed, while `match` works differently, although the functions are very similar. We can still say that this is a feature, not a bug. But then we need to standartize this behavior (by adding unit tests, notes in tutorials?) and understand it (is it because `match` is a purely grounded operation)?

vsbogd:
Adding new execute_bindings method into the CustomExecute trait in order to allow returning variable bindings if function is able doing it. Default implementation falls back to the call of CustomExecute::execute method. Modify interpreter to call CustomExecute::execute_bindings. Modify MatchOp implementation to return variable bindings. Fixes #911

### DIFF
--- a/hyperon-atom/src/lib.rs
+++ b/hyperon-atom/src/lib.rs
@@ -495,6 +495,9 @@ pub trait Grounded : Display {
     }
 }
 
+/// Boxed iterator type to simplify type annotations
+pub type BoxedIter<'a, T> = Box<dyn Iterator<Item=T> + 'a>;
+
 /// Trait for implementing custom execution logic. Using this trait one can
 /// represent a grounded function as an atom. In order to make it work
 /// one should also implement [Grounded::as_execute] method.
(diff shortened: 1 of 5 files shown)


Necr0x0Der:
I wonder if it will fix this?

https://github.com/trueagi-io/hyperon-experimental/blob/ec8b7c65d6edad28058f90c6be71708f5e910148/python/tests/scripts/b2_backchain.metta#L70

vsbogd:
@Necr0x0Der if I got you correctly then yes, it fixes this example see https://github.com/trueagi-io/hyperon-experimental/pull/993/commits/531f88f931b8d8d4cb8ea90d5fa8ebbf08fad7a5

DaddyWesker (approved):

PR #996, opened 2025-08-07
case refactor

DaddyWesker:
Case was rewritten on pure metta, CaseOp - deleted.

### DIFF
--- a/lib/src/metta/runner/stdlib/stdlib.metta
+++ b/lib/src/metta/runner/stdlib/stdlib.metta
@@ -325,9 +325,9 @@
   (@return "Result which corresponds to the pattern which is matched with the passed atom first"))
 (: switch (-> %Undefined% Expression %Undefined%))
 (= (switch $atom $cases)
-   (switch-minimal $atom $cases))
+   (id (switch-minimal $atom $cases)))
 
-(: switch-minimal (-> Atom Expression %Undefined%))
+(: switch-minimal (-> Atom Expression Atom))
 (= (switch-minimal $atom $cases)
   (function (chain (decons-atom $cases) $list
     (chain (eval (switch-internal $atom $list)) $res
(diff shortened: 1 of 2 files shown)


lib/src/metta/runner/stdlib/core.rs
```diff
@@ -407,6 +346,18 @@ mod tests {
...
+    #[test]
+    fn metta_case_error() {
+        let program = "
+            (= (err) (Error (err) \"Test error\"))
+            !(assertEqual (case (err) (
+              ((Error $a \"Test error\") ok)
+              ($_ (Error $_ \"Error is expected\"))
+              )) ok)
+        ";
+        assert_eq!(run_program(program), Ok(vec![vec![UNIT_ATOM]]));
```
vsbogd, line 358: I would not use MeTTa "assert" in Rust unit tests if it is possible. It makes test dependent on "assert" behavior. 
```suggestion
            (= (err) (Error (err) \"Test error\"))
            !(case (err) (
              ((Error $a \"Test error\") ())
              ($other (Error $other \"Error is expected\"))
              ))";
        assert_eq!(run_program(program), Ok(vec![vec![UNIT_ATOM]]));
```

TeamSPoon:
I would think it usefull to let switch-minimal return `Atom` so that it has wider usecases.. You already expected to eval; the return when used in a place or two

vsbogd:
> I would think it usefull to let some place of the switch-* logic return `Atom` (unevaluated) so that it has wider usecase.. You already expect to eval the return when used in a place or two anyway

I believe it is simple for `switch-minimal` and `switch-internal`, but I think it will not work as users expect for `switch`.

lib/src/metta/runner/stdlib/stdlib.metta
```diff
@@ -338,7 +342,7 @@
 (= (switch-internal $atom (($pattern $template) $tail))
```
vsbogd, line 342: I think it also makes sense to assign the type for the `switch-internal`. I expect type of the `switch-minimal` should work.

TeamSPoon:
I guess I imagined that it would end up being 

`(: switch-minimal (-> Atom Expression Atom)) `

`(: switch-internal (-> Atom Expression Atom)) `

vs `switch` and `case` then with `%Undefined%` would then evaluate what was returned

vsbogd:
Having `case/switch` which doesn't reduce the result automatically makes sense. @Necr0x0Der what do you think, should it be default behavior?

Issue #394, opened 2023-08-11
Bad performance on recursive tuple/list deconstruction


vsbogd:
This issue is to investigate the perfomance in case when list or tuple is recursively deconstructed.

According to @patham9 measurements:
```
(= (TupleCount $tuple) (if (== $tuple ()) 0 (+ 1 (TupleCount (cdr-atom $tuple)))))
!(TupleCount (1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24 25 26 27 28 29 30))
```
takes 2.8 seconds
```
(= (ListCount Nil) 0)
(= (ListCount (Cons $head $tail)) (+ 1 (ListCount $tail)))
!(ListCount (Cons 1 (Cons 2 (Cons 3 (Cons 4 (Cons 5 (Cons 6 (Cons 7 (Cons 8 (Cons 9 (Cons 10 (Cons 11 (Cons 12 (Cons 13 (Cons 14 (Cons 15 (Cons 16 (Cons 17 (Cons 18 (Cons 19 (Cons 20 (Cons 21 (Cons 22 (Cons 23 (Cons 24 (Cons 25 (Cons 26 (Cons 27 (Cons 28 (Cons 29 (Cons 30 Nil)))))))))))))))))))))))))))))))
```
takes 3 seconds
```
(= (BuildList $w $c) (if (> $c 0) (BuildList (Cons 1 $w) (- $c 1)) $w))
!(BuildList Nil 30)
```
and this 4.8 seconds for me. 
Same with tuples 3.8 seconds:
```
(= (BuildTuple $w $c) (if (> $c 0) (BuildTuple (1 $w) (- $c 1)) $w))
!(BuildTuple Nil 30)
```

patham9:
Workaround for one of the probably most critical cases, counting of elements in a collapse: https://gist.github.com/patham9/aa3e93bb369e3fc1660cb473e7e74b55

luketpeterson:
This isn’t fundamental to the issue at hand, and you may already be doing this, but…

Building hyperon with `cmake -DCMAKE_BUILD_TYPE=Release ../` as opposed to just `cmake ../` provides roughly a 50% boost to this test on my setup.

vsbogd:
[...]

PR #1000, opened 2025-08-08
Evaluate expression once until it is not changed

vsbogd:
This makes performance better on tasks like #394 examples and for SICP examples performance is two times better.


vsbogd:
Added unit test https://github.com/trueagi-io/hyperon-experimental/pull/1000/commits/4a96110a07181147f5bf10c9461c6eb63646d03f

Necr0x0Der (approved):

PR #1003, opened 2025-08-15
remove old das_gate

Necr0x0Der:


Necr0x0Der:
@arturgontijo , one thing is if new DAS Bus Client can have similar integration tests as old das_gate? In particular, can in RAM DAs be created for such tests?

arturgontijo:
Hey @Necr0x0Der, there is no in-RAM DAS anymore.

But we do have a tool for that called `das-cli`, this tool starts all necessary DAS components (services and DBs) in order to ease an initial setup.

I'll put up a PR adding `das-cli` (+ tests) to the Actions here and see what you guys think about it.

PR #1004, opened 2025-08-15
Build release by default

vsbogd:
Use Release type as a default for a single-configuration generators like "Unix Makeflies".
It also fixes build for the Ninja multi-configuration generator.


Necr0x0Der (approved):
Works on Mac

PR #1012, opened 2025-08-25
Fix DAS tests job

vsbogd:
Allow passing Python environment version. Fix `das-cli` setup to use this version and virtual environment.
@arturgontijo I have some questions about it:
1. job fails with Python 3.8 (https://github.com/vsbogd/hyperon-experimental/actions/runs/17214743919) what is the minimal version of Python supported by DAS cli?
2. DAS cli requires manual requirements installation (`python3 -m pip install -r requirements.txt`) (see https://github.com/vsbogd/hyperon-experimental/actions/runs/17214573397/job/48834855776) but one would expect `python3 -m pip install .` should be enough.
3. Job still fails after all fixes (https://github.com/vsbogd/hyperon-experimental/actions/runs/17214801060/job/48835629376) not sure why
4. I think the approach with the separate job is not efficient, it requires rebuilding hyperon-experimental before running DAS tests, thus I think it make sense to make this test a separate reusable action and include it into a `common.yml` with "run-das-tests" flag.

### DIFF
--- a/.github/workflows/ci-manual.yml
+++ b/.github/workflows/ci-manual.yml
@@ -28,11 +28,6 @@ on:
         required: true
         default: "[\"Debug\", \"Release\", \"RelWithDebInfo\", \"MinSizeRel\"]"
         type: string
-      run-das-tests:
-        description: 'Run DAS integration tests'
-        required: true
-        default: true
-        type: boolean
 
 jobs:
   ci-manual:
@@ -51,14 +46,3 @@ jobs:
       python-version: ${{ matrix.python-version }}
       cmake-version: ${{ matrix.cmake-version }}
       build-type: ${{ matrix.build-type }}
-
-  das-tests:
-    if: ${{ inputs.run-das-tests }}
-    strategy:
-      matrix:
-        os: ${{ fromJson(inputs.os) }}
-      max-parallel: 5
-    needs: ci-manual
-    uses: ./.github/workflows/das-tests.yml
-    with:
-      os: ${{ matrix.os }}
--- a/.github/workflows/ci-nightly.yml
+++ b/.github/workflows/ci-nightly.yml
@@ -25,13 +25,5 @@ jobs:
       build-type: ${{ matrix.build-type }}
 
   das-tests:
-    if: github.repository == 'trueagi-io/hyperon-experimental'
-    needs: ci-nightly
-    strategy:
-      fail-fast: false
-      matrix:
-        os:  ["ubuntu-22.04", "ubuntu-24.04", "macos-13", "macos-14", "macos-15", "windows-latest"]
-      max-parallel: 5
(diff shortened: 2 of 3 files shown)


vsbogd:
> 1 - das-cli minimal version is 3.10.x ([ref](https://github.com/singnet/das-toolbox/tree/master/das-cli#requirements))
> I think we can hardcode python version here as das-cli is just a handy helper and not a part of metta-bus-client

Agree

> 3 - Let's try pip install it using -e flag (python3 -m pip install -e .), that should be enough

Well, I removed `-e` because in original file after `pip install -e` the source code directory was being removed and this breaks the das-cli installation.

arturgontijo:
> Well, I removed -e because in original file after pip install -e the source code directory was being removed and this breaks the das-cli installation.

Oh I see, yeah let's not remove the TEMP dir...that should work

vsbogd:
> 1 - das-cli minimal version is 3.10.x ([ref](https://github.com/singnet/das-toolbox/tree/master/das-cli#requirements))

Btw, it probably should be the part of `setup.py` as well then.

> Oh I see, yeah let's not remove the TEMP dir...that should work

If `common.config` is a part of application should it also be installed?

arturgontijo:
I just ran these lines on Ubuntu 24.04 and it workes:
```
python3 -m venv /tmp/dastoolbox
source /tmp/dastoolbox/bin/activate

git clone https://github.com/singnet/das-toolbox.git
cd das-toolbox

git checkout tags/0.5.0

cd das-cli/src
python3 -m pip install -r requirements.txt
python3 -m pip install -e .

das-cli --help
```

vsbogd:
What if someone gets the das-cli code and perform `pip install`, the installation will be affected by the same issue.

vsbogd:
I will try to fix the job tomorrow.

vsbogd:
@arturgontijo do you have an idea why Redis doesn't start on macos images https://github.com/vsbogd/hyperon-experimental/actions/runs/17229876711/job/48881457666 or may be how to get the reason from `das-cli`?

vsbogd:
I afraid the reason is `das-cli` tries to run container but fails because CI job is already inside container.

arturgontijo:
> I afraid the reason is das-cli tries to run container but fails because CI job is already inside container.

Yeah, that's right...I see a possible way to expose Docker would be adding this stop before `- name: das-toolbox - DAS configuration`
```
      - name: Set up Docker Buildx
        uses: docker/setup-buildx-action@v3
```

Sorry for all these confusions, we definitely need to improve our tools and docs

vsbogd:
Looks like Docker are not supported by `macos` images (except `macos-13`) and has poor support even on real hardware: https://github.com/marketplace/actions/setup-docker-on-macos#arm64-processors-m-series-used-on-macos-14-images-and-beyond-are-unsupported 
Did you try the instruction from PR on MacOSX?

vsbogd:
@arturgontijo please review changes suggested by this PR.

Issue #872, opened 2025-03-06
Misleading error message when applying bind! to the same reference twice


ngeiswei:
**Describe the bug**
The error message resulting from applying `bind!` to an already bound reference is misleading.

**To Reproduce**
Run the following code

```
!(bind! &i 0)
!(bind! &i 1)
```

**Expected behavior**
An error message like "&i is already bound to 0".

**Actual behavior**
```
[(Error 0 BadType)]
```

**Additional context**
 - MeTTa version: 0.2.2+gc3d57c00

ngeiswei:
It's not a big deal once you know it, but it can be very puzzling the first time.

Necr0x0Der:
I'm not sure it is easy to improve it. The issue is that `bind!` modifies the tokenizer, which is the first thing that processes the code. So the second occurrence of `&i` is replaced with `0` before S-expression with `bind!` is constructed, and the interpreter just sees `(bind! 0 1)`. @vsbogd had an idea to change the tokenizer behavior in such the way that atomspace would keep original syntactic representations and their bindings would be kept in atomspace itself and done at runtime. This would help to resolve this issue, but it is difficult in general to do so for arbitrary tokens (strings, numbers, etc.), because you do need tokenization for them in the preprocessing stage.

ngeiswei:
I see.  Maybe then all that is needed is to improve the BadType error message in general.  By for instance telling the user where the bad type occurred.  Maybe something like (just making something up)
```
(Error 0 BadType in (bind! 0 1))
```

vsbogd:
I working on small type-checking refactoring, will see if I can do this in parallel.

vsbogd:
After #1015 the error is more explaining: `(Error (bind! 0 1) (BadArgType 1 Symbol Number))`

PR #1015, opened 2025-08-27
change badtype behavior

DaddyWesker:


vsbogd:
Fixes #872

DaddyWesker:
Reason for introduction of expression was [this particular assert](https://github.com/DaddyWesker/hyperon-experimental/blob/792590d1d5393dfd84ba600d053d00ce0f4bf44b/python/tests/scripts/d5_auto_types.metta#L32) . Here type contains variable and when I've outputted it inside of string it was impossible to make assert over it. Vitaly suggested to make output not a string but expression and use alpha equality for this assert.

lib/src/metta/runner/stdlib/mod.rs
```diff
@@ -242,26 +248,32 @@ mod tests {
...
             (: Cons (-> $t (List $t) (List $t)))
-            !(metta (Cons S (Cons Z Nil)) %Undefined% &self)
+            !(assertEqual
+                (metta (Cons S (Cons Z Nil)) %Undefined% &self)
+                (Error (Cons S (Cons Z Nil)) (BadType 2 (List (-> Nat Nat)) (List Nat))))
         ");
-        assert_eq!(result, Ok(vec![vec![expr!("Error" ("Cons" "Z" "Nil") "BadType")]]));
+        assert_eq!(result, Ok(vec![vec![UNIT_ATOM]]));
         let result = run_program("
             (: (a b) (C D))
```
vsbogd, line 267:
```suggestion
            (: (c d) (C D))
```

lib/src/metta/runner/stdlib/mod.rs
```diff
@@ -242,26 +248,32 @@ mod tests {
...
 
             (: foo (-> (A B) %Undefined%))
             (= (foo $x) succ)
 
-            !(foo (a b))
+            !(assertEqual
+                (foo (a b))
+                (Error (foo (a b)) (BadType 1 (A B) (C D))))
```
vsbogd, line 274:
```suggestion
            !(assertEqual
                (foo (c d))
                (Error (foo (c d)) (BadType 1 (A B) (C D))))
```

lib/src/metta/runner/stdlib/mod.rs
```diff
@@ -501,22 +519,34 @@ mod tests {
...
         ";
 
         assert_eq!(metta.run(SExprParser::new(program2)),
-            Ok(vec![vec![expr!("Error" "myAtom" "BadType")]]));
+            Ok(vec![vec![UNIT_ATOM]]));
+    }
+
+    #[test]
+    fn test_check_if_function_type_is_applicable_returns_bad_type() {
+        let program = "!(assertEqual (metta (+ 1 2) (-> Atom Atom $t) &self) (Error (+ 1 2) (BadType 1 (-> Atom Atom $t) Number)))";
```
vsbogd, line 546: In such situations format `(BadType <index> <expected> <actual>)` is confusing. Error says sub atom index 1 within `(+ 1 2)` has incorrect type, expected `(-> Atom Atom $t)` found `Number`. But here the error is related to the whole expression not to the atom with index 1 inside the expression. Similar issue exists for return type of the function.

I am not sure what is the best way to fix it. What comes into my mind:
1. Make `BadType` polymorphic and allow omitting the index: `(BadType  (-> Atom Atom $t) Number)`
2. Make index more complex expression. Let's say `(BadType (argument 0) String Number)` for the argument errors and `(BadType atom String Number)` for the full atom type error
3. Make two different errors: `(BadType  (-> Atom Atom $t) Number)` for the atom, `(BadArgType 0 String Number)` for the argument, and `(BadRetType String Number)` for the return type.

I like option 3 more. @Necr0x0Der , @DaddyWesker what do you think?

lib/src/metta/interpreter.rs
```diff
@@ -1047,20 +1048,25 @@ fn type_cast(space: Atom, atom: Atom, expected_type: Atom, bindings: Bindings) -
...
-        match first_match {
-            Some(bindings) => once((return_atom(atom), bindings)),
-            None => once((return_atom(error_atom(atom, BAD_TYPE_SYMBOL)), bindings)),
+                    Err(_) => unreachable!(),
+                };
+                return once((return_atom(atom), first_match.next().unwrap()));
+            }
         }
+        let actual_type = types.iter().nth((idx-1) as usize).unwrap();
+        once((return_atom(Atom::expr([ERROR_SYMBOL, atom.clone(), Atom::expr([BAD_TYPE_SYMBOL, Atom::gnd(Number::Integer(idx)), expected_type.clone(), actual_type.as_atom().clone()])])), bindings))
```
vsbogd, line 1069: Two issues here:
1. `types` vector is unnecessary cloned above
2. `idx` is not position inside the atom but returned as such inside `BadType` error

I tried to fix them in https://github.com/DaddyWesker/hyperon-experimental/pull/29 but there is still question what to return in such situation see my another comment.

lib/src/metta/runner/stdlib/mod.rs
DaddyWesker: I'm actually agree with your opinion. I think splitting BadType to several different errors will enhance readability of the errors by the end-user. Though I'm not sure is this matters in this case.

PR #1019, opened 2025-09-01
Add universal metta! macros to embed MeTTa syntax into Rust programs

vsbogd:
The goal of this PR is to add `metta!` macros which can be used to write MeTTa atoms in a similar to `.metta` files way. `metta!` allows writing symbols, variables, numbers, booleans, strings using usual syntax and it supports other grounded atoms using `{}` notation (similar to `expr!`). It creates constant expressions and symbols when it is possible and thus replaces `constexpr!` and `sym!`. It replaces `expr!` because allows writing same atoms in a more natural MeTTa notation. For example `metta!((-1 $a (A False) "text"))` is equal to `expr!({Number::Integer(-1) a ("A" {Bool(false)} {Str::from_str("text")}))`. I expect that at some point all `expr!` and `sym!` usages will be replaced by `metta!`.


hyperon-macros/src/lib.rs
```diff
@@ -0,0 +1,524 @@
...
+                    match tt {
+                        TokenTree::Literal(l) => {
+                            let s = l.to_string();
+                            let lit = litrs::Literal::parse(s.clone()).expect("Failed to parse literal");
+                            match lit {
+                                litrs::Literal::Integer(_) => (None, TS::Gnd(s, GndType::Int)),
+                                litrs::Literal::Float(_) => (None, TS::Gnd(s, GndType::Float)),
+                                litrs::Literal::String(_) => (None, TS::Gnd(s, GndType::Str)),
+                                litrs::Literal::Bool(_) => (None, TS::Symbol(s)),
+                                _ => (None, TS::Symbol(s)),
```
DaddyWesker, line 161: Since result for both `litrs::Literal::Bool(_)` and `_` is the same maybe it is better to unite them just by removing line `litrs::Literal::Bool(_) => (None, TS::Symbol(s)),`? Or is this intended if, in future, behavior for Bool will change?

vsbogd: Fixed, thanks!

hyperon-macros/src/lib.rs
```diff
@@ -0,0 +1,524 @@
...
+impl Tokenizer {
+    fn new(input: TokenStream) -> Self {
+        let mut prev_end = (0, 0);
+        let input = input.into_iter().flat_map(Self::unroll_group).flat_map(
+            move |it| -> Box<dyn Iterator<Item=InternalToken>> {
+                if prev_end != it.start() {
+                    prev_end = it.end();
+                    Box::new(std::iter::once(InternalToken::Space).chain(std::iter::once(it)))
+                } else {
+                    prev_end = it.end();
```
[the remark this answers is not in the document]
vsbogd: Fixed, thanks!

PR #1020, opened 2025-09-02
MeTTa runner doesn't read config dir by default

### LINKED ISSUE #1016: Metta::new(None) creates default config directory and reads git catalog
`Metta::new(None)` creates default `Environment` instance which in turn creates config directory and reads Git catalog. This leads to random nightly build fails like https://github.com/trueagi-io/hyperon-experimental/actions/runs/17256405639/job/48972116468

I would expect by default `Metta` creates empty environment and not touch filesystem and network. If user needs to read a configuration it should be done by calling additional methods and making additional initialization. For example Metta REPL code should do additional steps to create default config dir and load MeTTa git catalog.

vsbogd:
Fixes #1016 and expected to resolve nightly build random failures. Default behavior of `Environment` is changed:
- by default no configuration directory is expected and read as consequence no default git catalog is read
- REPL and `exts/agents.py` code is changed to read configuration from default configuration directory

`EnvBuilder::set_no_config_dir` is removed and `EnvBuilder::set_default_config_dir` is added instead. In Python API the default config dir is encoded as empty string. Originally I was going to write `EnvBuilder::default_config_dir()` method to return the default config dir, but as this call can fail the error processing makes callers code ugly (not mentioning passing resulting path to the caller via C API). Thus I decided to keep  `EnvBuilder::set_default_config_dir()` method instead.

### DIFF
--- a/python/hyperon/exts/agents/agent_base.py
+++ b/python/hyperon/exts/agents/agent_base.py
@@ -144,7 +144,7 @@ class AgentObject:
         # the caller space is not directly accessible as a context,
         # except the case when _metta is set via get_agent_atom with parent MeTTa
         if self._include_paths is not None:
-            env_builder = Environment.custom_env(include_paths=self._include_paths)
+            env_builder = Environment.custom_env(include_paths=self._include_paths, config_dir="")
             metta = MeTTa(env_builder=env_builder)
         else:
             metta = MeTTa()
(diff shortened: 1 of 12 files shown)


python/hyperon/exts/agents/agent_base.py
```diff
@@ -144,7 +144,7 @@ def _init_metta(self):
         # the caller space is not directly accessible as a context,
         # except the case when _metta is set via get_agent_atom with parent MeTTa
         if self._include_paths is not None:
-            env_builder = Environment.custom_env(include_paths=self._include_paths)
+            env_builder = Environment.custom_env(include_paths=self._include_paths, config_dir="")
```
vsbogd, line 147: @Necr0x0Der, @besSveta  I am not sure if agent should read configuration from file system or not, my opinion that probably not, but I reproduced previous behavior.

PR #1040, opened 2025-09-30
Fix equality for embedded grounded types

### LINKED ISSUE #1033: !(assertEqual (repr 42) "42") fails
**Describe the bug**
```scheme
!(assertEqual
  (repr 42)
  "42")
```
fails while `(repr 42)` correctly outputs `"42"`.

**To Reproduce**
Run `!(assertEqual (repr 42) "42")`

**Expected behavior**
```
[()]
```

**Actual behavior**
```
[(Error (assertEqual (repr 42) "42") "\nExpected: [\"42\"]\nGot: [\"42\"]\nMissed results: \"42\"\nExcessive results: \"42\"")]
```

**Additional context**
 - It looks like the error has been introduced by commit 200a9f5a.  At least I can tell that 6e2a02df behaves normally and 200a9f5a doesn't.

vsbogd:
Fixes #1033 replaces #1035 
`str.rs`, `number.rs` and `bool.rs` are moved into `hyperon_atom` module.
`gnd_eq` function is introduced to compare grounded atoms. Comprison via conversion for `String`, `Number` and `Bool` grounded types is hardcoded. This make comparison of Python strings slower because they are always converted into Rust string before comparison. On the other hand this is a bit faster than #1035 because it converts atoms according to their grounded types and in case of Rust data no conversion is made.

### DIFF
--- a/c/src/atom.rs
+++ b/c/src/atom.rs
@@ -15,7 +15,7 @@ use std::sync::atomic::{AtomicPtr, Ordering};
 
 use hyperon_atom::matcher::{Bindings, BindingsSet};
 use hyperon::metta::runner::bool::Bool;
-use hyperon::metta::runner::number::Number;
+use hyperon_atom::gnd::number::Number;
 
 // =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-
 // Atom Interface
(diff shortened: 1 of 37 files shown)


vsbogd:
@Bitseat please take a look (GitHub doesn't allow me adding you into the list of reviewers)

vsbogd:
I tried the following script:
```
!(let $x (repr 42) (== "42" $x)) ; returns [True]
!(let $x (repr 42) (== $x "42")) ; returns [True]
!(assertEqual (repr 42) "42") ; returns [()]
```
Please try cleanup old build artifacts and rebuild the project from scratch. Installing it using `python3 -m pip install -e ./python[dev]` can help as well.

vsbogd:
Please note that you need use Python interpreter `metta-py` or `metta` to run the test.

Issue #1079, opened 2026-05-03
Space::visit undercounts atoms in workloads with many same-head atoms


gouda-tr4wad4bi:
Bug report — Space::visit undercounts atoms in homogeneous-shape workloads
Summary
SpaceVisitor enumeration of a GroundingSpace undercounts atoms when many
atoms with the same head symbol but distinct sub-atoms are inserted. The
storage and per-atom Space::query are correct — every inserted atom is
findable by query — but visit traverses fewer leaves than were inserted.
This breaks !(get-atoms &self) and any code that relies on bulk
enumeration of a space (counting, snapshotting, iterating for persistence).
Environment

Tested against main at commit 3f76dc4 (the v0.2.10 release).
Also reproduces against the v0.2.10 tag identically.
Pure-Rust reproducer; no PyO3 / Python / C-API involvement.

Reproducer
rust// tests/silent_drop.rs

use hyperon::space::grounding::GroundingSpace;
use hyperon_atom::*;
use hyperon_space::{Space, SpaceVisitor};
use std::cell::Cell;

fn make_atom(n: usize) -> Atom {
    Atom::expr([
        sym!("item-shape-signal"),
        Atom::sym(format!("pattern-{}", n)),
        Atom::sym(format!("target-shape-{}", n)),
        Atom::sym(format!("{}", n)),
    ])
}

struct Counter(Cell<usize>);
impl SpaceVisitor for Counter {
    fn accept(&mut self, _: std::borrow::Cow<'_, Atom>) {
        self.0.set(self.0.get() + 1);
    }
}

#[test]
fn visit_undercounts_after_threshold() {
    let mut space = GroundingSpace::new();
    let target = 1500;

    // Insert 1500 atoms.
    for n in 0..target { space.add(make_atom(n)); }

    // Count via SpaceVisitor.
    let mut counter = Counter(Cell::new(0));
    let _ = space.visit(&mut counter);
    let visit_count = counter.0.get();

    // Confirm storage is intact: each atom is independently queryable.
    let mut query_found = 0usize;
    for n in 0..target {
        if !space.query(&make_atom(n)).is_empty() { query_found += 1; }
    }

    assert_eq!(query_found, target,
        "query() loses atoms");           // PASSES — storage OK
    assert_eq!(visit_count, target,
        "visit() undercounts: got {} of {}", visit_count, target);  // FAILS
}
Cargo.toml:
toml[dependencies]
hyperon = "0.2.10"
[...]

PR #1081, opened 2026-07-30
Fix operator precedence in TrieKey::value() corrupting decodes past 1024 stored keys

### LINKED ISSUE #1076: GroundingSpace trie crash (trie.rs:179) on match/get-atoms after loading ~1,900+ expressions
## Summary

After a GroundingSpace accumulates approximately 1,900 or more expressions via `metta.run()`, subsequent queries (`match`, `get-atoms`) on that space trigger a panic at `trie.rs:179` (`unwrap()` on `None`).

The crash occurs on the **first query after loading** — no explicit `add-atom` call is needed if the loaded modules contain executable statements that insert data atoms as side effects.

## Environment

- **hyperon**: 0.2.10 (pip)
- **Python**: 3.10+
- **OS**: Windows (also previously reproduced on Ubuntu 22.04)

### Quick version (requires project files)

Clone [Project Dagaz](https://github.com/Jolnir-Sefir/dagaz) and run the test script from the project root:

```bash
python test_hyperon_trie.py
```

This loads 22 MeTTa modules (~12,000 lines, ~3,000+ expressions) via `metta.run()`, then attempts `!(get-atoms &self)`. The crash occurs on that first query.

<details>
<summary>test_hyperon_trie.py (full script)</summary>

```python
import re
import sys
from pathlib import Path

from hyperon import MeTTa

MODULES = [
    'foundations', 'beliefs', 'affect', 'actions', 'safety',
    'structure_learning', 'atom_lifecycle', 'cycle', 'policy_efe',
    'planning', 'action_learning', 'self_model',
    'analogy_blending', 'conversation_model', 'perception',
    'proprioception', 'abduction', 'semantic_primitives',
    'dimensional_primitives', 'semantic_grounding',
    'action_grounding', 'grounding_hypotheses'
]

def find_module(mod_name, base):
    for pattern in [f"core_{mod_name}.metta", f"{mod_name}.metta", f"core/{mod_name}.metta"]:
        p = base / pattern
        if p.exists():
            return p
    return None

def main():
    base = Path(".")
    metta = MeTTa()

    for mod in MODULES:
        fpath = find_module(mod, base)
        if fpath is None:
            print(f"  SKIP: {mod}")
            continue
        code = fpath.read_text(encoding='utf-8')
        code = re.sub(r'!\(import!\s+[^)]*\)', '', code)
        code = re.sub(r'!\(bind!\s+[^)]*\)', '', code)
        try:
            metta.run(code)
            print(f"  OK: {mod}")
        except Exception as e:
            print(f"  FAIL: {mod} — {e}")

    print("\n--- Querying &self (crash expected here) ---")
    try:
        result = metta.run("!(get-atoms &self)")
        print(f"get-atoms returned {len(result[0]) if result and result[0] else '?'} atoms")
    except Exception as e:
        print(f"CRASH: {e}")

if __name__ == "__main__":
    main()
```
</details>

### Standalone version (no project files needed)

I also tested a synthetic version that generates 1,200 flat function definitions (`(= (func-N $x) (result-N $x))`) and then does `add-atom` + `match`. That version **did not crash**, suggesting the trigger depends on the structural diversity of the expressions (nested `let*`, `match`, `collapse`, arrow types, etc.), not just the count.

If you need a fully self-contained reproduction without cloning the project, let me know and I can work on narrowing down which module or expression pattern tips the trie over.

## Stack Trace

```
thread '<unnamed>' (21100) panicked at hyperon-space\src\index\trie.rs:179:71:
called `Option::unwrap()` on a `None` value

thread '<unnamed>' (21100) panicked at library\core\src\panicking.rs:225:5:
panic in a function that cannot unwind
stack backtrace:
   0:     0x7ffb2a5ca682 - git_blob_create_fromdisk
   1:     0x7ffb2a54e7ca - git_blob_create_fromdisk
   [... full trace truncated for readability ...]
  78:     0x7ff67feb1230 - <unknown>
  79:     0x7ffbc860e8d7 - BaseThreadInitThunk
  80:     0x7ffbca5cc48c - RtlUserThreadStart
thread caused non-unwinding panic. aborting.
```

## Observed Behavior

All 22 modules load successfully via `metta.run()`. The first subsequent query (`get-atoms` or `match`) on `&self` triggers the panic.

## Expected Behavior

`!(get-atoms &self)` should return the contents of the space. `!(match &self (belief $o $v $p) $o)` should return matching atoms.

## Characterization

From earlier systematic testing (before I built an alternative runtime):

| Test | Result | Meaning |
|------|--------|---------|
| ~3,200 exprs loaded, no post-load query involving mutated branches | **PASS** | Trie builds during loading |
| ~3,200 exprs, then add 1 data atom, then match on that atom | **CRASH** | Post-load mutation + query crashes |
| ~460 exprs + post-load add-atom + match | **PASS** | Below threshold |
| ~1,025 exprs + post-load add-atom + match | **PASS** | Below threshold |
| ~1,900 exprs + post-load add-atom + match | **CRASH** | **Threshold between ~1,025–1,900** |
| ~3,200 exprs + add fresh symbol `xbelief`, match xbelief | **CRASH** | Not symbol-specific |
| ~3,200 exprs + add belief, match on *different* symbol `observation` | **PASS** | Corruption is branch-local |

Key properties:

1. **Size-dependent, not content-dependent.** Any symbol crashes past the threshold.
2. **Branch-local corruption.** Querying unaffected trie branches still works after a corrupting insertion.
3. **Structural diversity matters.** 1,200 flat synthetic expressions did not trigger the crash; ~1,900 structurally diverse real expressions (nested `let*`, `match`, `collapse`, arrow types) did. The threshold may depend on trie topology, not just atom count.

## Impact

This blocks any MeTTa program that loads a nontrivial codebase and then queries or mutates runtime state — the basic pattern for cognitive loops, simulations, or any stateful application.

## Workaround

Separate static code from dynamic state into different GroundingSpaces:
- `&self`: function definitions only (never mutated after loading)
- `&state`: runtime data atoms (small, stays below threshold)

This prevents post-load mutation of the large trie. Cross-space queries (`match &state ...` from functions in `&self`) work correctly.

## Context

Discovered while building a 22-module Active Inference cognitive architecture in MeTTa (~12,000 lines). The cognitive loop needs to add/remove belief, observation, and structural hypothesis atoms every cycle. With ~3,000 function definitions in `&self`, every runtime mutation is a potential crash. Full investigation notes: [`TRIE_CRASH_FINDINGS.md`](https://github.com/Jolnir-Sefir/dagaz/blob/main/TRIE_CRASH_FINDINGS.md).

I'm aware that MORK/PathMap is replacing the trie infrastructure. This report may still be useful as a regression test to ensure PathMap handles dynamic insertion after large bulk loads correctly.A clear and concise description of what the bug is.

**To Reproduce**
1. Compile MeTTa using settings '...'
2. Run the following MeTTa script '....'
3. See output

**Expected behavior**

**Actual behavior**
If applicable, add logs to help explain your problem.

**Additional context**
 - MeTTa version: [output of `metta --version`]
 - anything else which is relevant

MohBani:
Fixes #1076. Addresses #1079 (its verbatim reproducer exercises the same
broken decoder locally, although the reported silent-undercount surface
was not reproduced).

## Problem

`TrieKey::value()` in `hyperon-space/src/index/trie.rs` reads:

    self.0 & TK_VALUE_MASK - TK_MAX_EXPRESSION_SIZE

`-` binds tighter than `&`, so this is `self.0 & (TK_VALUE_MASK - 1024)`:
the +1024 offset added by `TrieKey::new()` is never subtracted, and bit 10
of the payload is force-cleared. For storage ids `v < 1024` the two errors
cancel (v+1024 is v with bit 10 set), so decode is correct by accident;
from id 1024 the carry breaks the cancellation and every decode is garbage.

Decode runs only on read-back paths (result materialization, iter/visit)
via `get_atom_unchecked`: the Hash arm panics at trie.rs:179
(`get_atom(garbage).unwrap()` on None — the #1076 signature), the Index arm
is an out-of-bounds `HoleyVec::get_unchecked` (segfault/UB). Insertion and
exact-key comparison never decode: the defect stays latent during
insertion, the stored data remains intact, exact queries keep working, and
failure appears when a read-back path decodes an affected key — consistent
with #1079's intact storage and lossy enumeration.

## Fix

    (self.0 & TK_VALUE_MASK) - TK_MAX_EXPRESSION_SIZE

## Tests

- `trie_key_value_roundtrip_above_max_expression_size`: round-trips
  `new()`/`value()` across the 1024 boundary, both stores, both match modes.
  Fails at v=1024 without the fix.
- `atom_index_survives_key_ids_above_max_expression_size`: 3000 distinct
  expressions into `AtomIndex`, variable query, expect 3000 results.
  Without the fix it reproduces the exact #1076 panic (trie.rs:179).
- `atom_index_iter_enumerates_all_atoms_above_key_id_threshold`: the
  enumeration surface at the index layer — 1500 same-head atoms,
  `iter().count()` must equal 1500. Without the fix it fails at trie.rs:179.
- `grounding_space_visit_enumerates_all_atoms_above_key_id_threshold`
  (lib/src/space/grounding/mod.rs): #1079's reproducer, verbatim — its
  exact atom shape and Cell-based visitor, 1500 atoms, each asserted
  individually queryable, then `Space::visit` must count 1500. Without the
  fix: all 1500 exact queries pass, then the enumeration panics at
  trie.rs:179 (the silent 1293 undercount of the original report was not
  reproducible on v0.2.10).

## Validation

- `cargo test -p hyperon-space` green (32 + 6, incl. the three tests there);
  `cargo test -p hyperon space::grounding::test` green (28, incl. the visit
  regression).
- Python-level repro from #1076 (1500 same-head atoms + variable match):
  abort → 1500 clean results across the fix.
- A large structurally-diverse MeTTa workload that crashed
  deterministically (both surfaces) completes cleanly, with A/B-identical
  results on all paths that completed under both builds.


MohBani:
Context on how this was found, for the curious: the workload behind the report is **Topology of Thought** — a deterministic forward-saturation theorem-proving substrate built on hyperon/MeTTa (Python pipeline + MeTTa rule space + LLM-assisted side-condition certification; the reasoning itself is done by saturation, not by the LLM). It recently completed all 41 questions of the ENS Lyon–Cachan 1996 exam — a staged proof of the van der Waerden permanent conjecture (the Egorychev–Falikman theorem) — end-to-end on the patched interpreter. The reductio probe of question V.4 was the first space in the bench to cross 1024 distinct stored keys, which is how this defect surfaced. Happy to share more about the workload, and to stress-test future space implementations (e.g. the MORK/PathMap direction) against this class of high-cardinality, structurally diverse spaces.

tiararodney:
Hi,

just wanted to acknowledge that this PR fixes the issue for me. Thank you, @MohBani!

I stumbled over this PR in regards to #1076, and #1079, since I ran into the same issue with around ~20k typed atoms...

- hyperon v0.2.10-1-g1e4030c5
- Python 3.14.4
- Linux 7.0.0-31-generic x86_64

```python
from hyperon import MeTTa

m = MeTTa()
atoms = "\n".join(f"(: item_{i} SomeType)" for i in range(1024))
m.run(atoms)
result = m.run("!(match &self (: $x SomeType) $x)")  # that panics..
```

1022 atoms pass; 1023 crash... From what I was able to debug myself, pretty much aligns with the findings of this PR.

